PR #589, opened 2026-02-16
fix: strip unsupported parameters from Codex model requests

0-don:
Codex models (e.g. `gpt-5.3-codex`) reject `temperature`, `top_p`, `frequency_penalty`, and `presence_penalty` when sent via the Responses API, causing 502 errors when these parameters are passed through.

This extends the existing parameter stripping in `applyCodexOAuthTransform` to also remove these unsupported parameters.


Crypticelo:
+1

PR #874, opened 2026-03-09
fix: increase SSE scanner max line size from 40MB to 500MB

touwaeriol:
## 背景 / Background

通过 Gemini 生成 4K 分辨率图片时，上游响应中的 base64 数据行可能超过 40MB，导致 `bufio.Scanner` 报 `token too long` 错误，客户端收到空的 candidates。

When generating 4K resolution images via Gemini, the base64 data line in upstream SSE responses can exceed 40MB, causing `bufio.Scanner` to fail with `token too long` error. The client receives empty candidates as a result.

---

## 目的 / Purpose

将 SSE scanner 的最大行大小从 40MB 提升到 500MB，支持 4K 甚至 8K 图片的 base64 传输。`bufio.Scanner` 本身是自适应的（初始 64KB，按需增长），增大上限不会影响正常文本响应的内存使用。

Increase the SSE scanner max line size from 40MB to 500MB to support 4K and even 8K image base64 transfer. `bufio.Scanner` is adaptive (starts at 64KB, grows as needed), so increasing the cap has zero impact on memory usage for normal text responses.

---

### 后端 / Backend

- **`gateway_service.go`**：`defaultMaxLineSize` 常量从 `40 * 1024 * 1024` 改为 `500 * 1024 * 1024`
- **`config.go`**：`viper.SetDefault("gateway.max_line_size", ...)` 同步从 40MB 改为 500MB（viper 配置默认值会覆盖代码常量，两处必须同步修改）

---

- **`gateway_service.go`**: Update `defaultMaxLineSize` constant from `40 * 1024 * 1024` to `500 * 1024 * 1024`
- **`config.go`**: Update `viper.SetDefault("gateway.max_line_size", ...)` from 40MB to 500MB (viper config default overrides the code constant, both must be updated in sync)

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -1402,7 +1402,7 @@ func setDefaults() {
 	viper.SetDefault("gateway.concurrency_slot_ttl_minutes", 30) // 并发槽位过期时间（支持超长请求）
 	viper.SetDefault("gateway.stream_data_interval_timeout", 180)
 	viper.SetDefault("gateway.stream_keepalive_interval", 10)
-	viper.SetDefault("gateway.max_line_size", 40*1024*1024)
+	viper.SetDefault("gateway.max_line_size", 500*1024*1024)
 	viper.SetDefault("gateway.scheduling.sticky_session_max_waiting", 3)
 	viper.SetDefault("gateway.scheduling.sticky_session_wait_timeout", 120*time.Second)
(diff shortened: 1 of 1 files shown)


ztj7728:
gemini渠道已成功出图4K，但是反重力渠道仍然无法正常出图4K。

PR #1204, opened 2026-03-21
fix(settings): prevent SMTP config overwrite and stabilize SMTP test after refresh

Eilen6316:
## Summary

This PR fixes two SMTP issues in admin settings:

1. SMTP settings could be unintentionally overwritten with empty values when settings load failed on frontend, leading to mail service becoming unavailable after some time.
2. SMTP test could fail with 535 after page refresh due to stale/autofilled password behavior and inconsistent fallback handling.

## Changes

- Backend: protect existing SMTP config from empty overwrite during settings update.
- Backend: normalize SMTP fields and improve `test-smtp` / `send-test-email` fallback to saved SMTP config.
- Backend: trim persisted SMTP values when reading config from settings store.
- Frontend: add load-failure guard (`loadFailed`) to prevent unsafe save/test actions.
- Frontend: prevent browser autofill confusion on SMTP password and only send test password when user manually edited it in current session.

## Files

- `backend/internal/handler/admin/setting_handler.go`
- `backend/internal/service/email_service.go`
- `frontend/src/views/admin/SettingsView.vue`

## Notes

- This PR intentionally keeps scope focused on SMTP stability and refresh/test consistency.

### DIFF
--- a/backend/internal/handler/admin/setting_handler.go
+++ b/backend/internal/handler/admin/setting_handler.go
@@ -231,10 +231,26 @@ func (h *SettingHandler) UpdateSettings(c *gin.Context) {
 	if req.DefaultBalance < 0 {
 		req.DefaultBalance = 0
 	}
+	req.SMTPHost = strings.TrimSpace(req.SMTPHost)
+	req.SMTPUsername = strings.TrimSpace(req.SMTPUsername)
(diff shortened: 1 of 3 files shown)


xiyanghenguai-creator:
你好我sub2api已经是最新的了 我也看了1102，但是我找不见那个对应文件夹

PR #1216, opened 2026-03-22
feat: add support for OpenAI-compatible providers using Chat Completions API

Coke1120:
## Summary

Adds support for OpenAI-compatible API providers that use the standard Chat Completions API (`/v1/chat/completions`) instead of the OpenAI Responses API (`/v1/responses`).

This enables Sub2API to work with providers like **Alibaba Coding Plan** which only support the Chat Completions endpoint.

## Changes

- Add `IsOpenAIChatCompletionsMode()` method to Account struct
- Add `openai_chat_completions_mode` field in account's `extra` configuration  
- Modify account test service to use `/chat/completions` endpoint when enabled
- Modify gateway service to build correct URL for Chat Completions mode
- Add direct forwarding function for Chat Completions API without Responses conversion

## Usage

Set `openai_chat_completions_mode: true` in the account's `extra` field to enable Chat Completions mode.

**Example account configuration:**
```json
{
  "platform": "openai",
  "type": "apikey",
  "credentials": {
    "base_url": "https://coding-intl.dashscope.aliyuncs.com/v1",
    "api_key": "sk-sp-xxx"
  },
  "extra": {
    "openai_chat_completions_mode": true
  }
}
```

## Use Cases

- Alibaba Coding Plan (`coding-intl.dashscope.aliyuncs.com`)
- Other OpenAI-compatible providers that don't support the Responses API

## Testing

Tested with Alibaba Coding Plan endpoint:
- Account connection test works correctly
- Chat Completions requests are forwarded properly
- Both streaming and non-streaming responses handled correctly

### DIFF
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -935,6 +935,22 @@ func (a *Account) IsOpenAIPassthroughEnabled() bool {
 	return false
 }
 
+
+// IsOpenAIChatCompletionsMode 返回 OpenAI 账号是否使用标准 Chat Completions API。
+// 某些 OpenAI 兼容的第三方 API（如 Alibaba Coding Plan）不支持 Responses API，
+// 需要使用标准的 /v1/chat/completions 端点。
+//
+// 字段：accounts.extra.openai_chat_completions_mode。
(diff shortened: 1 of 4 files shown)


otakniu:
怎么还没有合到主分支？

PR #1262, opened 2026-03-24
fix(openai): preserve bare gpt-5.3-codex-spark across forwarding paths

InCerryGit:
## Summary

This fixes an OpenAI forwarding inconsistency where the final upstream model choice for `gpt-5.3-codex-spark` was not decided in one explicit place.

image

Previously, the actual upstream model could be implicitly changed by a mix of:
- account/group model mapping,
- `normalizeCodexModel()`,
- `applyCodexOAuthTransform()`,
- and per-entry forwarding logic.

That made bare `gpt-5.3-codex-spark` handling inconsistent across the normal OpenAI path, OpenAI-compatible paths, WebSocket forwarding, `UpstreamModel` reporting, and compat prompt-cache behavior.

## What was wrong

The core problem was a responsibility boundary issue:

1. **billing / account mapping**
2. **final upstream model selection**
3. **transport / OAuth request cleanup**

These concerns were partially mixed together.

As a result, changing one Spark mapping was not enough, because different forwarding paths could still diverge:
- HTTP forwarding could resolve one model,
- compat handlers could report another,
- WebSocket forwarding could rewrite differently,
- and compat prompt-cache logic could derive keys from a different model family than the one actually sent upstream.

### 1. Added an explicit upstream model resolver
Introduced `resolveOpenAIUpstreamModel()` in:

- `backend/internal/service/openai_model_mapping.go`

This is now the single place that decides the **final upstream model** for OpenAI forwarding behavior, including the bare `gpt-5.3-codex-spark` case.

### 2. Wired all real forwarding entry points to the same resolver
Updated these paths to use the explicit upstream resolver:

- `backend/internal/service/openai_gateway_service.go`
- `backend/internal/service/openai_gateway_chat_completions.go`
- `backend/internal/service/openai_gateway_messages.go`
- `backend/internal/service/openai_ws_forwarder.go`

This keeps the normal OpenAI path, compat paths, and WS forwarding aligned.

### 3. Pulled forwarding policy back out of OAuth/transform cleanup
Updated:

- `backend/internal/service/openai_codex_transform.go`

So that `applyCodexOAuthTransform()` no longer silently decides the final upstream model. It now stays focused on request cleanup / transport adaptation instead of forwarding policy.

### 4. Aligned compat prompt-cache and reporting with the final upstream model
Updated:

- `backend/internal/service/openai_compat_prompt_cache_key.go`

So prompt-cache derivation follows the same resolved upstream model family.

Also aligned `UpstreamModel` reporting to match the actual model chosen by the new resolver.

## Tests

Added/updated targeted tests for:
- explicit upstream model resolution,
- bare Spark preservation in the forwarding policy layer,
- OAuth transform no longer rewriting already-resolved bare Spark,
- compat prompt-cache behavior for Spark family.

Validated in Docker with targeted service tests:

- `TestResolveOpenAIForwardModel_PreventsClaudeModelFromFallingBackToGpt51`
- `TestResolveOpenAIUpstreamModel`
- `TestShouldAutoInjectPromptCacheKeyForCompat`
- `TestDeriveCompatPromptCacheKey_UsesResolvedSparkFamily`
- `TestApplyCodexOAuthTransform_PreservesBareSparkModel`
- `TestApplyCodexOAuthTransform_TrimmedModelWithoutPolicyRewrite`

## Notes

This change does **not** change billing semantics.  
It only makes the final upstream forwarding decision explicit and consistent across all OpenAI forwarding surfaces.

### DIFF
--- a/backend/internal/service/openai_gateway_messages.go
+++ b/backend/internal/service/openai_gateway_messages.go
@@ -41,6 +41,7 @@ func (s *OpenAIGatewayService) ForwardAsAnthropic(
 	}
 	originalModel := anthropicReq.Model
 	applyOpenAICompatModelNormalization(&anthropicReq)
+	normalizedModel := anthropicReq.Model
 	clientStream := anthropicReq.Stream // client's original stream preference
 
 	// 2. Convert Anthropic → Responses
@@ -60,13 +61,14 @@ func (s *OpenAIGatewayService) ForwardAsAnthropic(
 	}
 
 	// 3. Model mapping
-	billingModel := resolveOpenAIForwardModel(account, originalModel, defaultMappedModel)
+	billingModel := resolveOpenAIForwardModel(account, normalizedModel, defaultMappedModel)
 	upstreamModel := resolveOpenAIUpstreamModel(billingModel)
 	responsesReq.Model = upstreamModel
 
 	logger.L().Debug("openai messages: model mapping applied",
 		zap.Int64("account_id", account.ID),
 		zap.String("original_model", originalModel),
+		zap.String("normalized_model", normalizedModel),
 		zap.String("billing_model", billingModel),
 		zap.String("upstream_model", upstreamModel),
 		zap.Bool("stream", isStream),


InCerryGit:
@Wei-Shaw Could you help review and merge this PR? I can still only use the version I compiled myself. Thank you very much.

PR #1452, opened 2026-04-04
OpenAI: preserve Chat Completions response_format when forwarding to Responses

zhuhaow:
## What changed
- preserve Chat Completions response_format when requests are forwarded through the Responses API
- keep structured output schemas intact for both OpenAI API key and OpenAI OAuth accounts
- add regression tests for request conversion and forwarding

## Why
- Chat Completions requests were dropping response_format during Chat -> Responses translation
- that caused structured output schemas to be lost before the upstream request was sent

## Testing
- GOCACHE=/tmp/sub2api-go-build GOMODCACHE=/tmp/sub2api-go-mod go test ./internal/pkg/apicompat ./internal/service -run 'TestChatCompletionsToResponses_ResponseFormat|TestOpenAIGatewayService_ForwardAsChatCompletions_(APIKey|OAuth)PreservesStructuredOutput'

### DIFF
--- a/backend/internal/pkg/apicompat/chatcompletions_responses_test.go
+++ b/backend/internal/pkg/apicompat/chatcompletions_responses_test.go
@@ -157,6 +157,32 @@ func TestChatCompletionsToResponses_ReasoningEffort(t *testing.T) {
 	assert.Equal(t, "auto", resp.Reasoning.Summary)
 }
 
+func TestChatCompletionsToResponses_ResponseFormat(t *testing.T) {
+	req := &ChatCompletionsRequest{
+		Model: "gpt-4o",
+		Messages: []ChatMessage{
+			{Role: "user", Content: json.RawMessage(`"Return structured data"`)}},
+		ResponseFormat: json.RawMessage(`{
+			"type":"json_schema",
+			"json_schema":{
+				"name":"weather",
+				"strict":true,
+				"schema":{
+					"type":"object",
+					"properties":{"city":{"type":"string"}},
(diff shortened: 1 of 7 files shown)


ddnio:
Nice fix for the OpenAI path — we hit the same bug in production (langchain-openai 1.1.x `with_structured_output` returned plain text and pydantic `json_invalid`).

One additional gap worth flagging: the **Anthropic-platform path** has the symmetric problem. After CC→Responses conversion, `ResponsesToAnthropicRequest` in `backend/internal/pkg/apicompat/responses_to_anthropic_request.go` doesn't consume `req.Text.Format`, and `AnthropicOutputConfig` in `types.go` only has `Effort` (no `Format` field). So even if this PR lands, a `response_format` through a group whose platform is Anthropic (handler: `GatewayService.ForwardAsChatCompletions` → Anthropic) still gets dropped before reaching Claude.

Anthropic shipped structured outputs GA on 2025-11 (`output_config.format`, supports `type: "json_schema"` and `type: "json"` — note `json` not `json_object`). Mapping is straightforward:

```
Responses text.format {type: json_schema, schema, name, strict}
  → Anthropic output_config.format {type: json_schema, schema}  // name/strict dropped
Responses text.format {type: json_object}
  → Anthropic output_config.format {type: json}
```

Docs: https://platform.claude.com/docs/en/build-with-claude/structured-outputs

Happy to send a follow-up PR for the Anthropic path once this lands, or include it here if you'd prefer one unified PR — let me know which is easier to review.

PR #1488, opened 2026-04-07
fix: support STARTTLS for SMTP on port 587

### LINKED ISSUE #1470: Mailersend SMTP connection problem
I am using Mailersend as SMTP server with TLS. However when I configured all the required details with "Enable TLS encryption for SMTP connection" enabled, no matter it is "test connection" or "send test email", it kept reporting "Failed to send test email: tls dial: tls: first record does not look like a TLS handshake". Based on analysis, it might be because Sub2API is trying implicit TLS, while MailerSend on port 587 expects STARTTLS. This is just my guess. Does anyone else have the same issue?

octo-patch:
Fixes #1470

## Problem

When **Enable TLS encryption for SMTP connection** is turned on, Sub2API called `tls.Dial` to establish an implicit TLS connection. This works correctly for port 465 (SMTPS), but port 587 — used by providers like MailerSend — expects **STARTTLS**: a plain TCP connection that is subsequently upgraded to TLS via the `STARTTLS` SMTP command.

As a result, users who configured port 587 + TLS received:
```
tls dial: tls: first record does not look like a TLS handshake
```

## Solution

Split the TLS send path into two functions based on the configured port:

- **Port 465** → `sendMailImplicitTLS`: connects directly via `tls.Dial` (existing behaviour, unchanged).
- **All other ports (e.g. 587)** → `sendMailSTARTTLS`: opens a plain TCP connection, then calls `client.StartTLS()` to negotiate TLS before authentication.

Both paths enforce TLS 1.2+ and use the same TLS config (`ServerName` set to the SMTP host).

No configuration changes are needed — existing users are unaffected because the routing is automatic based on the port number.

## Testing

- Port 587 + STARTTLS (e.g. MailerSend, Gmail relay): connection now upgrades via STARTTLS correctly.
- Port 465 + implicit TLS: existing path unchanged, continues to work as before.
- TLS disabled (`UseTLS = false`): unchanged, falls through to `smtp.SendMail`.

### DIFF
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"math/big"
+	"net"
 	"net/smtp"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


Marukoyo:
好像还是不支持

lekoOwO:
@Wei-Shaw please merge this

MaxHu-arpg:
时至今日依然不支持STARTTLS

Marukoyo:
作者不愿意写 没办法

MaxHu-arpg:
v0.1.169 好像支持了，只是后台测试连接不成功，发送测试邮件实际上能发

Wei-Shaw:
该问题已在 #5147 中实现并合入 main（连同评论中「测试连接失败但发信成功」的不一致一并修复）：UseTLS 时先尝试隐式 TLS（465 语义），服务器以明文应答（587 提交端口语义）则自动降级为强制 STARTTLS；测试连接与发送共用同一建连路径。相比本 PR 的按端口路由方案，对非标准端口上的隐式 TLS 部署零回归。感谢这个 PR 的分析，实现思路参考了它。

PR #1589, opened 2026-04-12
Fix IPv6-only proxy probing

vvrfxyz:
## Summary
- replace the proxy probe URLs with HTTPS endpoints that are reachable from IPv6-only proxies
- parse `ifconfig.co` and `api64.ipify.org` responses instead of `ip-api` / `httpbin`
- update the repository tests to cover the new probe sources

## Context
The admin proxy test / quality check currently probes `ip-api.com` and `httpbin.org`. In production we hit a socks5 proxy with an IPv6-only exit: account traffic worked through the proxy, but IP management always marked it unavailable because those probe endpoints failed while IPv6-capable HTTPS endpoints succeeded.

## Verification
- validated on a live IPv6-only socks5 proxy that `https://ifconfig.co/json` and `https://api64.ipify.org?format=json` return the exit IPv6
- updated the unit tests for the new probe parsers and fallback order

### DIFF
--- a/backend/internal/repository/proxy_probe_service.go
+++ b/backend/internal/repository/proxy_probe_service.go
@@ -44,14 +44,14 @@ const (
 	defaultProxyProbeResponseMaxBytes = int64(1024 * 1024)
 )
 
-// probeURLs 按优先级排列的探测 URL 列表
-// 某些 AI API 专用代理只允许访问特定域名，因此需要多个备选
+// probeURLs 按优先级排列的探测 URL 列表。
+// 这里必须选择同时支持 HTTPS 和 IPv6 的站点，否则 IPv6-only 代理会被误判为不可用。
 var probeURLs = []struct {
 	url    string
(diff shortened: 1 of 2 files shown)


gopkg-dev:
good job 👍

GottenHeave:
Hi @Wei-Shaw 👋

I'm currently blocked by the issue described in #642 / #1112 — my setup runs through an IPv6-only HTTP proxy, and the proxy probe keeps marking it as unavailable even though traffic works fine through it. This PR directly addresses the root cause by switching the probe endpoints to IPv6-capable ones.

Both Claude and OpenAI's API endpoints support IPv6, so ruling out IPv6-only proxies is incorrect behavior, fixing this would be a great improvement.

This PR would also resolve #642 and #1112.

Could this be reviewed and merged when you get a chance? Really appreciate the work from @vvrfxyz and would love to see it land. Thanks!

PR #1634, opened 2026-04-14
fix(apicompat): flatten tool_choice for Responses API

ddnio:
## Summary

Chat Completions requests with a forced function tool choice fail with upstream HTTP 400 when the gateway converts them to the Responses API:

```
{"error":{"message":"Unknown parameter: 'tool_choice.function'.","type":"invalid_request_error"}}
```

The Chat Completions API wraps the function name in a nested object:

```json
{"type":"function","function":{"name":"X"}}
```

The Responses API expects it flat:

```json
{"type":"function","name":"X"}
```

`ChatCompletionsToResponses` was passing `req.ToolChoice` through unchanged (the inline comment said "already compatible format — pass through directly"), so the nested shape leaked upstream and was rejected.

## Changes

- `chatcompletions_to_responses.go`
  - Add `remapChatToolChoiceToResponses` that flattens any `type=function` tool_choice, tolerating mixed shapes (e.g. both top-level `name` and nested `function.name`). Strings (`auto`/`none`/`required`) and other tool types pass through unchanged.
  - `convertChatFunctionCallToToolChoice` now emits the flat Responses shape for the legacy `function_call` field.
- `anthropic_to_responses.go`
  - `convertAnthropicToolChoiceToResponses` emits the flat shape for the `tool` branch (Anthropic → Responses forced tool choice).
- `responses_to_anthropic_request.go`
  - `convertResponsesToAnthropicToolChoice` accepts both the flat and legacy nested shapes, avoiding a reverse-path regression after the forward converters switched to flat.

## Tests

- `TestChatCompletionsToResponses_ToolChoice_NestedObject` — covers nested → flat, already flat (idempotent), `auto`/`required` strings, mixed shapes.
- `TestConvertResponsesToAnthropicToolChoice` (new file) — covers `auto`/`none`/`required` strings, flat Responses format, and legacy nested format.
- Existing `TestChatCompletionsToResponses_LegacyFunctions` and `TestAnthropicToResponses_ToolChoiceSpecific` were strengthened to explicitly assert the flat layout (no nested `function` field).

All `./internal/pkg/apicompat/...` tests pass.

## Reference

OpenAI API reference (Realtime / Responses) defines the forced function tool choice as:

```python
tool_choice = {"type": "function", "name": "get_weather"}
```

## Test plan
- [x] Unit tests in `backend/internal/pkg/apicompat/` all pass
- [x] Verified end-to-end against production traffic (`gpt-5.2` + forced `SummaryInfo`): 200 streaming SSE with valid `tool_calls` delta instead of upstream 400

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_responses_test.go
+++ b/backend/internal/pkg/apicompat/anthropic_responses_test.go
@@ -832,12 +832,13 @@ func TestAnthropicToResponses_ToolChoiceSpecific(t *testing.T) {
 	resp, err := AnthropicToResponses(req)
 	require.NoError(t, err)
 
+	// Responses API uses flat {"type":"function","name":"X"}, not nested.
 	var tc map[string]any
 	require.NoError(t, json.Unmarshal(resp.ToolChoice, &tc))
 	assert.Equal(t, "function", tc["type"])
-	fn, ok := tc["function"].(map[string]any)
-	require.True(t, ok)
-	assert.Equal(t, "get_weather", fn["name"])
+	assert.Equal(t, "get_weather", tc["name"])
+	_, hasNested := tc["function"]
+	assert.False(t, hasNested, "Responses API rejects nested 'function' field in tool_choice")
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


zhang-qihao:
Still in progresss?

PR #1692, opened 2026-04-16
fix(usage): support searchable API key filter on /usage page

hansnow:
## Summary
- keep the existing `/usage` API key filter UI as a `Select`, but make it searchable
- stop preloading only the first 100 API keys and load matching candidates on demand through the existing `/keys` API
- add regression coverage for the new `Select` search event and the user usage page filter flow

## Problem
The user usage page could only filter API keys from a preloaded list, which made the control effectively unsearchable and also meant keys beyond the first 100 entries were hard to reach.

## Solution
- add an optional `search` event to the shared `Select` component
- update the user usage page to debounce remote key lookup and keep the selected key label stable across result refreshes
- preserve existing usage query, stats, pagination, sorting, and CSV export behavior

## Testing
- `docker run --rm -v "$PWD/frontend:/app" -w /app node:24-alpine sh -lc 'corepack enable && corepack prepare pnpm@10.33.0 --activate && pnpm install --frozen-lockfile && pnpm exec vitest run src/components/common/__tests__/Select.spec.ts src/views/user/__tests__/UsageView.spec.ts && pnpm run typecheck'`
- `cd deploy && POSTGRES_PASSWORD=testpass JWT_SECRET=0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef TOTP_ENCRYPTION_KEY=abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789 ADMIN_PASSWORD=admin123 docker compose -f docker-compose.dev.yml build sub2api`
- manual browser verification completed locally by the reporter

## Demo

screen recording

https://github.com/user-attachments/assets/a6cbe22f-ee27-424f-b3d2-ec98264f64c4

### DIFF
--- a/frontend/src/components/common/Select.vue
+++ b/frontend/src/components/common/Select.vue
@@ -140,6 +140,7 @@ interface Props {
 interface Emits {
   (e: 'update:modelValue', value: string | number | boolean | null): void
   (e: 'change', value: string | number | boolean | null, option: SelectOption | null): void
+  (e: 'search', query: string): void
 }
 
 const props = withDefaults(defineProps<Props>(), {
@@ -329,6 +330,7 @@ watch(isOpen, (open) => {
     }
 
     if (props.searchable) {
+      emit('search', searchQuery.value)
       nextTick(() => searchInputRef.value?.focus())
     }
     // Add scroll listener to update position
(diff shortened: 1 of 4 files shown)


hansnow:
Closing this in favor of #2770. The newer PR is based on the current main branch, avoids the merge conflict here, and carries the same usage API key filter fix with explicit remote-search support, loading state handling, selected-option preservation, and stale-response guarding.

PR #1772, opened 2026-04-21
[codex] reconcile OpenAI admin test rate-limit state

### LINKED ISSUE #1770: [Bug] OpenAI OAuth account error state can become stale after upstream failure mode changes
## Summary

In production, a batch of OpenAI OAuth accounts was marked as:

- `status=error`
- `error_message = Access forbidden (403): account may be suspended or lack permissions`

But later, manual account tests showed that most of these accounts were no longer returning `403`.

After retesting the affected accounts:
- 11 out of 14 accounts recovered successfully and were switched back to `active`
- the remaining 3 accounts returned `429 usage_limit_reached` instead of `403`

This suggests the persisted account error state can become stale and no longer reflect the real upstream condition.

## Observed behavior

A production batch of OpenAI OAuth accounts entered `error` state after a burst of `403` auth-like failures.

Later, manual `/api/v1/admin/accounts/:id/test` runs showed:
- most accounts could pass the test and auto-recover
- some accounts were actually rate-limited (`429`) rather than still forbidden (`403`)

So the admin/account state and the real upstream result drifted apart.

## Why this is a problem

This can mislead operators into thinking a large number of accounts are permanently broken, even though many are actually recoverable or already healthy again.

It also means operational decisions are made based on stale persisted state rather than current upstream behavior.

## Expected behavior

If a later probe/test shows that the upstream behavior has changed, the account state should be reconciled accordingly instead of continuing to display an outdated `403` error.

## Relevant code

- `backend/internal/service/account_test_service.go`
- `backend/internal/service/ratelimit_service.go`
- `backend/internal/handler/admin/account_handler.go`

## Proposed direction

We intend to help fix this by improving state reconciliation for OpenAI OAuth accounts, so stale persisted `403` errors do not survive after the upstream condition has already changed.

### LINKED ISSUE #1771: [Bug] OpenAI admin account test does not persist 429 rate-limit state, leaving stale 403 error data
## Summary

The OpenAI admin account test path can return `429 usage_limit_reached`, but that result is not persisted back into account runtime state in the same way as the normal rate-limit handling flow.

In practice, this can leave an account with stale persisted `403` error data in the database, while live retesting already shows `429`.

## Observed behavior

For some OpenAI OAuth accounts:

- persisted DB state remained:
  - `status=error`
  - `error_message='Access forbidden (403): account may be suspended or lack permissions'`

But live `/api/v1/admin/accounts/:id/test` returned:
- `429 usage_limit_reached`

This creates visible state drift between the real upstream result and the stored account status.

## Root cause

In `testOpenAIAccountConnection`, non-200 responses are returned directly to the UI.

Currently:
- `401` explicitly persists an account error
- `429` is returned to the caller, but does not update account runtime state through the rate-limit persistence path in this flow

Meanwhile, the rate-limit service already has logic to parse and persist OpenAI `429` reset times.

## Expected behavior

When the admin test path receives an OpenAI `429 usage_limit_reached`, it should persist rate-limit state consistently, including fields such as:
- `rate_limited_at`
- `rate_limit_reset_at`

and it should avoid leaving stale `403` error data behind.

## Relevant code

- `backend/internal/service/account_test_service.go`
- `backend/internal/service/ratelimit_service.go`

## Proposed direction

We intend to help fix this by wiring the OpenAI admin test path into the existing rate-limit persistence/reconciliation flow, so retest results update runtime state instead of leaving stale error status in place.

KnowSky404:
## What changed

- reconcile OpenAI admin test `429 usage_limit_reached` results into persisted account runtime state
- persist rate-limit state during the admin test path instead of only returning the error to the UI
- clear stale `status=error` / old `403` error message when a retest shows the account is actually rate-limited instead of forbidden
- add unit coverage for the OpenAI admin test reconciliation path

## Why

We reproduced a production state drift where some OpenAI OAuth accounts stayed persisted as:

- `status=error`
- `error_message = Access forbidden (403): account may be suspended or lack permissions`

but live admin retesting later returned `429 usage_limit_reached` instead.

That left stale `403` state in the database even though the real upstream condition had already changed.

## Root cause

`testOpenAIAccountConnection` only persisted permanent error state for `401` responses.

For `429`, the admin test path returned the error to the caller and persisted codex snapshot headers, but it did not reuse the existing rate-limit state handling path. As a result:

- `rate_limited_at` / `rate_limit_reset_at` were not updated
- stale `status=error` and old `403` error messages could remain in place

## User impact

This makes the admin UI and stored account state align better with the actual upstream result during manual retests:

- stale `403` error state no longer survives when the current result is `429`
- operators see the account as rate-limited rather than permanently broken
- successful retests continue to use the existing recovery flow unchanged

## Validation

- `GOCACHE=/tmp/go-build GOMODCACHE=/tmp/go-mod go test -tags=unit ./internal/service -run 'Test(AccountTestService_OpenAI|CalculateOpenAI429ResetTime|Handle429_OpenAIPersistsCodexSnapshotImmediately|RateLimitService_RecoverAccount(AfterSuccessfulTest|State))'`

## Related issues

Fixes #1770
Fixes #1771

### DIFF
--- a/backend/internal/service/account_test_service.go
+++ b/backend/internal/service/account_test_service.go
@@ -538,6 +538,9 @@ func (s *AccountTestService) testOpenAIAccountConnection(c *gin.Context, account
 
 	if resp.StatusCode != http.StatusOK {
 		body, _ := io.ReadAll(resp.Body)
+		if resp.StatusCode == http.StatusTooManyRequests {
+			s.reconcileOpenAI429State(ctx, account, resp.Header, body)
+		}
 		// 401 Unauthorized: 标记账号为永久错误
 		if resp.StatusCode == http.StatusUnauthorized && s.accountRepo != nil {
 			errMsg := fmt.Sprintf("Authentication failed (401): %s", string(body))
@@ -548,6 +551,39 @@ func (s *AccountTestService) testOpenAIAccountConnection(c *gin.Context, account
 
 	// Process SSE stream
 	return s.processOpenAIStream(c, resp.Body)
+}
+
+func (s *AccountTestService) reconcileOpenAI429State(ctx context.Context, account *Account, headers http.Header, body []byte) {
+	if s == nil || s.accountRepo == nil || account == nil {
+		return
+	}
+
+	var resetAt *time.Time
+	if calculated := calculateOpenAI429ResetTime(headers); calculated != nil {
+		resetAt = calculated
+	} else if unixTs := parseOpenAIRateLimitResetTime(body); unixTs != nil {
+		t := time.Unix(*unixTs, 0)
+		resetAt = &t
+	}
+	if resetAt == nil {
+		return
+	}
+
+	if err := s.accountRepo.SetRateLimited(ctx, account.ID, *resetAt); err != nil {
+		return
+	}
+
+	now := time.Now()
+	account.RateLimitedAt = &now
+	account.RateLimitResetAt = resetAt
+
+	if account.Status == StatusError {
+		if err := s.accountRepo.ClearError(ctx, account.ID); err != nil {
+			return
+		}
+		account.Status = StatusActive
+		account.ErrorMessage = ""
+	}
 }
 
 // testGeminiAccountConnection tests a Gemini account's connection
--- a/backend/internal/service/account_test_service_openai_test.go
+++ b/backend/internal/service/account_test_service_openai_test.go
@@ -61,10 +61,12 @@ func newTestContext() (*gin.Context, *httptest.ResponseRecorder) {
 
 type openAIAccountTestRepo struct {
 	mockAccountRepoForGemini
-	updatedExtra map[string]any
-	rateLimitedID int64
-	rateLimitedAt *time.Time
+	updatedExtra   map[string]any
+	rateLimitedID  int64
+	rateLimitedAt  *time.Time
 	clearedErrorID int64
+	setErrorID     int64
+	setErrorMsg    string
 }
 
 func (r *openAIAccountTestRepo) UpdateExtra(_ context.Context, _ int64, updates map[string]any) error {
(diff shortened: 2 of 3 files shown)


KnowSky404:
Added another regression-test pass for the admin OpenAI account test path so the state transitions are covered more explicitly.\n\nNew test coverage now includes:\n-  with codex headers -> persists rate-limit state and clears stale  error state\n-  with only body  -> still persists rate-limit state and clears stale error state\n-  on an already  account -> persists rate-limit state without unnecessary error clearing\n-  without any reset signal -> does not mutate runtime state\n-  -> still keeps the permanent-auth-error behavior and does not mark rate-limited\n\nThis was added mainly because the production symptom is difficult to verify manually and we wanted the state reconciliation behavior locked down with focused unit tests.

KnowSky404:
Added another regression-test pass for the admin OpenAI account test path so the state transitions are covered more explicitly.

New test coverage now includes:
- `429` with codex headers -> persists rate-limit state and clears stale `403` error state
- `429` with only body `resets_at` -> still persists rate-limit state and clears stale error state
- `429` on an already `active` account -> persists rate-limit state without unnecessary error clearing
- `429` without any reset signal -> does not mutate runtime state
- `401` -> still keeps the permanent-auth-error behavior and does not mark rate-limited

This was added mainly because the production symptom is difficult to verify manually and we wanted the state reconciliation behavior locked down with focused unit tests.

Wei-Shaw:
rebase下 @KnowSky404

KnowSky404:

KnowSky404:
已 rebase 到最新的 `main`。

rebase 后顺手修了一处测试签名不兼容的问题，本地 `make test-unit` 已通过，PR checks 也已经重新触发；目前除了最后一个 `test` 还在跑，其他检查都已经通过了。

KnowSky404:
@Wei-Shaw 有时间帮忙再review一下哈

PR #1793, opened 2026-04-22
fix: stabilize local e2e entrypoints

josh87523:
## Summary
- add a checked-in e2e test entry script and route Make targets through it
- lower default Go compile parallelism to avoid local OOM kills
- skip integration cases consistently when the target service is unreachable

## Verification
- `PATH="/tmp/go-toolchains/go/bin:$PATH" ./scripts/e2e-test.sh`
- `PATH="/tmp/go-toolchains/go/bin:$PATH" /tmp/go-toolchains/go/bin/go test -tags=e2e -list . ./internal/integration/...`

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build generate test test-unit test-integration test-e2e
+.PHONY: build generate test test-unit test-integration test-e2e test-e2e-local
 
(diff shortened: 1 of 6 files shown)


josh87523:
Closing and archiving this branch instead of merging. The work is kept for reference, but we do not want to continue shipping or promoting this sub2api path.

PR #1844, opened 2026-04-23
fix(oauth): prefer compat_email over synthetic email in pending choice response

evlic:
## Problem

When a new user signs up via third-party OAuth (OIDC/LinuxDo/WeChat) and no existing account matches the provider email (`compat_email`), the completion response returns a synthetic email (e.g. `oidc-xxx@oidc-connect.invalid`) in both the `email` and `resolved_email` fields.

The frontend pre-fills the create-account form with this synthetic address. If the user submits without manually changing it, the backend's `isReservedEmail()` check rejects it with **"email is reserved"**.

This affects all three OAuth providers (OIDC, LinuxDo, WeChat) with the same root cause.

## Scenario

1. User initiates OAuth login via OIDC provider (e.g. Feishu/Lark)
2. Backend receives callback, generates synthetic email from issuer+subject hash
3. Provider returns a real email as `compat_email`
4. No existing account matches `compat_email` → `compatEmailUser` is nil
5. Backend sets `email` and `resolved_email` to the synthetic address
6. Frontend receives `choose_account_action_required` step
7. User clicks "Create New Account" → form pre-fills with synthetic email
8. Submit → `isReservedEmail()` rejects → **"email is reserved"**

## Root Cause

In `createXxxOAuthChoicePendingSession()`, when `compatEmailUser == nil` (new user, no existing account match), the `email` and `resolved_email` fields default to the synthetic email. The `compat_email` field is returned as supplementary data but never used to override the primary email fields.

The frontend's `extractPendingAccountEmail()` fallback chain does not include `compat_email`, so it picks up `resolved_email` (synthetic) instead.

## Fix

In all three OAuth handlers (`auth_oidc_oauth.go`, `auth_linuxdo_oauth.go`, `auth_wechat_oauth.go`):

When `compat_email` is non-empty and `compatEmailUser` is nil, override `suggestionEmail` and `canonicalEmail` with `compat_email` before building the completion response. This ensures the frontend receives a usable real email address for pre-filling the create-account form.

## Changes

- `backend/internal/handler/auth_oidc_oauth.go`
- `backend/internal/handler/auth_linuxdo_oauth.go`
- `backend/internal/handler/auth_wechat_oauth.go`

All three files receive the same minimal fix: 6 lines added per file.

## Testing

- Go build: ✅
- Handler tests (66 tests): ✅

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,0 +1,50 @@
+name: Build & Push Docker Image
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set lowercase owner
+        id: lowercase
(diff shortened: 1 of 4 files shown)


evlic:

Wei-Shaw:
.github/workflows/build-image.yml 这个移除一下

PR #1846, opened 2026-04-23
fix: compatible with non-standard SSE format (no space after colon) from some Anthropic-compatible upstreams

s0rcererlee:
## What this PR does

Fixes SSE parsing compatibility for upstreams that return `event:name` / `data:{...}` format without a space after the colon, while maintaining full compatibility with the standard `event: name` / `data: {...}` format.

## Problem

Some Anthropic-compatible API providers (e.g. Alibaba Cloud DashScope / 百炼) return SSE events without a space after the colon:

```
event:message_start
data:{"message":{"usage":{"input_tokens":2,"output_tokens":0}},"type":"message_start"}
```

The existing SSE parsers in `handleCCBufferedFromAnthropic`, `handleCCStreamingFromAnthropic`, `handleResponsesBufferedStreamingResponse`, and `handleResponsesStreamingResponse` use `strings.HasPrefix(line, "event: ")` which requires a space after the colon. This causes **all SSE events to be silently skipped**, resulting in:

- `ForwardResult.Usage` stays at zero values
- Token counts (input/output/cache) are all recorded as 0
- Billing fails silently — users make requests but no usage is recorded or charged

## Solution

Added helper functions (`parseCCeventLine`/`parseCCdataLine` and `parseRespEventLine`/`parseRespDataLine`) that accept both formats:
- `event: name` (standard, with space) ← checked first
- `event:name` (non-standard, without space) ← fallback

Also added `event.Type = eventType` fallback to ensure the event type is always populated even when the JSON payload doesn't contain a `type` field.

## Files changed

| File | Changes |
|------|---------|
| `backend/internal/service/gateway_forward_as_chat_completions.go` | 2 functions modified + 2 helpers added |
| `backend/internal/service/gateway_forward_as_responses.go` | 2 functions modified + 2 helpers added |

## Scope

- 4 SSE parsing locations updated
- Fully backward compatible — standard format is checked first
- No changes to business logic, only SSE line prefix parsing

## Tested

Verified with Alibaba Cloud DashScope (`coding.dashscope.aliyuncs.com/apps/anthropic`) using `qwen3.6-plus` model:
- Before fix: `input_tokens=0, output_tokens=0` in `usage_logs`
- After fix: `input_tokens=33893, output_tokens=257` correctly recorded

### DIFF
--- a/backend/internal/service/gateway_forward_as_chat_completions.go
+++ b/backend/internal/service/gateway_forward_as_chat_completions.go
@@ -230,7 +230,8 @@ func (s *GatewayService) handleCCBufferedFromAnthropic(
 
 	for scanner.Scan() {
 		line := scanner.Text()
-		if !strings.HasPrefix(line, "event: ") {
+		eventType, ok := parseCCeventLine(line)
+		if !ok {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


s0rcererlee:

github-actions[bot] ***@***.***> 于2026年4月23日周四 18:19写道：

PR #1889, opened 2026-04-24
fix: resolve unaligned 64-bit atomic operation on 32-bit architectures

whyuds:
This PR fixes a critical panic caused by unaligned 64-bit atomic operations on 32-bit architectures (like x86).

Reordered `upstreamClientEntry` fields to place `int64` variables at the top of the struct, ensuring the 8-byte alignment required by `sync/atomic` on 32-bit systems.


whyuds:

whyuds:
recheck

PR #1907, opened 2026-04-24
Fix sticky session fallback caused by missing GroupIDs in scheduler metadata

lfshao:
## Problem

  Sticky session routing could unexpectedly switch accounts even when the client
  session id was stable.

  The issue happened when the sticky-hit account object came from scheduler
  metadata/cache:
  - scheduler metadata did not include `GroupIDs`
  - `isAccountInGroup` only checked `AccountGroups`
  - the sticky-hit account could be incorrectly treated as outside the current group
  - routing then fell back to another eligible duplicate account and rewrote sticky binding

  This was especially visible in setups where multiple local accounts represented
  the same upstream Claude account.

  ## Root Cause

  Two pieces combined to cause the bug:

  1. `buildSchedulerMetadataAccount(...)` dropped `GroupIDs` when writing
     scheduler metadata to Redis.
  2. `isAccountInGroup(...)` only checked `AccountGroups`, which may be empty on
     runtime snapshot objects.

  As a result, cached/snapshotted accounts could lose effective group membership
  during sticky-session validation.

  ## Changes

  - Preserve `GroupIDs` in scheduler metadata snapshots.
  - Update `isAccountInGroup(...)` to fall back to `GroupIDs` when
    `AccountGroups` is empty.

  ## Tests

  Added focused unit coverage for:
  - `isAccountInGroup(...)` group-membership fallback via `GroupIDs`
  - scheduler metadata retaining `GroupIDs`

  ## Impact

  This restores correct sticky-session behavior for accounts loaded from
  scheduler/cache and prevents unnecessary fallback/rebinding caused by false
  "group mismatch" checks.


lfshao:

PR #1912, opened 2026-04-24
fix(gemini): apply model_mapping to oauth accounts

brushax:
## Summary
- apply Gemini `model_mapping` for OAuth accounts in both compat and native forwarding paths
- keep the existing API key mapping behavior unchanged
- fix OAuth-backed Gemini upstreams such as Code Assist where a client-facing model alias should be remapped before the upstream request is built

## Testing
- `docker compose build sub2api`
- manual `POST /v1beta/models/gemini-2.0-flash:generateContent` against a local deploy with an OAuth-backed Gemini account
- manual `POST /v1beta/models/gemini-2.0-flash:streamGenerateContent?alt=sse` against the same deploy


brushax:

PR #1947, opened 2026-04-25
fix(apicompat): ensure thinking/text content blocks always include their field in JSON

clansty:
## Summary

Fixes a bug where `content_block_start` events for `thinking` type blocks are serialized without the `thinking` field, causing downstream schema validation errors:

```
Type validation failed: Value: {"type":"content_block_start","index":0,"content_block":{"type":"thinking"}}.
Error message: [{"expected":"string","code":"invalid_type","path":["content_block","thinking"],"message":"Invalid input: expected string, received undefined"}]
```

## Root Cause

`AnthropicContentBlock.Thinking` (and `.Text`) use `json:",omitempty"` tags. When a `thinking` content block is created with `Thinking: ""` (as in streaming `content_block_start`), Go's `json.Marshal` omits the field entirely. Clients validating against the Anthropic API schema expect a string value and fail with "expected string, received undefined".

## Fix

Add a custom `MarshalJSON` on `AnthropicContentBlock` that ensures:
- `type=thinking` blocks always include `"thinking"` (even if empty string)
- `type=text` blocks always include `"text"` (even if empty string)
- All other block types are unaffected

## Testing

- Added serialization assertions to existing `TestStreamingReasoning` and `TestStreamingTextOnly` tests
- All existing tests pass with no regressions

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_responses_test.go
+++ b/backend/internal/pkg/apicompat/anthropic_responses_test.go
@@ -308,6 +308,13 @@ func TestStreamingTextOnly(t *testing.T) {
 	require.Len(t, events, 2) // content_block_start + content_block_delta
 	assert.Equal(t, "content_block_start", events[0].Type)
 	assert.Equal(t, "text", events[0].ContentBlock.Type)
+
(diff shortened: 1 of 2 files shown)


clansty:

PR #1953, opened 2026-04-25
fix(monitor): extract Anthropic text blocks

mark-ly-wang:
## Summary
- Extract Anthropic monitor response text from all `content[]` text blocks instead of only `content[0].text`
- Keep the existing `textPath` behavior for OpenAI and Gemini monitors
- Add unit coverage for thinking-before-text, single text, thinking-only, and multiple text block responses

## Why
Anthropic Messages responses may include a `thinking` block before the actual text block. The channel monitor previously read only `content.0.text`, which made successful responses look empty and caused challenge mismatch failures.

## Tests
- `go test -tags=unit ./internal/service -run "TestExtractAnthropicMonitorText|TestValidateChallenge_AnthropicTextAfterThinking|TestRunCheckForModel"`

### DIFF
--- a/backend/internal/service/channel_monitor_checker.go
+++ b/backend/internal/service/channel_monitor_checker.go
@@ -146,11 +146,11 @@ func pingEndpointOrigin(ctx context.Context, endpoint string) *int {
 	return &ms
 }
 
-// providerAdapter 描述某个 provider 在 challenge 检测中需要的 4 件事：
+// providerAdapter 描述某个 provider 在 challenge 检测中需要的几件事：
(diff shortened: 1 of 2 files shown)


mark-ly-wang:
recheck

mark-ly-wang:

heathermhuang:
Refreshed on current `main` in #4537. The original author attribution and cherry-pick source are preserved; the refresh resolves the conflict with the later OpenAI Responses text extractor while keeping this PR’s Anthropic behavior and tests. #4537 is intended to supersede this stale branch.

PR #1990, opened 2026-04-26
feat(httputil): decode zstd/gzip/deflate request bodies

haha1903:
## Background

Recent versions of the OpenAI Codex CLI (0.125+) default to compressing outbound request bodies with `Content-Encoding: zstd`. Other tooling (curl with `--compressed`, custom clients, Cloudflare in some configurations) similarly send `gzip` or `deflate`.

Today `ReadRequestBodyWithPrealloc` reads the raw bytes and hands them to gjson without inspecting `Content-Encoding`. The gateway then rejects the request with:

```
{"error":{"code":"invalid_request_error","message":"Failed to parse request body"}}
```

This is reproducible end-to-end: Codex 0.125 → sub2api `/v1/responses` returns 400 even though the upstream OpenAI/Copilot endpoint would have happily accepted the (compressed) body. Disabling compression on the client side via `enable_request_compression = false` is a workaround, but newer Codex builds default it on and many users won't know to flip it.

## Change

`ReadRequestBodyWithPrealloc` now:

1. Reads the body as before (preallocated buffer keyed on `Content-Length`).
2. Looks at `Content-Encoding`. If empty or `identity` → returns raw bytes (existing behavior).
3. Otherwise dispatches to `decompressRequestBody` which handles `zstd`, `gzip`/`x-gzip`, and `deflate`. Unsupported encodings (e.g. `br`) return a wrapped error so handlers surface a clear 400 instead of a confusing parse failure.
4. After decoding, strips `Content-Encoding` / `Content-Length` and updates `req.ContentLength` so any downstream code that reuses the request (forwarding, logging, hashing) sees the decoded payload consistently.

`zstd` uses `github.com/klauspost/compress/zstd` (already an indirect dependency, now promoted to direct). `gzip`/`deflate` use the stdlib.

## Tests

New file `backend/internal/pkg/httputil/body_test.go` covers:

- identity passthrough
- zstd round-trip + header/ContentLength rewrite
- gzip round-trip
- deflate (zlib) round-trip
- unsupported encoding returns descriptive error
- corrupt zstd payload surfaces decoder error
- nil body short-circuit
- explicit `Content-Encoding: identity`

```
$ go test ./internal/pkg/httputil/...
ok  	github.com/Wei-Shaw/sub2api/backend/internal/pkg/httputil
$ go build ./...
(clean)
```

## Risk / Compatibility

- Existing uncompressed traffic is unchanged: the early `enc == ""` fast path returns immediately with the raw buffer.
- Bodies that legitimately advertise an encoding the gateway can now decode are no longer 400'd.
- Memory: peak usage is `compressed + decompressed` for the duration of the call. For typical Codex / chat payloads this is well under a megabyte.
- No changes to public APIs or to handler call-sites; the function signature is unchanged.

Happy to adjust naming, error wording, or split the dependency promotion into a separate commit if preferred.

### DIFF
--- a/backend/internal/pkg/httputil/body.go
+++ b/backend/internal/pkg/httputil/body.go
@@ -2,8 +2,15 @@ package httputil
 
 import (
 	"bytes"
+	"compress/gzip"
+	"compress/zlib"
+	"errors"
+	"fmt"
 	"io"
 	"net/http"
+	"strings"
+
+	"github.com/klauspost/compress/zstd"
 )
 
 const (
@@ -11,7 +18,9 @@ const (
 	requestBodyReadMaxInitCap = 1 << 20
 )
 
-// ReadRequestBodyWithPrealloc reads request body with preallocated buffer based on content length.
+// ReadRequestBodyWithPrealloc reads request body with preallocated buffer based
+// on content length, transparently decoding any Content-Encoding the upstream
+// client used to compress the body (zstd, gzip, deflate).
 func ReadRequestBodyWithPrealloc(req *http.Request) ([]byte, error) {
 	if req == nil || req.Body == nil {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


haha1903:

PR #1998, opened 2026-04-26
fix: synthesize response when upstream lacks response.completed event

aqlkzf:
## Summary

- When sub2api is behind an OpenAI-compatible forwarder (e.g. newapi), the upstream may close the SSE stream with `[DONE]` without emitting a `response.completed` terminal event.
- The two buffered (non-streaming) handlers — Anthropic Messages compat and Chat Completions compat — previously required a terminal event to build `finalResponse`, causing a hard 502 error: *"Upstream stream ended without a terminal response event"*.
- This PR adds a fallback: when `finalResponse` is nil but the `BufferedResponseAccumulator` has accumulated content from delta events, a synthetic `ResponsesResponse` is built from those deltas instead of failing. The original error path is preserved for the case where no content was accumulated at all.

## What changed

| File | Change |
|------|--------|
| `openai_gateway_messages.go` | `handleAnthropicBufferedStreamingResponse`: fallback to `acc.BuildOutput()` when no terminal event |
| `openai_gateway_chat_completions.go` | `handleChatBufferedStreamingResponse`: same fallback |

## Why streaming paths are unaffected

The streaming handlers already call `FinalizeResponsesAnthropicStream` / `FinalizeResponsesChatStream` at stream end, which synthesize proper termination events (`message_stop`, finish chunk) even without `response.completed`. The passthrough path's `openAIStreamEventIsTerminal` also treats `[DONE]` as terminal. Only the two buffered compat paths had the hard failure.

## Test plan

- [ ] Verify GPT model requests through a newapi forwarder no longer return 502
- [ ] Verify normal OpenAI upstream (with `response.completed`) still works as before
- [ ] Verify streaming mode is unaffected


aqlkzf:




Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2021, opened 2026-04-27
feat(openai): pass through SSE comment keepalives during reasoning stalls

haha1903:
## Background

`forwardOpenAIPassthrough` collects all upstream SSE lines in a `pendingLines`
slice until it sees the first \"client output\" event. This is by design — it
hides intermediate non-output events from clients until something visible
arrives.

It also has an unintended side effect: SSE **comment lines** (lines beginning
with \`:\`) emitted by upstreams as keepalive heartbeats end up in the same
buffer and never leave the gateway during the reasoning window.

## Symptom

GPT-5 / o1-class models with high reasoning effort can spend 60–120 s on a
single reasoning step before producing the first \`response.output_text.delta\`.
During that window the upstream connection is alive and the upstream sends
\`: ping\` / \`: keepalive\` lines, but the gateway holds them in the buffer.

Any intermediary with an idle-timeout shorter than the reasoning step closes
the SSE connection. The most common one is Cloudflare's ~100 s origin idle
limit. The user-visible error is:

```
Stream disconnected before completion: stream closed before response.completed
```

The gateway logs show no error — from its perspective the upstream is still
streaming, the client just \"disconnected\" silently a minute or two in.

## Change

Bypass the pending buffer for SSE comment lines. When a line begins with
\`:\` (after optional whitespace, per the
[EventSource spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream)),
write it directly and \`flusher.Flush()\` immediately, without setting
\`clientOutputStarted\` or contributing to first-token timing.

Comment lines are explicitly defined as no-ops by the spec, so this is safe
for any compliant SSE client. But every byte that hits the wire resets idle
timers on Cloudflare, nginx, AWS ALB, and similar middleboxes.

## Tests

New unit test \`TestIsSSECommentLine\` covers the helper across spaces, tabs,
empty input, and non-comment lines.

```
$ go test -run TestIsSSECommentLine ./internal/service/...
ok  	github.com/Wei-Shaw/sub2api/internal/service
$ go build ./...
(clean)
```

## Risk / Compatibility

- No change for streams that don't contain comment lines (the new branch is
  guarded by a single \`HasPrefix\` check).
- No change to \`firstTokenMs\` accounting — comment lines are explicitly
  excluded from \"client output\" detection.
- No change to failure semantics. \`response.failed\`, \`[DONE]\`, terminal-event
  detection, and \`pendingLines\` flushing on first real output are untouched.
- Memory: removes lines from the buffer immediately, so peak buffer size only
  decreases.

Happy to adjust the helper name or split the test if preferred.

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -3375,6 +3375,20 @@ func (s *OpenAIGatewayService) handleStreamingResponsePassthrough(
 		}
 
 		if !clientDisconnected {
+			// SSE comment lines (":...") are no-ops per spec but reset idle
+			// timers on intermediaries (e.g. Cloudflare's ~100s limit). Pass
+			// them through immediately, even before client output has started,
+			// so upstream keepalives during long reasoning stalls actually
+			// reach the client without polluting the pending buffer.
+			if isSSECommentLine(line) {
+				if _, err := fmt.Fprintln(w, line); err != nil {
+					clientDisconnected = true
+					logger.LegacyPrintf("service.openai_gateway", "[OpenAI passthrough] Client disconnected during streaming, continue draining upstream for usage: account=%d", account.ID)
+				} else {
+					flusher.Flush()
+				}
+				continue
+			}
 			if !clientOutputStarted && !lineStartsClientOutput {
 				pendingLines = append(pendingLines, line)
 				continue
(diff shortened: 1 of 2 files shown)


haha1903:

gaoren002:
实际上是PR#1960的进一步补充，不过#1960对下游处理，这里处理上游渠道

haha1903:
@gaoren002 感谢补充，理解得很准确。

两个 PR 确实是互补的：

- **#1960（下游方向）**：上游只有被暂存 preamble 事件（`response.created` / `response.in_progress`）时，网关基于下游空闲时间**主动发**自定义 keepalive，让下游持续有字节写出。
- **本 PR（上游方向）**：上游主动发来的 SSE 注释行（`: ping` / `: keepalive`，按 EventSource 规范本身就是 no-op）不再塞进 `pendingLines` 等第一个可见事件，而是立即透传 flush。

叠加之后，无论上游是"只发 preamble"还是"发 `:` 注释行"，下游都会持续有字节、Cloudflare/nginx 等中间层的 idle timer 会被重置。GPT-5 高 reasoning effort 下 60–120s 的单步思考场景中两种形式都会遇到。

实现上刻意只匹配 `:` 前缀（按 EventSource 规范允许可选前导空白），并且不触发 `clientOutputStarted`、不参与 `firstTokenMs` 统计，完全保留原有首字节计时与 pending 缓冲语义，风险面很小。

PR #2027, opened 2026-04-27
fix(api-key): reset rate limit usage cache

hansnow:
## Summary

Fix API key daily rate-limit reset so it takes effect immediately from both user and admin paths.

Root causes:
- `APIKeyService.Update(... reset_rate_limit_usage=true)` reset DB counters, but its rate-limit cache invalidator was never wired, so gateway billing checks could continue reading stale `apikey:rate:{keyID}` Redis usage. Raising `rate_limit_1d` appeared to unblock requests because the stale usage was then below the higher limit.
- Admin `PUT /api/v1/admin/api-keys/:id` only handled `group_id`; `reset_rate_limit_usage` was accepted by JSON decoding but silently ignored.

Changes:
- Wire `APIKeyService` with `BillingCacheService` as the `RateLimitCacheInvalidator`.
- Add `BillingCacheService.InvalidateAPIKeyRateLimit`.
- Add admin reset handling for `reset_rate_limit_usage` and invalidate auth + rate-limit caches.
- Add regression coverage for admin reset behavior.

## Verification

- `go test ./internal/handler/admin -run 'TestAdminAPIKeyHandler'`\n- `go test ./internal/service -run 'Test(APIKeyService|BillingCacheService)'`\n- `go test ./cmd/server`


hansnow:

PR #2034, opened 2026-04-27
 fix(anthropic): map system prompt to responses instructions

ma0608:
## Summary

  - Map Anthropic system prompts to the Responses API top-level `instructions` field instead of injecting them as input items
  - Keep Responses input items limited to user/assistant message content
  - Add a default coding-assistant instruction when forwarding OpenAI messages without explicit instructions

  ## Test plan

  - Updated Anthropic-to-Responses system prompt tests
  - Not run locally


ma0608:

Issue #1957, opened 2026-04-25
gpt-5.5 fails via /v1/responses


stefaneliu:
### Describe the bug

`gpt-5.5` works correctly via `/v1/chat/completions`, but fails via `/v1/responses` in real multi-turn / streaming usage.

The `/v1/responses` path appears to have an issue with item persistence / continuation:

- upstream returns:
  - `Item with id 'rs_...' not found`
  - `Items are not persisted when store is set to false`
- sub2api then returns `502 Upstream request failed`

This affects both:
- OpenClaw using `api: "openai-responses"`
- a separate Python project using OpenAI SDK `client.responses.stream(...)`

So this does **not** appear to be specific to OpenClaw.

---

### Environment

- sub2api version: `0.1.117`
- build commit: `a4e329c18b133e950f9345a27bfbd76624519759`
- deployment: Docker
- upstream type: OpenAI OAuth accounts
- model: `gpt-5.5`

---

#### 1. `/v1/chat/completions` works fine with `gpt-5.5`
Tested successfully with:

- non-stream
- stream
- multi-turn by resending full `messages` context each turn

#### 2. `/v1/responses` single-turn can also return 200
But real usage fails once response/item continuation behavior is involved.

---

#### A. `/v1/responses` in real app usage
In both OpenClaw and a separate Python project using `client.responses.stream(...)`, `gpt-5.5` fails with:

- `502 Upstream request failed`

sub2api logs show upstream error:

```text
Item with id 'rs_...' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.
```

#### B. `store=true` appears to be ignored on `/v1/responses`
I tested direct HTTP requests to sub2api:

```json
{
  "model": "gpt-5.5",
  "store": true,
  "input": "Reply with exactly: OK55S"
}
```

The request returns `200`, but the response body still contains:

```json
"store": false
```

So it looks like `store=true` is not actually preserved / forwarded.

#### C. `previous_response_id` is rejected on HTTP `/v1/responses`
Testing a 2-step chain with `previous_response_id` returns:

```json
{
  "error": {
    "message": "previous_response_id is only supported on Responses WebSocket v2",
    "type": "invalid_request_error"
  }
}
```

This suggests the current HTTP `/v1/responses` path does not fully support continuation semantics expected by some clients.

---

#### 1. `/v1/chat/completions` succeeds

`gpt-5.5`, non-stream:

```bash
[...]

PR #2068, opened 2026-04-28
fix(openai): drop reasoning items from /v1/responses input on OAuth path

### LINKED ISSUE #2067: OAuth /v1/responses: replayed reasoning items (rs_*) cause 502 on every multi-turn request with tools
## Summary

When a client follows the OpenAI Responses API spec and replays the previous turn's assistant items in `input[]` (the standard pattern for multi-turn conversations), every turn ≥ 2 against `/v1/responses` on an OAuth account fails with **502 "Upstream request failed"**. The upstream returns a 404 with a very specific error:

> Item with id 'rs_...' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.

This affects any client built against the standard Responses API contract — confirmed with [OpenClaw](https://github.com/some/openclaw) and reproducible with a 30-line Python script using nothing but `urllib`.

This is related to but distinct from #1957 — that report focuses on `previous_response_id` rejection on HTTP; this one focuses on the more common pattern where clients replay `output[]` items inline (which happens automatically for any reasoning-capable model).

## Reproduction

Two sequential `/v1/responses` requests with `tools` declared, on `gpt-5.5`. Reproducer attached at the end.

**Turn 1** (succeeds): client sends a user message + tool definitions. The upstream stream emits, among other items:

```json
{"type": "response.output_item.added",
 "item": {"type": "reasoning", "id": "rs_0672f12450da0b9c0169f07220a6c08198b68c2455ced99344", "summary": []}}
{"type": "response.output_item.added",
 "item": {"type": "function_call", "call_id": "call_xxx", "name": "record"}}
```

The client persists those items in its conversation history, exactly as the OpenAI Responses API spec implies it should.

**Turn 2** (fails with 502): client sends the next user message together with the prior turn's items in `input[]`:

```json
{
  "model": "gpt-5.5",
  "tools": [...],
  "input": [
    {"role": "user", "content": "..."},
    {"type": "reasoning", "id": "rs_0672f12450da0b9c0169f07220a6c08198b68c2455ced99344"},
    {"type": "function_call", "call_id": "call_xxx"},
    {"type": "function_call_output", "call_id": "call_xxx", "output": "..."},
    {"role": "user", "content": "follow-up"}
  ]
}
```

`sub2api`'s admin `request-errors` endpoint logs the upstream 404 with the exact same `rs_*` ID the client received in Turn 1. End-to-end correlation is unambiguous.

## Why this happens

`applyCodexOAuthTransform` (in `backend/internal/service/openai_codex_transform.go`) **forces `store: false`** for every OAuth account, which is correct — the OAuth path routes to `chatgpt.com/backend-api/codex/responses`, where `store: true` is rejected outright.

But under `store: false`, the upstream **does not persist** any of the items it emits. The `rs_*` IDs are effectively ephemeral. When the client legitimately replays them on the next turn, `chatgpt.com` looks up the ID, finds nothing, and returns 404. `sub2api` then wraps that as 502.

`filterCodexInput` already strips `id` fields from non-tool items when `preserveReferences == false`, but the moment any client declares `tools` in the request (which happens on every turn of any agent / coding workflow), `preserveReferences` becomes `true` and the `rs_*` reference goes through unchanged.

## Why gpt-5.4 doesn't trigger this (interesting note)

Empirically verified by sending identical requests to both models (single env var changed):

| model | upstream emits `output_item.added` types | `rs_*` IDs in stream |
|---|---|---|
| `gpt-5.4` | `message` + `function_call` | 0 |
| `gpt-5.5` | `reasoning` + `message` + `function_call` | 1 |

`chatgpt.com`'s codex backend simply doesn't generate reasoning items for `gpt-5.4` in its SSE stream. So clients have nothing to replay, and the bug never fires for that model. This made the issue look model-specific for a while, but it is fully explained by upstream emission behavior.

For 5.5 (and presumably any future reasoning-heavy model on the same path), the bug is deterministic — every multi-turn-with-tools request fails on turn 2 onward.

## Impact

- Any client following the standard OpenAI Responses API multi-turn contract is broken on OAuth accounts with `gpt-5.5` (and any model that emits reasoning items via this upstream).
- Workarounds in the client (strip `rs_*` items before sending) are possible but require every client to re-implement the same defense. The fix belongs in the gateway's protocol-adapter layer.
- API-key accounts on `api.openai.com/v1/responses` are unaffected — `store: true` works there, items persist, multi-turn replay works as designed.

## Suggested direction

In `filterCodexInput`, items with `type == "reasoning"` should be dropped from `input[]` regardless of `preserveReferences`. The model never reads reasoning summary text from input (only `encrypted_content` can carry reasoning context across turns, and `chatgpt.com` under `store: false` does not emit it), so this is a no-op for the model and a clean removal of unreachable upstream lookups.

I have a PR ready that does exactly this with a regression test — will link it as soon as it's open.

## Reproducer

Minimal end-to-end reproducer (Python 3, no dependencies). Replace `BASE` and `KEY` and run:

````python
import json, urllib.request, urllib.error

BASE = "https://your.sub2api.host"
KEY  = "sk-..."  # any sub2api key on a group containing an OAuth OpenAI account
UA   = "Mozilla/5.0 (compat; rs-repro)"

TOOLS = [{
    "type": "function", "name": "record", "description": "r",
    "parameters": {
        "type": "object",
        "properties": {"items": {
            "type": "array",
            "items": {"type": "object",
                "properties": {"n": {"type": "integer"}, "prime": {"type": "boolean"}},
                "required": ["n", "prime"], "additionalProperties": False}}},
        "required": ["items"], "additionalProperties": False}}]

PROMPT = ("Three numbers: 17, 23, 41. Decide prime/composite for each, "
          "then call the record tool with the results.")

def stream(body):
    req = urllib.request.Request(BASE + "/v1/responses",
        data=json.dumps(body).encode(), method="POST",
        headers={"Authorization": "Bearer " + KEY,
                 "Content-Type": "application/json",
                 "Accept": "text/event-stream",
                 "User-Agent": UA})
    items = []
    try:
        with urllib.request.urlopen(req, timeout=120) as r:
            ev = None; buf = []
            for line in r:
                line = line.decode("utf-8", "replace").rstrip("\r\n")
                if line.startswith("event:"): ev = line[6:].strip()
                elif line.startswith("data:"): buf.append(line[5:].lstrip())
                elif line == "" and buf:
                    d = json.loads("\n".join(buf)); buf = []
                    if ev == "response.output_item.done" and isinstance(d.get("item"), dict):
                        items.append(d["item"])
                    ev = None
        return items, None
    except urllib.error.HTTPError as e:
        return [], (e.code, e.read()[:400].decode("utf-8", "replace"))

# Turn 1
items, err = stream({"model": "gpt-5.5", "tools": TOOLS, "stream": True,
                     "input": [{"role": "user", "content": PROMPT}]})
assert err is None, err
print("turn1 items:", [it["type"] for it in items])

# Turn 2 — replay items as any spec-following client would
next_input = [{"role": "user", "content": PROMPT}]
for it in items:
    if it["type"] in ("reasoning", "function_call"):
        next_input.append(it)
for it in items:
    if it["type"] == "function_call":
        next_input.append({"type": "function_call_output",
                           "call_id": it["call_id"], "output": "{}"})
next_input.append({"role": "user", "content": "summarize in one line"})

_, err2 = stream({"model": "gpt-5.5", "tools": TOOLS,
                  "stream": True, "input": next_input})
print("turn2 err:", err2)   # expect: (502, "Upstream request failed")
````

## Environment

- sub2api: `0.1.119` (also reproduced on Docker image `weishaw/sub2api:latest` at filing time)
- Upstream: `chatgpt.com/backend-api/codex/responses` via OAuth
- Confirmed against multiple OAuth accounts; account choice is irrelevant since `store=false` is forced for all of them

Ogannesson:
## Summary

Drops items with `type == "reasoning"` from `input[]` in `filterCodexInput`. The OAuth path forces `store: false` upstream, which means `rs_*` IDs are never persisted and replaying them in subsequent turns guarantees a 404 → 502.

Closes #2067
Refs #1957

## What changes

| File | Lines | Purpose |
|---|---|---|
| `backend/internal/service/openai_codex_transform.go` | +8 (5 code, 4 doc) | Drop `type == "reasoning"` items at the top of `filterCodexInput`'s loop |
| `backend/internal/service/openai_codex_transform_test.go` | +53 | Regression test covering both `preserveReferences=true` and `preserveReferences=false` |

Total: +61 / -0 across 2 files.

## Why this is safe

The change is gated by the existing `applyCodexOAuthTransform` call site (`openai_gateway_service.go:2233-2245`), which only fires for `account.Type == AccountTypeOAuth`. Other paths are unaffected:

- **API-key OpenAI accounts** (going to `api.openai.com/v1/responses`): not affected — `store: true` works there and `rs_*` references must be preserved.
- **Anthropic / Gemini platform groups**: different transform pipeline, not touched.
- **`/v1/chat/completions`**: no concept of reasoning items, not touched.
- **`item_reference` items**: different `type`, not touched. Only `type == "reasoning"` is dropped.

## Why dropping is a no-op for the model

Reasoning items in `input[]` carry semantic value to the upstream model only via `encrypted_content` (the opaque OpenAI reasoning state). Under `store: false`, `chatgpt.com`'s codex backend does not emit `encrypted_content`, so the items being dropped have no model-visible payload — only the `rs_*` ID, which the upstream then fails to look up. Codex CLI, the reference client for this path, also never replays reasoning items.

## Verification

Tests run on a clean checkout of the branch:

- `go test ./internal/service/ -run "Codex|Tool|OAuth"` — full pre-existing suite passes
- New test `TestFilterCodexInput_DropsReasoningItemsRegardlessOfPreserveReferences` (with both sub-tests) passes
- End-to-end: built `weishaw/sub2api:0.1.119-patch1`, swapped into a real deployment, ran a 30-line Python reproducer:
  - Pre-patch: `gpt-5.5` turn-2-with-tools → 502 (with the exact `rs_*` ID surfacing in `request-errors`)
  - Post-patch: same reproducer → 200, three-turn deep-loop variant also 200
  - `gpt-5.4` multi-turn unchanged
  - `/v1/chat/completions` unchanged
- Patched image has been running in production for several hours under real OpenClaw load with no regressions observed

## 概述

在 `filterCodexInput` 中把 `type == "reasoning"` 的 `input[]` items 整个 drop 掉。OAuth 路径下 `store: false` 是强制的，意味着 `rs_*` ID 永远不会在上游持久化，客户端在后续轮次回放它们必然导致上游 404 → 网关 502。

Closes #2067
Refs #1957

## 改了什么

| 文件 | 行数 | 用途 |
|---|---|---|
| `backend/internal/service/openai_codex_transform.go` | +8（5 行代码 + 4 行注释）| 在 `filterCodexInput` 循环顶部 drop `type == "reasoning"` items |
| `backend/internal/service/openai_codex_transform_test.go` | +53 | 回归测试，覆盖 `preserveReferences=true` 和 `preserveReferences=false` 两条路径 |

合计 +61 / -0，2 个文件。

## 为什么是安全的

变更被现有的 `applyCodexOAuthTransform` 调用位置（`openai_gateway_service.go:2233-2245`）天然门控——只有 `account.Type == AccountTypeOAuth` 才会进。其他路径不受影响：

- **API-key OpenAI 账号**（走 `api.openai.com/v1/responses`）：不受影响——那条路 `store: true` 真生效，`rs_*` 引用必须保留
- **Anthropic / Gemini 平台账号**：走不同的 transform 流水线，未触碰
- **`/v1/chat/completions`**：没有 reasoning item 概念，未触碰
- **`item_reference` items**：`type` 不同，未触碰。仅 `type == "reasoning"` 被 drop

## 为什么 drop 对模型是 no-op

`input[]` 里的 reasoning item 对上游模型有语义价值的载体只有 `encrypted_content`（OpenAI 的 opaque reasoning state）。但在 `store: false` 下，`chatgpt.com` 的 codex 后端**不下发** `encrypted_content`，所以被 drop 的这些 items 实际上对模型没有可见 payload——只剩一个 `rs_*` ID，而上游必然查找失败。

Codex CLI（这条路径上 OpenAI 自家的参考客户端）本身也从不回放 reasoning items，进一步印证了这是上游协议层面的设计共识。

## 验证

在干净的 branch checkout 上跑了：

- `go test ./internal/service/ -run "Codex|Tool|OAuth"`：全套既有测试通过
- 新增测试 `TestFilterCodexInput_DropsReasoningItemsRegardlessOfPreserveReferences`（含两个 sub-test）通过
- 端到端验证：构建 `weishaw/sub2api:0.1.119-patch1`，部署到真实环境，跑 30 行 Python reproducer：
  - patch 前：`gpt-5.5` 双轮带 tools → 502（`request-errors` 里的 `rs_*` ID 跟客户端 turn1 收到的字符级一致）
  - patch 后：同 reproducer → 200，三轮深度循环变体也 200
  - `gpt-5.4` 多轮请求行为不变
  - `/v1/chat/completions` 行为不变
- 打 patch 的镜像已经在生产环境真实 OpenClaw 负载下跑了数小时，未观察到任何 regression

### DIFF
--- a/backend/internal/service/openai_codex_transform.go
+++ b/backend/internal/service/openai_codex_transform.go
@@ -853,6 +853,14 @@ func filterCodexInput(input []any, preserveReferences bool) []any {
 		}
 		typ, _ := m["type"].(string)
 
+		// chatgpt.com codex backend (OAuth path) does not persist reasoning
+		// items because applyCodexOAuthTransform forces store=false. Any rs_*
+		// reference replayed in input is guaranteed to 404 upstream
+		// ("Item with id 'rs_...' not found"). Drop reasoning items entirely.
+		if typ == "reasoning" {
+			continue
+		}
+
 		// 仅修正真正的 tool/function call 标识，避免误改普通 message/reasoning id；
 		// 若 item_reference 指向 legacy call_* 标识，则仅修正该引用本身。
 		fixCallIDPrefix := func(id string) string {
--- a/backend/internal/service/openai_codex_transform_test.go
+++ b/backend/internal/service/openai_codex_transform_test.go
@@ -1,6 +1,8 @@
 package service
 
 import (
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -1091,6 +1093,57 @@ func TestIsInstructionsEmpty(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			result := isInstructionsEmpty(tt.reqBody)
 			require.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestFilterCodexInput_DropsReasoningItemsRegardlessOfPreserveReferences(t *testing.T) {
+	// Reasoning items in input[] reference rs_* IDs that were emitted by
+	// chatgpt.com under store=false (forced by applyCodexOAuthTransform).
+	// They are never persisted upstream, so forwarding them produces a
+	// guaranteed 404 ("Item with id 'rs_...' not found"). Drop them
+	// regardless of preserveReferences. See: Wei-Shaw/sub2api issue #1957.
+
+	build := func() []any {
+		return []any{
+			map[string]any{"type": "message", "id": "msg_0", "role": "user", "content": "hi"},
+			map[string]any{
+				"type":    "reasoning",
+				"id":      "rs_0672f12450da0b9c0169f07220a6c08198b68c2455ced99344",
+				"summary": []any{},
+			},
+			map[string]any{"type": "function_call", "id": "fc_1", "call_id": "call_1", "name": "tool"},
+			map[string]any{"type": "function_call_output", "call_id": "call_1", "output": "{}"},
+		}
+	}
+
+	for _, preserve := range []bool{true, false} {
+		preserve := preserve
+		t.Run(fmt.Sprintf("preserveReferences=%v", preserve), func(t *testing.T) {
+			filtered := filterCodexInput(build(), preserve)
+
+			for _, raw := range filtered {
+				item, ok := raw.(map[string]any)
+				require.True(t, ok)
+				require.NotEqual(t, "reasoning", item["type"],
+					"reasoning items must be dropped from input on the OAuth path")
+				if id, ok := item["id"].(string); ok {
+					require.False(t, strings.HasPrefix(id, "rs_"),
+						"no item carrying an rs_* id should survive the filter")
+				}
+			}
+
+			// Sanity check: the non-reasoning items should still be present.
+			gotTypes := make(map[string]int)
+			for _, raw := range filtered {
+				item, ok := raw.(map[string]any)
+				require.True(t, ok)
+				gotTypes[item["type"].(string)]++
+			}
+			require.Equal(t, 1, gotTypes["message"])
+			require.Equal(t, 1, gotTypes["function_call"])
+			require.Equal(t, 1, gotTypes["function_call_output"])
+			require.Equal(t, 0, gotTypes["reasoning"])
 		})
 	}
 }


Ogannesson:

Ogannesson:
recheck

PR #2069, opened 2026-04-28
feat: Get models from the Base URL

anchogin:
When adding a new account, if it's a custom base URL, available models can be automatically retrieved via v1/models, and there's no need to manually input each model one by one.


anchogin:

PR #2125, opened 2026-04-30
fix(payment): default provider checkout currencies

cong91:
## Summary
- Default SePay provider/method capabilities to VND when legacy configs have no explicit `allowed_payment_currencies`, preventing checkout from falling back to ledger USD.
- Add provider setup UI for accepted checkout currencies so admins choose USD/VND/CNY/KRW instead of editing raw config JSON.
- Persist selected currencies into provider config and method limits, hydrate existing providers from saved limits, and keep SePay locked to VND.
- Update user checkout currency selection to respect provider/method `allowed_payment_currencies` from API data.

## Verification
- `go test -tags unit ./internal/payment ./internal/service`
- `npm run typecheck`
- `npm run lint:check`
- `npm run build`
- `git diff --check`

## Notes
- Build still reports existing Vite warnings about dynamic/static imports and chunk size; build succeeds.


cong91:
Closing because this PR was opened against upstream by mistake. Re-opening against origin fork instead.

PR #2126, opened 2026-04-30
fix(frontend): remove build warnings

cong91:
## Summary
- Remove Vite dynamic/static import warnings by replacing locale-change dynamic imports with a browser event handled by the router.
- Reuse router title logic through a single `updateDocumentTitle` helper so language switches still refresh the active page title.
- Lazy-load heavy Accounts admin modals so the Accounts route chunk drops below Vite's warning threshold instead of hiding the warning.

## Verification
- `npm run typecheck`
- `npm run build` — verified no `(!)` Vite warnings and no chunk-size warning in `/tmp/sub2api-build-warning-fix.log`
- `npm run lint:check`
- `npm run test:run -- src/router/__tests__/title.spec.ts`
- `git diff --check`

## Notes
- Backend unchanged.


cong91:
Closing because this PR was opened against upstream by mistake. Re-opening against origin fork instead.

PR #2227, opened 2026-05-06
feat: add account usage statistics view

specialpointcentral:
## Summary
- Add an admin account usage statistics page with filters, sortable account rows, pagination, and persisted auto-refresh settings.
- Add account detail statistics with date-range usage summaries and user activity tables.
- Add backend account stats and recent/range user endpoints, including account billing and user charge totals.

## Notes
- The recent active users table is shown before the date-range active users table.
- Recent active users show current requests based on real-time user concurrency, while the date-range active users table omits the current requests column.
- Auto-refresh updates the table and open detail dialog silently to avoid full-page loading flicker.

## Tests
- Not run after branch split; changes were previously validated with frontend typecheck and Docker build on the combined branch.


gebdalaoli-arch:
功能看起来不错

PR #2234, opened 2026-05-06
feat(easypay): add per-channel payment type override

billzhong:
## Summary
- Add `typeAlipay` / `typeWxpay` optional config keys to EasyPay, allowing
  per-channel payment type overrides (e.g. mapping `alipay` → `alipay_h5`)
- Extract `resolveByChannel` helper to consolidate the channel-specific
  config lookup pattern (previously only in `resolveCID`, now shared with `resolveType`)
- Frontend: allow clearing optional non-sensitive overrides by preserving
  empty values on save; only drop empty sensitive fields to let the backend
  keep stored secrets


billzhong:

PR #2273, opened 2026-05-08
fix: allow direct first login for LinuxDo OAuth

handsomjack0:
## Summary
- allow direct account creation and login for first-time LinuxDo OAuth users when there is no compatible email conflict
- keep the existing direct-login path for already bound LinuxDo identities
- keep invite-only signup working by using an invitation-only pending flow instead of falling back to legacy email completion
- add regression tests for both direct-login and invite-only flows

## Validation
- local test command:
  `go test ./internal/handler -run ''TestLinuxDoOAuth|TestCompleteLinuxDoOAuthRegistration'' -count=1`
- result: passed

## Notes
- the currently deployed hotfix image was built from the same logic and is already running on the DO host


handsomjack0:

PR #2356, opened 2026-05-11
Preserve multi-tool context in OpenAI messages continuation

jack-atlas:
Claude Code can send one assistant turn with multiple tool_use blocks followed by a user turn containing matching tool_result blocks. The OpenAI /v1/messages compatibility path trimmed continuation input to the last user turn plus adjacent tool outputs, which could leave a function_call_output without its earlier function_call when previous_response_id was attached.

This keeps all function_call items needed by retained function_call_output entries so the upstream Responses API can resolve every call_id.

Constraint: Applies only to the OpenAI /v1/messages -> Responses compatibility continuation path.

Rejected: Disable previous_response_id for all tool outputs | loses continuation and cache benefits for valid turns.

Confidence: high

Scope-risk: narrow

Directive: Do not trim function_call_output entries without preserving their matching function_call call_id context.

Tested: go test ./internal/service -run 'TestForwardAsAnthropic_(PreviousResponseIDKeepsMultiToolCallContext|AttachesPreviousResponseIDForCompatContinuation|OAuthPreservesClaudeCodeToolCallID)' -count=1

Tested: go test ./internal/service -run 'TestForwardAsAnthropic|TestApplyAnthropicCompatFullReplayGuard|TestOpenAICompat|Test.*ToolContinuation' -count=1

Tested: go test ./internal/pkg/apicompat -count=1

Related: #2337

### DIFF
--- a/backend/internal/service/openai_compat_model_test.go
+++ b/backend/internal/service/openai_compat_model_test.go
@@ -480,6 +480,61 @@ func TestForwardAsAnthropic_AttachesPreviousResponseIDForCompatContinuation(t *t
 	require.Equal(t, "second", gjson.GetBytes(upstream.lastBody, "input.1.content.0.text").String())
 }
 
+func TestForwardAsAnthropic_PreviousResponseIDKeepsMultiToolCallContext(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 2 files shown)


jack-atlas:

PR #2395, opened 2026-05-12
fix(auth): prefer OIDC compat email in pending flow

imlewc:
## Summary
- include `compat_email` in the OIDC pending completion type
- prefer provider-compatible email before synthetic OIDC emails when pre-filling the pending account flow

## Why
OIDC pending account flow can receive a real provider email in `compat_email`, while the internal `email`/`resolved_email` may be a synthetic `@oidc-connect.invalid` address. Preferring `compat_email` avoids showing the synthetic address to users when a real email is available.

## Testing
- `pnpm --dir frontend run build`


imlewc:
recheck

imlewc:

PR #2463, opened 2026-05-14
Fix OpenAI image edit mask handling

nanzemxx:
## Summary
- Normalize uploaded image edit masks before forwarding them through the OAuth Responses image bridge.
- Limit returned Responses image results to the caller-requested n for both streaming and non-streaming image routes.
- Add tests for mask normalization and requested image count limiting.

## Details
This keeps mask semantics based on alpha while clearing RGB values in transparent pixels, which avoids downstream/upstream layers accidentally interpreting transparent mask pixels as visible black content. It also ensures that if an upstream Responses image event returns more image outputs than requested, the gateway only returns the requested count and bills/logs the matching image count.

## Tests
- go test ./internal/service -run 'OpenAIImages|Images|NormalizeOpenAIImageMaskUpload'

### DIFF
--- a/backend/internal/service/openai_images_responses.go
+++ b/backend/internal/service/openai_images_responses.go
@@ -6,6 +6,9 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"image"
+	"image/color"
+	"image/png"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


nanzemxx:

nanzemxx:
This should help with the masked image edit path reported in #2251 and is related to the OAuth image edits failures discussed in #2232.

sapipoZZZ:
Is there any progress now?

Issue #2152, opened 2026-05-01
gpt-5.5压缩失败


qingqing1207:
sub2api的容器日志
2026-05-01T21:20:29.749+0800    WARN    handler/openai_gateway_handler.go:377   openai.forward_failed   {"service": "sub2api", "env": "production", "component": "http", "request_id": "5a33bf9f-569c-4a09-8c9a-044555a6eecf", "client_request_id": "", "path": "/responses/compact", "method": "POST", "client_request_id": "8722016d-01f2-4e41-bd4f-d8c319d5e01f", "component": "handler.openai_gateway.responses", "user_id": 1, "api_key_id": 12, "group_id": 13, "model": "gpt-5.5", "stream": false, "account_id": 26, "fallback_error_response_written": false, "error": "upstream request failed: Post \"https://chatgpt.com/backend-api/codex/responses/compact\": context canceled"}
2026-05-01T21:20:29.750+0800    WARN    handler/openai_gateway_handler.go:508   codex.remote_compact.failed     {"service": "sub2api", "env": "production", "component": "http", "request_id": "5a33bf9f-569c-4a09-8c9a-044555a6eecf", "client_request_id": "", "path": "/responses/compact", "method": "POST", "client_request_id": "8722016d-01f2-4e41-bd4f-d8c319d5e01f", "component": "handler.openai_gateway.responses", "remote_compact": true, "compact_outcome": "failed", "status_code": 502, "latency_ms": 56505, "path": "/responses/compact", "force_codex_cli": false, "request_user_agent": "codex_vscode/0.126.0-alpha.8 (Mac OS 15.7.3; arm64) unknown (VS Code; 26.422.71525)", "request_model": "gpt-5.5", "account_id": 26, "upstream_request_id": "5a33bf9f-569c-4a09-8c9a-044555a6eecf"}
[...]

Issue #1609, opened 2026-04-13
Error running remote compact task: unexpected status 502 Bad Gateway: Upstream request failed, url: http://***/v1/responses/compact


chroime:
Error running remote compact task: unexpected status 502 Bad Gateway: Upstream request failed, url: http://***/v1/responses/compact

每次到背景信息压缩这就报错

ShangTX:
我最近遇到这个问题发现在codex会话切换一下模型，然后随便发点什么，再把模型切换回去继续就可以了

PR #2494, opened 2026-05-15
fix: preserve Codex Responses headers

cyhhao:
## What

- Preserve Codex-specific Responses headers, including `X-Codex-Beta-Features`, `Version`, turn metadata/state, window id, parent thread id, and installation id.
- Infer `x-codex-beta-features: remote_compaction_v2` when a Codex `context_compaction` item is present but an intermediate client/proxy omitted the header.
- Pass `X-Codex-Turn-State` response headers back to Codex clients so later requests in the same turn can replay the sticky-routing token.
- Add streaming coverage to ensure `context_compaction` SSE output items pass through unchanged.
- Add coverage for both normal OpenAI forwarding and OpenAI passthrough forwarding.

## Why

Recent Codex remote compaction v2 requests send `context_compaction` items through `/v1/responses` and advertise the protocol with `x-codex-beta-features: remote_compaction_v2`. Official Codex also sends a `version` header on Responses requests. If sub2api filters these headers, ChatGPT Codex backend can validate the request as a normal/older Responses payload and reject `context_compaction` as an invalid item type.

The v2 flow is streaming. Codex also captures `x-codex-turn-state` from the stream response headers and replays it on subsequent requests in the same turn. Preserving that response header keeps the v2 streaming contract intact.

## Related issues

- Directly related: #2490
- Similar remote compact / compact compatibility reports that may share the same header/schema mismatch symptom: #1609, #2016, #2152, #1887, #1865, #1900, #1927, #1933

Some of the older reports also mention model mapping, fixed Codex version headers, or timeout/context-canceled behavior, so this PR intentionally references them as related rather than claiming to close all of them.

## Test

- `go test ./internal/util/responseheaders`

## Risk

Low. This extends existing Codex-specific allowlists and preserves the upstream SSE body shape; it does not rewrite request or response bodies.

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -38,9 +38,10 @@ const (
 	// ChatGPT internal API for OAuth accounts
 	chatgptCodexURL = "https://chatgpt.com/backend-api/codex/responses"
 	// OpenAI Platform API for API Key accounts (fallback)
-	openaiPlatformAPIURL   = "https://api.openai.com/v1/responses"
-	openaiStickySessionTTL = time.Hour // 粘性会话TTL
-	codexCLIUserAgent      = "codex_cli_rs/0.125.0"
+	openaiPlatformAPIURL           = "https://api.openai.com/v1/responses"
+	openaiStickySessionTTL         = time.Hour // 粘性会话TTL
+	codexCLIUserAgent              = "codex_cli_rs/0.125.0"
+	codexRemoteCompactionV2Feature = "remote_compaction_v2"
 	// codex_cli_only 拒绝时单个请求头日志长度上限（字符）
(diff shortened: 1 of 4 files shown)


cyhhao:

cyhhao:
recheck

47seek:
Hi, I’m also hitting this GPT-5.5 remote compact failure. I noticed #2494 seems related to preserving Codex Responses headers / remote_compaction_v2, but it was closed without being merged. Was that PR superseded by another fix, or was there a blocker with that approach?

cyhhao:
@047lovelife 
This fix didn't resolve my issue. I eventually chose to disable this feature.
`~/.codex/config.toml`

```
[features]
remote_compaction_v2 = false
```

However, you can try following this PR of thought. It is a repair based on Codex source code speculation, which makes some sense.

PR #2543, opened 2026-05-18
  fix: preserve DeepSeek reasoning_content in chat compatibility paths

L494264Tt:
This preserves DeepSeek reasoning_content across chat compatibility paths.

  DeepSeek thinking-mode tool-call conversations may require the assistant reasoning_content from previous turns to be
  sent back in later requests. Without preserving it, those conversations can fail or lose reasoning context.

  Changes:
  - Preserve assistant reasoning_content when converting Chat Completions messages to Responses input by wrapping it as
  a thinking block.
  - Add regression coverage for non-streaming DeepSeek responses.
  - Add regression coverage for streaming DeepSeek deltas.
  - Add regression coverage that request-side messages[].reasoning_content is passed through with tool calls.

  Tests:
  go test -tags=unit ./internal/pkg/apicompat ./internal/service -run
  'TestChatCompletionsToResponses_AssistantReasoningContentPreserved|
  TestChatCompletionsToResponses_AssistantThinkingTagPreserved|
  TestForwardAsRawChatCompletions_PreservesDeepSeekReasoningContent|
  TestForwardAsRawChatCompletions_ForcesStreamUsageUpstreamAndPassesUsageDownstream


L494264Tt:

PR #2603, opened 2026-05-20
Sanitize Claude thinking history before OpenAI Responses forwarding

warkcod:
## Summary

This fixes OpenAI Responses requests that fail when Claude-style clients replay historical thinking fields in the input history.

Some clients can send Responses input items containing Anthropic-style fields such as:

- `input[N].thinking`
- content blocks with `type: "thinking"`
- content blocks/items with `type: "redacted_thinking"`

The OpenAI Responses upstream rejects these with `unknown_parameter`, for example:

```text
Unknown parameter: 'input[130].thinking'
```

That upstream 400 is currently surfaced by the gateway as a failed Responses request, and in failover paths can become a 502 to the client.

## Changes

- Normalize OpenAI Responses input history before upstream forwarding.
- Apply the normalization in both OAuth forwarding paths:
  - transformed OAuth Responses path
  - OAuth passthrough path
- Remove unsupported `thinking` / `redacted_thinking` fields from input items and content blocks.
- Convert explicit `type: "thinking"` content blocks to normal text blocks so historical context is preserved where possible.
- Preserve valid OpenAI Responses `type: "reasoning"` items instead of dropping them wholesale.
- Add regression coverage for both OAuth legacy transform and OAuth passthrough behavior.

## Why not route to Chat Completions?

Routing long-context traffic to Chat Completions avoids this specific schema error, but it changes endpoint semantics, reasoning parameters, usage accounting, tool/event behavior, and prompt-cache behavior. This PR keeps traffic on Responses and normalizes only the unsupported input-history fields.

## Test Plan

```bash
cd backend
go test ./internal/service -run 'OpenAI|Passthrough|Responses'
go test ./...
```

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -2369,6 +2369,10 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 		bodyModified = true
 		disablePatch()
 	}
+	if sanitizeOpenAIResponsesInputThinkingFields(reqBody) {
+		bodyModified = true
+		disablePatch()
+	}
 
 	// Apply OpenAI fast policy (参照 Claude BetaPolicy 的 fast-mode 过滤)：
 	// 针对 body 的 service_tier 字段（"priority" 即 fast，"flex"），按策略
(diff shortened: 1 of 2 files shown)


warkcod:

PR #2621, opened 2026-05-20
feat(accounts): support group_name in account data import

hongdongjian:
## Summary
- support per-account `group_name` in admin account data import JSON
- bind imported accounts only when a same-name group exists on the same platform
- skip group binding without failing when the group is missing or belongs to another platform

## Tests
- `go test ./internal/handler/admin`
- `pnpm run typecheck`


hongdongjian:
recheck

PR #2622, opened 2026-05-20
feat(accounts): 支持导入账号按分组名绑定

hongdongjian:
## Summary
- support per-account `group_name` in admin account data import JSON
- bind imported accounts only when a same-name group exists on the same platform
- skip group binding without failing when the group is missing or belongs to another platform

## Tests
- `go test ./internal/handler/admin`
- `pnpm run typecheck`


hongdongjian:

hongdongjian:
recheck

PR #2669, opened 2026-05-21
fix(antigravity): tool_choice forwarding, system_instruction oneof, and non-stream functionCall collection

xykong:
## Summary

Three bugs fixed in the Antigravity gateway:

### 1. `tool_choice` not forwarded to Gemini `FunctionCallingConfig`

**Problem:** OpenAI `tool_choice` parameter (e.g. `{"type":"function","function":{"name":"get_weather"}}`) was completely ignored when constructing Gemini upstream requests. `ClaudeRequest` had no `ToolChoice` field, and `GeminiFunctionCallingConfig` was hardcoded to `Mode: "VALIDATED"` with no `AllowedFunctionNames` support.

**Impact:** Non-streaming requests with forced `tool_choice` returned `tool_calls=null` — the model spent all tokens on reasoning instead of calling the specified tool.

**Fix:**
- Added `ToolChoice json.RawMessage` to `ClaudeRequest`
- Added `AllowedFunctionNames []string` to `GeminiFunctionCallingConfig`
- `TransformClaudeToGeminiWithOptions` now maps `tool_choice` to Gemini modes:
  - `auto` → `AUTO`
  - `any` → `ANY`
  - `tool`/`function` with name → `ANY` + `AllowedFunctionNames`
  - `none` → `NONE`
  - absent → keeps existing `VALIDATED` (backward compatible)

### 2. `system_instruction` oneof conflict (HTTP 400) when tools present

**Problem:** Requests with both a `system` message and `tools` produced HTTP 400 from Antigravity upstream:
```
Invalid value at 'request' (oneof), oneof field '_system_instruction' is already set.
Cannot set 'system_instruction'
```

**Root cause:** The Antigravity upstream protobuf schema treats `system_instruction` as a oneof field that conflicts with `tools`/`toolConfig` in the same request.

**Fix:**
- When tools are present, system instruction content is prepended to `contents` instead of using the `systemInstruction` field
- Applied in both the Claude-to-Gemini transform path (`request_transformer.go`) and the native Gemini forwarding path (`injectIdentityPatchToGeminiRequest` in `antigravity_gateway_service.go`)
- Behavior unchanged when no tools are present

### 3. Non-streaming responses dropping `functionCall` parts

**Problem:** `handleGeminiStreamToNonStreaming` only collected `text` and `inlineData` parts from streaming chunks, silently dropping `functionCall`, `thought`, and all other part types. Since `ForwardGemini` always uses `streamGenerateContent` upstream (even for non-streaming client requests), this meant non-streaming tool calls never worked.

**Fix:** Replaced the type-specific `collectedImageParts`/`collectedTextParts` collection with a unified `collectedParts []map[string]any` that captures all part types, then calls the existing `mergeCollectedPartsToResponse` helper — matching the pattern already used by `handleClaudeStreamToNonStreaming`.

## Testing

- 7 new regression tests added in `request_transformer_test.go` covering tool_choice mapping and system+tools handling
- All existing tests pass: `go test ./internal/pkg/antigravity/...` and `go test -tags unit ./internal/service/...`
- Verified in production with live requests:
  - Non-streaming forced `tool_choice` now returns `finish_reason=tool_calls` with correct function call
  - System + tools requests return HTTP 200 (previously 400)
  - Full tool round-trip (Turn 1: tool call → Turn 2: tool result → natural language response) works end-to-end
  - Streaming tool calls and simple requests unchanged (regression verified)

## Changed Files

| File | Changes |
|---|---|
| `backend/internal/pkg/antigravity/claude_types.go` | +1: `ToolChoice` field |
| `backend/internal/pkg/antigravity/gemini_types.go` | +2: `AllowedFunctionNames` field |
| `backend/internal/pkg/antigravity/request_transformer.go` | +64: tool_choice mapping + system→contents migration |
| `backend/internal/pkg/antigravity/request_transformer_test.go` | +123: regression tests |
| `backend/internal/service/antigravity_gateway_service.go` | +45/-37: system_instruction oneof fix + unified parts collection |

### DIFF
--- a/backend/internal/pkg/antigravity/claude_types.go
+++ b/backend/internal/pkg/antigravity/claude_types.go
@@ -15,6 +15,7 @@ type ClaudeRequest struct {
 	TopP        *float64        `json:"top_p,omitempty"`
 	TopK        *int            `json:"top_k,omitempty"`
 	Tools       []ClaudeTool    `json:"tools,omitempty"`
+	ToolChoice  json.RawMessage `json:"tool_choice,omitempty"`
 	Thinking    *ThinkingConfig `json:"thinking,omitempty"`
 	Metadata    *ClaudeMetadata `json:"metadata,omitempty"`
 }
--- a/backend/internal/pkg/antigravity/gemini_types.go
+++ b/backend/internal/pkg/antigravity/gemini_types.go
@@ -117,7 +117,8 @@ type GeminiToolConfig struct {
 
 // GeminiFunctionCallingConfig 函数调用配置
 type GeminiFunctionCallingConfig struct {
-	Mode string `json:"mode,omitempty"` // VALIDATED, AUTO, NONE
+	Mode                 string   `json:"mode,omitempty"` // VALIDATED, AUTO, ANY, NONE
+	AllowedFunctionNames []string `json:"allowedFunctionNames,omitempty"`
 }
 
 // GeminiSafetySetting Gemini 安全设置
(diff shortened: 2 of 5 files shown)


xykong:

PR #2703, opened 2026-05-23
docs: add OpenAI image job sidecar example

rwang23:
## Summary
- add an example Node.js sidecar for async OpenAI-compatible image jobs
- document the 202 + job_id polling flow for long-running /v1/images/generations and /v1/images/edits requests
- link the sidecar example from the main README

## Why
Deployments behind a CDN or reverse proxy can have a shorter origin response timeout than high-resolution image generation. The sidecar keeps public requests short while forwarding the long-running image request to local Sub2API in the background.

## Test Plan
- npm run check in examples/openai-image-job-sidecar
- local smoke test: /healthz succeeds, POST /image-jobs returns queued, polling returns failed when a fake upstream is unreachable

## Notes
This is intentionally an example sidecar. It does not change Sub2API gateway internals or existing synchronous image endpoints.


rwang23:

PR #2760, opened 2026-05-25
fix: encode unicode email headers

ANDTea2023:
## Summary

- encode non-ASCII email subjects with RFC 2047 before sending SMTP DATA
- format sender display names with `net/mail` so Unicode names are encoded correctly
- keep existing CR/LF header injection sanitization in place

## Why

Some strict recipient servers reject messages containing raw Unicode in headers, for example:

`550 Message denied: non-RFC2047 unicode in headers`

This can happen when the site name or SMTP sender display name contains non-ASCII text.

## Tests

- `go test ./internal/service -run 'Test(EncodeEmailTextHeader|FormatEmailAddressHeader)'`

### DIFF
--- a/backend/internal/service/email_service.go
+++ b/backend/internal/service/email_service.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 	"log/slog"
 	"math/big"
+	"mime"
 	"net"
+	"net/mail"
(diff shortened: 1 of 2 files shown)


ANDTea2023:

PR #2770, opened 2026-05-26
fix: search usage API keys remotely

hansnow:
## Background

The user usage page only loaded a small first page of API keys for the API key filter, then filtered that local list inside the select component. When a user had more keys than the initial page contained, keys outside that page could not be found from the usage filter even though they existed in the API keys page.

## Solution

- Add remote-search support to the shared Select component, including a loading state and search event.
- Change the user usage API key filter to debounce search input and query `/keys` with the search term.
- Keep the currently selected key option available after the dropdown refreshes back to the default page.
- Guard against stale in-flight search responses overwriting newer results.

## Tests

- `pnpm test:run src/views/user/__tests__/UsageView.spec.ts`
- `pnpm typecheck`

### DIFF
--- a/frontend/src/components/common/Select.vue
+++ b/frontend/src/components/common/Select.vue
@@ -61,7 +61,7 @@
           <!-- Options list -->
           <div class="select-options" ref="optionsListRef">
             <div
-              v-for="(option, index) in filteredOptions"
+              v-for="(option, index) in visibleOptions"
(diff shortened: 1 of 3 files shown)


hansnow:
This supersedes the earlier #1692 PR for the same /usage API key filter issue. The old PR had drifted into a merge conflict, while this PR is rebased on current main and keeps the fix focused around explicit remote-search behavior in the shared Select component, plus loading state handling, selected-option preservation, and stale-response guarding.

PR #2818, opened 2026-05-27
feat: add aligned 5-hour user platform quota

xcrhzbk:
## Summary
- Add per-user per-platform 5-hour quota limits alongside existing daily / weekly / monthly quotas.
- Use administrator-configured aligned 5-hour windows instead of rolling windows.
- Expose 5-hour quota configuration and reset controls in admin UI.
- Show 5-hour quota usage in user quota views.

## Behavior
- `five_hour_limit_usd = null` means unlimited for the 5-hour window.
- `five_hour_align_minutes` is minutes after local midnight, e.g. `60` means windows start at 01:00, 06:00, 11:00, etc.
- 5-hour reset time is calculated from the aligned window, including after manual reset.
- Old clients that omit `five_hour_align_minutes` preserve the existing alignment value.

## Database / Migration
- Adds `five_hour_limit_usd`, `five_hour_usage_usd`, `five_hour_window_start`, and `five_hour_align_minutes` to `user_platform_quotas`.
- Adds constraints for valid alignment minutes.
- Ent generated files are included.
- User platform quota Redis cache schema is bumped so old cache entries miss and refresh from DB.

## Safety
- Old Redis quota cache entries cannot bypass the new 5-hour limit because schema version mismatch forces DB refresh.
- Existing administrator alignment settings are preserved when clients omit `five_hour_align_minutes`.
- Registration default quota insertion does not overwrite existing user-specific 5-hour alignment unnecessarily.
- Admin UI supports 5-hour limit, alignment, reset confirmation, and display.

## Tests
- `go test ./...`
- `go test -tags=unit ./...`
- `go generate ./ent`
- `golangci-lint run ./...`
- `govulncheck ./...`
- `pnpm --dir frontend install --frozen-lockfile --ignore-scripts`
- `make test-frontend`
- `git diff --check`

## Notes
- Local `go test -tags=integration ./...` could not complete because Docker is unavailable locally; repository integration tests should run in GitHub CI.


xcrhzbk:

xcrhzbk:
recheck

PR #2827, opened 2026-05-27
fix: fallback to SSE body for API key responses

### LINKED ISSUE #2826: API key non-streaming Responses can miss SSE fallback when upstream omits event-stream content type
## Problem

`POST /v1/responses` with `stream=false` can still receive an SSE-formatted body from an OpenAI-compatible upstream. The current non-streaming OpenAI gateway handles this when the upstream sets `Content-Type: text/event-stream`, and it also has a body-content fallback for OAuth accounts.

For API key accounts, if an upstream sends SSE data but omits or mislabels the content type, the response falls through to JSON parsing and returns `parse response: invalid json response` instead of converting the terminal `response.completed` event back to a JSON Responses payload.

This is most visible in chained compatible deployments such as:

```text
client
  -> sub2api /v1/responses stream=false
  -> OpenAI-compatible upstream that returns SSE text
     without text/event-stream
```

## Expected behavior

When the body is not valid JSON and looks like OpenAI SSE, the non-streaming path should reuse the existing SSE-to-JSON conversion path. That path already reconstructs `response.output` from accumulated `response.output_text.delta` events when the terminal response has `output: []`.

## Scope

Minimal backend fix only:

- keep the existing `Content-Type: text/event-stream` detection
- keep the existing OAuth early body fallback
- add a late fallback for any account type only after normal JSON usage parsing fails
- add a regression test for an API key account with a mislabeled SSE body

Related: #1493

ttt132:
## Summary

- add a late SSE-body fallback for non-streaming OpenAI Responses API calls after normal JSON parsing fails
- keep the existing early OAuth-only body heuristic to avoid changing valid JSON handling
- add regression coverage for API key accounts when an upstream returns SSE text with a wrong `Content-Type`

Fixes #2826
Related #1493

## Tests

```bash
GOPROXY=https://goproxy.cn,direct go test ./internal/service -run 'TestHandleNonStreamingResponse_APIKeyFallsBackToSSEBodyWhenContentTypeIsWrong|TestHandleSSEToJSON_CompletedEventReturnsJSON|TestHandleSSEToJSON_ReconstructsImageGenerationOutputItemDone' -count=1
GOPROXY=https://goproxy.cn,direct go test ./internal/service -count=1
```

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -4903,20 +4903,22 @@ func (s *OpenAIGatewayService) handleNonStreamingResponse(ctx context.Context, r
 	if isEventStreamResponse(resp.Header) {
 		return s.handleSSEToJSON(resp, c, body, originalModel, mappedModel)
 	}
+	bodyLooksLikeSSE := bytes.Contains(body, []byte("data:")) || bytes.Contains(body, []byte("event:"))
+
 	// For OAuth accounts, also fall back to a body-content heuristic because
 	// the upstream may omit the Content-Type header while still sending SSE.
 	// This heuristic is NOT applied to API-key accounts to avoid false
 	// positives on JSON responses that coincidentally contain "data:" or
 	// "event:" in their text content.
-	if account.Type == AccountTypeOAuth {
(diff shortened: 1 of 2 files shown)


ttt132:

PR #2845, opened 2026-05-28
feat(admin): add user API key creation endpoint

### LINKED ISSUE #2843: feat: add admin API for provisioning user API keys
## Problem

API keys can currently be created through the authenticated user API, which requires a user JWT session. This works well for the built-in web UI, but it is awkward for admin-side integrations or provisioning systems that already manage users externally and need to create a sub2api API key for a known user.

Without an admin provisioning endpoint, external systems have to choose between less ideal options:

- Ask the user to log into sub2api and create the key manually.
- Handle or impersonate the user's JWT session.
- Store or reset the user's sub2api password just to call the user-facing key creation endpoint.

For integrations, an admin-authenticated API is a cleaner boundary than handling user login state.

## Proposal

Add an admin-only endpoint that creates an API key for a target user ID.

Suggested contract:

- `POST /api/v1/admin/users/:id/api-keys`
- Request body: `{ "name": "..." }`
- Response: created API key using the existing response envelope and DTO shape

The endpoint should reuse the existing API key creation service so generated keys, validation, persistence, status, and cache invalidation remain consistent with normal user-created keys.

## Why this fits the existing admin API

This complements existing admin-side key management capabilities:

- Admins can already list a user's API keys through `GET /api/v1/admin/users/:id/api-keys`.
- Admins can already update API key admin-managed fields through `PUT /api/v1/admin/api-keys/:id`.
- This proposal fills the provisioning gap without changing the existing user JWT `/keys` flow.

## Security Notes

This is security-sensitive because API key creation returns the full key at creation time. The endpoint should therefore:

- Be protected by the existing admin route authentication.
- Reuse the existing API key creation service instead of introducing a separate persistence path.
- Preserve the existing one-time secret-return behavior used by the current create flow.
- Avoid requiring admin integrations to impersonate users or handle user JWT sessions.

## Scope

In scope:

- Admin-authenticated creation of an API key for a specified user ID.
- Reuse of the existing API key creation service.
- Handler coverage for success and error paths.

Out of scope:

- Environment-variable or docker-compose based API key provisioning.
- Bulk API key creation.
- Custom key import or migration tooling.
- Changes to the existing user-facing `/keys` API.

## Acceptance Criteria

- Admin clients can create an API key for a target user ID through an admin-authenticated API call.
- Invalid target user IDs return a client error.
- Service errors return the existing standardized error response.
- Existing user-owned API key creation behavior remains unchanged.
- Tests cover the new admin creation endpoint.

## Related

Related but different from #2284, which focuses on startup-time API key provisioning through environment variables or docker-compose.

JeanphiloGong:
## Summary
- Add `POST /api/v1/admin/users/:id/api-keys` so admins can create API keys for a specific user.
- Reuse the existing `APIKeyService.Create` path instead of duplicating API key creation logic.
- Add admin handler tests for success, invalid user ID, missing user, and API key service failure cases.

## Security Notes
- The endpoint is registered under the existing admin route group and uses the existing admin auth boundary.
- The full API key is returned only in the creation response, matching the existing API key creation behavior.

## Tests
- `cd backend && go test ./internal/handler/admin`
- `cd backend && make test-unit`

## CLA

Closes #2843


JeanphiloGong:

PR #2846, opened 2026-05-28
feat(admin/proxies): show exit IP address in location column

6rok:
## Summary

- Display the proxy exit IP address in the admin proxies table's location column
- IP shown as a compact `<code>` block below the country/city text
- Only renders when `ip_address` is present in the latency cache data

## Changes

- `frontend/src/views/admin/ProxiesView.vue`: Updated location cell template to show `row.ip_address` alongside existing geo info

## Screenshot

The location column now shows:
```
🇺🇸 United States
203.0.113.42
```

Instead of just:
```
🇺🇸 United States
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,5 +134,6 @@ docs/*
 .serena/
 .codex/
 frontend/coverage/
+.pnpm-store/
 aicodex
 output/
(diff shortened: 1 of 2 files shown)


6rok:

PR #2865, opened 2026-05-28
fix(gateway): preserve usage request context

wey-gu:
## Background

Gateway usage records are written asynchronously. Some handlers submitted those tasks with a fresh worker context, so request-scoped identifiers were lost before `RecordUsage` built the usage-log request id.

That makes otherwise successful requests harder to reconcile downstream: the response can finish with one request id, while the usage log falls back to a generated id.

## Change

- Expose the generated `X-Client-Request-ID` response header from the gateway request middleware.
- Preserve request-scoped context when OpenAI-compatible, Claude-compatible, and Gemini gateway handlers submit async usage-record tasks.
- Keep the billing id server-generated; this does not trust a caller-provided id for usage dedupe.

## Verification

- `go test ./internal/server/middleware ./internal/handler ./internal/service -run 'TestClientRequestID|TestSubmitUsageRecordTaskCopiesRequestContext|TestOpenAISubmitUsageRecordTaskCopiesRequestContext|TestOpenAIGatewayServiceRecordUsage_PrefersClientRequestIDOverUpstreamRequestID|TestGatewayServiceRecordUsage_PrefersClientRequestIDOverUpstreamRequestID|TestOpenAI'`

### DIFF
--- a/backend/internal/handler/gateway_handler.go
+++ b/backend/internal/handler/gateway_handler.go
@@ -510,7 +510,7 @@ func (h *GatewayHandler) Messages(c *gin.Context) {
 
 			// 使用量记录通过有界 worker 池提交，避免请求热路径创建无界 goroutine。
 			quotaPlatform := service.QuotaPlatform(c.Request.Context(), apiKey)
-			h.submitUsageRecordTask(func(ctx context.Context) {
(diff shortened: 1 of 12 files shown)


wey-gu:
麻烦 @Wei-Shaw 有时间帮忙辛苦 review 下哈 ♥️，不急 🫡

PR #2916, opened 2026-05-30
fix(apicompat): repair codex Responses<->Anthropic/ChatCompletions conversion

hanyunsushi:
## Problem

When a Responses API client (e.g. codex) is routed through an Anthropic upstream or a chat/completions-only upstream, requests fail or render blank due to several bugs in the `apicompat` conversion layer. Account "test connection" passes because it uses a self-constructed simple request that never exercises the Responses input parsing, masking these issues.

Related: #2913

### Request direction (`ResponsesInputItem`)

1. **`Arguments`/`Output` typed as `string`** — codex sends `function_call.arguments` as a JSON object and `function_call_output.output` as an array. The strict `string` type caused `json.Unmarshal` to fail (502 on the Anthropic path) or `rawString` to silently drop the content (chat/completions path). Both are now `json.RawMessage` with normalization helpers (`normalizeResponsesArguments`, `extractResponsesOutputText`, `responsesArgumentsToChatString`); reverse-direction builders wrap strings with `jsonRawString` to preserve the OpenAI wire format.

2. **Tools without `parameters` produced a null `input_schema`** — codex `namespace` tools (MCP/agent) and bare `web_search` carry no parameters; Anthropic 422s on a null schema. Now backfilled with an empty object schema.

3. **`web_search` emitted as the Anthropic server tool `web_search_20250305`** — some Anthropic-compatible upstreams don't implement server tools (422). Now emitted as a plain function tool.

4. **Top-level `instructions` and `developer`-role items were dropped** — codex puts its primary system prompt in `instructions`; the old code only read `system`-role items, leaving the model with no instructions and (worse) leaking developer `input_text` blocks into a user message. Both now map to the Anthropic system field.

5. **Empty/whitespace system + string-form system** — an empty system 422s; and a string-form system combined with tools 422s on some upstreams. System is now omitted when empty and emitted in array form (`[{"type":"text","text":...}]`), which is spec-valid and what the official Claude Code client uses.

6. **`reasoning_effort: "xhigh"`** — codex sends `xhigh` for top-tier models; chat/completions upstreams only accept low/medium/high and 400 on xhigh. Now normalized to `high`.

### Response direction (streaming → Responses events)

7. **Missing `response.content_part.added`/`.done`** — strict clients (codex) need a content part before `output_text.delta`; without it text deltas have nothing to attach to and the client renders nothing. Now emitted around the text block on both conversion paths.

8. **`output_item.done` lacked content/tool-call fields** — codex collects the final output and tool calls from `OutputItemDone` items, not from delta events. Message items now carry their full `content`, and function_call items carry `call_id`/`name`/`arguments`.

9. **Ghost empty deltas** — some upstreams send a leading `{"content":""}` chunk; emitting it produced an empty `output_text.delta` and a premature message item, leaving codex stuck on "thinking" with no output. Empty content/reasoning deltas are now skipped.

## Tests

Adds coverage for all of the above: polymorphic arguments/output (string/object/array), tool schema backfill, web_search as function tool, instructions/developer → system, empty-system omission, xhigh normalization, content_part event ordering, output_item content/tool-call fields, and empty-delta skipping. All existing `apicompat` tests pass; `gofmt`/`go vet` clean.

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_responses_test.go
+++ b/backend/internal/pkg/apicompat/anthropic_responses_test.go
@@ -143,7 +143,7 @@ func TestAnthropicToResponses_ToolUse(t *testing.T) {
 	assert.Empty(t, items[2].ID)
 	assert.Equal(t, "function_call_output", items[3].Type)
 	assert.Equal(t, "call_1", items[3].CallID)
-	assert.Equal(t, "Sunny, 72°F", items[3].Output)
+	assert.Equal(t, `"Sunny, 72°F"`, string(items[3].Output))
 }
 
 func TestAnthropicToResponses_ThinkingIgnored(t *testing.T) {
@@ -1340,7 +1340,7 @@ func TestAnthropicToResponses_ToolResultWithImage(t *testing.T) {
 	// function_call_output should have text-only output (no image).
 	assert.Equal(t, "function_call_output", items[2].Type)
 	assert.Equal(t, "toolu_1", items[2].CallID)
-	assert.Equal(t, "(empty)", items[2].Output)
+	assert.Equal(t, `"(empty)"`, string(items[2].Output))
 
 	// Image should be in a separate user message.
 	assert.Equal(t, "user", items[3].Role)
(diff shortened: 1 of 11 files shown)


hanyunsushi:

PR #2922, opened 2026-05-31
fix: avoid OpenAI WS usage dedup conflicts

gsh20040816:
## Summary

- use each OpenAI WebSocket turn's upstream `resp_*` id as the usage billing id when `OpenAIWSMode` is true
- keep the existing `client_request_id` preference for normal OpenAI usage records
- add a regression test for WS usage records with a connection-level client request id

## Root cause

`v0.1.133` started preserving request context for async usage recording. That is correct for normal HTTP requests, but OpenAI Responses WebSocket sessions can contain multiple billable turns under one connection-level `client_request_id`.

When every turn reuses `client:<client_request_id>` as the usage billing key, the first turn records successfully and later turns hit `usage billing request fingerprint conflict` because their token/cost fingerprint differs. Using the per-turn upstream response id keeps WebSocket usage idempotent per billable turn.

## Testing

- `go test ./internal/service -run 'TestOpenAIGatewayServiceRecordUsage_(WSModePrefersUpstreamRequestIDOverClientRequestID|PrefersClientRequestIDOverUpstreamRequestID|UsesFallbackRequestIDForBillingAndUsageLog|BillingFingerprintIncludesRequestPayloadHash)' -count=1`
- `go test ./internal/handler -run 'TestOpenAIGatewayHandlerSubmit|TestOpenAIResponsesWebSocket_PassthroughUsageLog|TestOpenAIResponsesWebSocket_FailoverOnUpstreamUsageLimitEvent' -count=1`

### DIFF
--- a/backend/internal/service/openai_gateway_record_usage_test.go
+++ b/backend/internal/service/openai_gateway_record_usage_test.go
@@ -721,6 +721,37 @@ func TestOpenAIGatewayServiceRecordUsage_PrefersClientRequestIDOverUpstreamReque
 	require.Equal(t, "client:openai-client-stable-123", usageRepo.lastLog.RequestID)
 }
 
+func TestOpenAIGatewayServiceRecordUsage_WSModePrefersUpstreamRequestIDOverClientRequestID(t *testing.T) {
+	usageRepo := &openAIRecordUsageLogRepoStub{}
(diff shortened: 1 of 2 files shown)


gsh20040816:

PR #2936, opened 2026-06-01
fix(openai): bridge oversized websocket requests

xlx0852:
## Summary
- make the OpenAI Responses WS client read limit configurable and bridge oversized first frames through HTTP/SSE while keeping the downstream WebSocket open
- replay collected tool-call context for bridged and recovered turns so `function_call_output` can continue without `previous_response_id`
- allow store-disabled WS sessions to recover from stale upstream connections when replay context is available

## Tests
- go test ./...
- mvn compile skipped: no pom.xml in this Go project


xlx0852:
recheck

xlx0852:

PR #2951, opened 2026-06-01
Add codex-auto-review to OpenAI default models

EricLi404:
## Summary
- add `codex-auto-review` to the OpenAI default `/v1/models` fallback catalog
- add a focused regression test so the default model list keeps exposing it

## Validation
- `go test ./internal/pkg/openai ./internal/handler -run 'TestDefaultModelsContainsCodexAutoReview|TestGatewayModels'`\n- `make test-unit`\n- `git diff --check`\n\n## Notes\n- `golangci-lint run ./...` was started locally with golangci-lint 2.11.4, but it ran for several minutes without output and was stopped; CI uses v2.9 and should provide the authoritative lint result.


EricLi404:

PR #2953, opened 2026-06-01
feat(admin): add API key provisioning

qinghui316:


qinghui316:
Closing this PR because this branch is for private deployment only and is not intended for upstream contribution.

PR #2997, opened 2026-06-03
Bind OpenAI HTTP response IDs to selected accounts

### LINKED ISSUE #2991: OpenAI Responses encrypted context can fail when routed across upstream accounts
## Problem

When Sub2API is used as an OpenAI-compatible relay with multiple upstream OpenAI accounts in the same group, Codex / OpenAI Responses API sessions can fail after a session is resumed.

The client retries `POST /responses`, but some retries are routed to a different upstream OpenAI account than the account that produced the previous encrypted context. The upstream OpenAI API then rejects the request with an error like:

```text
The encrypted content ... could not be verified. Reason: Encrypted content could not be decrypted or parsed.
```

Sub2API surfaces this to the client as a `502 Bad Gateway`, so Codex Desktop shows repeated reconnect attempts.

## Expected behavior

For OpenAI Responses API / Codex requests that carry encrypted context, the same logical session should remain pinned to the same upstream OpenAI account.

## Actual behavior

Requests from the same restored Codex session can be routed across different upstream accounts in the same Sub2API group.

When the request is routed to the upstream account that can decrypt the context, it succeeds. When it is routed to another upstream account, upstream returns `400` for invalid / unverifiable encrypted content, and the relay returns `502` to the client.

## Evidence from a production incident

An anonymized production trace showed:

- Endpoint: `POST /responses`
- Client: Codex Desktop
- Same Sub2API API key and same group
- Multiple upstream OpenAI accounts available in the group
- Successful requests when routed to one upstream account
- Repeated failures when routed to a different upstream account
- Upstream status: `400`
- Relay/client status: `502`
- Upstream message: encrypted content could not be decrypted or parsed

This suggests that OpenAI Responses encrypted context is upstream-account / organization bound and cannot safely be load-balanced across accounts once a session has state.

## Suggested fix

Add OpenAI Responses session affinity for encrypted-context requests.

Possible approach:

1. Derive a stable sticky key from available Responses / Codex request signals, such as:
   - `session_id` header
   - `conversation_id`
   - `previous_response_id`
   - Codex client request metadata
   - request metadata / encrypted context presence where available

2. Bind that sticky key to the selected upstream account after the first successful request.

3. On subsequent requests, prefer the bound upstream account unless it is unavailable.

4. Avoid cross-account fallback for errors that indicate invalid encrypted content, because another upstream account is unlikely to decrypt the same context.

## Notes

The codebase already appears to have sticky-session concepts for some gateway paths. The missing behavior seems to be specifically around OpenAI Responses / Codex encrypted-context continuity in multi-account OpenAI groups.

bobobo2026:
## Summary
- Bind successful HTTP/SSE OpenAI Responses `response.id` values to the selected upstream account.
- Reuse the existing `previous_response_id` scheduler affinity store so resumed Responses/Codex turns stay on the account that created the encrypted context.
- Cover both JSON and SSE-shaped response IDs with focused unit tests.

## Why
OpenAI Responses encrypted context can fail when a resumed request with `previous_response_id` is routed to a different upstream account/org. The WS path already records `response_id -> account_id`, but the HTTP/SSE Responses path did not, so later turns could miss affinity and load-balance across accounts.

## Tests

Closes #2991

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -3040,6 +3040,7 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 		// Handle normal response
 		var usage *OpenAIUsage
 		var firstTokenMs *int
+		responseID := ""
 		imageCount := 0
 		var imageOutputSizes []string
 		if reqStream {
@@ -3049,6 +3050,7 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 			}
 			usage = streamResult.usage
 			firstTokenMs = streamResult.firstTokenMs
+			responseID = strings.TrimSpace(streamResult.responseID)
 			imageCount = streamResult.imageCount
 			imageOutputSizes = streamResult.imageOutputSizes
 		} else {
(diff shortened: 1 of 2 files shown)


bobobo2026:

bobobo2026:
recheck

PR #3000, opened 2026-06-03
fix: validate stream field type across OpenAI-compatible handlers

Pluviobyte:
## Summary
- add a shared OpenAI-compatible `stream` parser that only accepts missing or boolean `stream` values
- reject string/number `stream` values with `invalid_request_error` / `invalid stream field type`
- apply the check to gateway Responses, gateway Chat Completions, OpenAI Chat Completions, and the existing OpenAI Responses path
- add unit coverage for invalid `stream` values and boolean values continuing through the normal handler flow

## Tests
- `cd backend && go test -tags=unit ./internal/handler -count=1`
- `cd backend && go test -tags=unit ./internal/handler -run 'TestOpenAICompatibleHandlersRejectInvalidStreamFieldType|TestGatewayOpenAICompatibleHandlersAllowBooleanStreamToContinue|TestOpenAIHandler_GjsonExtraction|TestOpenAIResponses_Rejects' -count=1`

### DIFF
--- a/backend/internal/handler/gateway_handler_chat_completions.go
+++ b/backend/internal/handler/gateway_handler_chat_completions.go
@@ -75,7 +75,11 @@ func (h *GatewayHandler) ChatCompletions(c *gin.Context) {
 		return
 	}
 	reqModel := modelResult.String()
-	reqStream := gjson.GetBytes(body, "stream").Bool()
+	reqStream, ok := parseOpenAICompatibleStream(body)
+	if !ok {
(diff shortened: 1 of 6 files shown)


Pluviobyte:
Note: the only failing check here is `backend-security`, which is currently tripping on Go 1.26.3 standard-library vulnerabilities reported by `govulncheck`. I opened #3001 as a separate patch-version update to Go 1.26.4; that PR now has `backend-security` passing.

PR #3009, opened 2026-06-03
fix(images): surface real upstream error instead of generic 502

liulinhuai:
## Problem

Requests to `/v1/images/generations` and `/v1/images/edits` (e.g. `gpt-image-2`) return an opaque **502 "Upstream request failed"** for many upstream errors, hiding the real status code and message. For example, a `400 invalid_request_error` (unsupported parameter, moderation, invalid size, …) from the upstream reaches the client as a generic 502, making failures undebuggable.

## Root cause

Both image forward paths routed every **non-failover** upstream error through the generic `handleErrorResponse`, whose final `switch` collapses anything that isn't 401/402/403/429 into a hardcoded `502` with the message `"Upstream request failed"` — discarding the upstream status code, type, code, message, and param.

The sibling **Chat Completions** and **Messages** compat paths already avoid this via `handleCompatErrorResponse`, which preserves the real upstream status and message. The images paths were the only ones still masking.

## Fix

- Add `handleOpenAIImagesErrorResponse`, modeled on `handleCompatErrorResponse`. It keeps all existing side-effects (ops logging, admin error-passthrough rules, `ShouldHandleErrorCode`, account-disable / secondary failover) but surfaces the **real** upstream status code + `type`/`code`/`message`/`param`, reusing the existing `OpenAIImagesUpstreamError` + `writeOpenAIImagesUpstreamErrorResponse` machinery. Returning `OpenAIImagesUpstreamError` also lets the images handler treat it as terminal (no fallback double-write).
- Both `forwardOpenAIImagesOAuth` and `forwardOpenAIImagesAPIKey` now call it instead of `handleErrorResponse` for non-failover errors.
- Added helpers `openAIImagesUpstreamErrorFromHTTP` and `openAIImagesErrorTypeForStatus`.

**Failover behavior is unchanged** — 5xx / 401 / 403 / 429 / 529 still trigger account failover exactly as before. This only affects errors the gateway had already decided not to retry, which is the class that was being masked.

## Tests

Added `TestOpenAIGatewayServiceForwardImages_OAuthUpstreamHTTPErrorSurfacesRealError`, asserting an upstream `400` is passed through with the real status code, `type`, `code`, `param`, and message instead of a generic 502. Full `internal/service` and `internal/handler` packages pass.

### DIFF
--- a/backend/internal/service/openai_images.go
+++ b/backend/internal/service/openai_images.go
@@ -645,7 +645,7 @@ func (s *OpenAIGatewayService) forwardOpenAIImagesAPIKey(
 				RetryableOnSameAccount: account.IsPoolMode() && account.IsPoolModeRetryableStatus(resp.StatusCode),
 			}
 		}
-		return s.handleErrorResponse(upstreamCtx, resp, c, account, forwardBody)
+		return s.handleOpenAIImagesErrorResponse(upstreamCtx, resp, c, account, upstreamModel)
 	}
 	defer func() { _ = resp.Body.Close() }()
 
--- a/backend/internal/service/openai_images_responses.go
+++ b/backend/internal/service/openai_images_responses.go
@@ -579,6 +579,176 @@ func openAIImagesUpstreamErrorFromGJSON(errorObj gjson.Result, upstreamRequestID
 	}
 }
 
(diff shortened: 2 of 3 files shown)


liulinhuai:

liulinhuai:
recheck

PR #3036, opened 2026-06-04
fix(openai): preserve upstream response.failed errors

whatIsNextToTheMoon:
## Summary

     - Preserve upstream `response.failed` / protocol error responses that have already been written to the client.
     - Avoid appending a generic fallback `Upstream request failed` after the upstream terminal error was already forwarded.
     - Apply the guard to OpenAI Responses, Chat Completions, and Images forwarding paths.
     - Add regression coverage for already-communicated upstream errors.

     ## Tests

     - `go test ./internal/handler`


whatIsNextToTheMoon:

PR #3083, opened 2026-06-06
fix: force Content-Type to application/json on non-streaming responses

xzz0081:
When the gateway forces streaming to the upstream, the upstream returns Content-Type: text/event-stream. In the non-streaming branches, the body is correctly aggregated into JSON, but WriteFilteredHeaders has already passed the upstream's text/event-stream header through. Since Gin's c.JSON/c.Data only set Content-Type via writeContentType when it is not already present, the SSE header is never overwritten -- so the body is JSON while the header says SSE.

Downstream middlewares that decide "is this streaming?" by Content-Type (e.g. new-api) then mis-handle the response as SSE, breaking non-streaming JSON clients.

Fix: explicitly Set Content-Type to application/json right after WriteFilteredHeaders in the three non-streaming JSON write paths.


xzz0081:

PR #3119, opened 2026-06-08
fix(openai): avoid Codex image bridge on compact requests

liu-qingyuan:
## Summary

- Detect Codex `/responses/compact` requests by path and by compact-task payload signatures.
- Do not inject the Codex `image_generation` bridge or default coding-assistant instructions into compact requests unless the payload explicitly asks for images.
- For compact requests only, cancel upstream and stop draining after client disconnects by default, with `gateway.compact_drain_after_client_disconnect` available to opt back into drain behavior.
- Hash image response result keys instead of retaining full base64 result payloads in the dedupe key.

## Related issues

Related to compact timeout / slow header symptoms reported in:

- #2773
- #2738

This PR is not a full timeout-policy overhaul. It addresses one compact-specific amplification path: Codex compact requests being treated like normal image-capable responses requests and, after disconnects/timeouts, continuing upstream draining work by default.

## Why

Codex compact calls are internal text summarization/context-compaction requests. When they are mutated into image-generation-capable requests, they can become slower/heavier and may continue consuming upstream work even after the client has timed out or disconnected. This keeps normal non-compact streaming drain behavior unchanged for usage accounting.

## Tests

- `go test ./internal/service -run 'CodexCompact|Compact|ImageResultKey|DrainAfterClientDisconnect|ImageGenerationBridge'`
- `go test ./internal/config`

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -692,6 +692,9 @@ type GatewayConfig struct {
 	// OpenAIResponseHeaderTimeout: OpenAI/Codex 上游等待响应头的超时时间（秒），0表示无超时
 	// OpenAI/Codex 请求可能在上游排队较久；默认不使用通用响应头超时截断。
 	OpenAIResponseHeaderTimeout int `mapstructure:"openai_response_header_timeout"`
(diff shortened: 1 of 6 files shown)


liu-qingyuan:

PR #3123, opened 2026-06-08
fix: auto-grant creator access for exclusive groups

jaychoubaby:
## Summary
  - auto-grant the creator access when creating an exclusive standard group
  - pass the operator user id from the admin group create handler into CreateGroup
  - make group repository create use tx-aware ent client so group creation and allow-list insert stay in one transaction
  - add tests covering exclusive/public/subscription group creation behavior

  ## Problem
  After recent exclusive-group auth enforcement, an admin who creates an exclusive standard group can immediately lose access to that group because no corresponding `user_allowed_groups` record is
  created for the creator.

  ## Fix
  When creating an exclusive standard group, automatically insert the creator into `user_allowed_groups` in the same transaction as group creation.


jaychoubaby:
https://github.com/Wei-Shaw/sub2api/issues/3116

PR #3139, opened 2026-06-08
fix(openai): drop stale reasoning items on encrypted content retry

### LINKED ISSUE #3138: OpenAI Responses invalid_encrypted_content retry skips stale reasoning items
## Summary

Production `/v1/responses` requests can repeatedly fail with upstream OpenAI `400 invalid_encrypted_content` when the request contains stale Responses reasoning input items that no longer carry an `encrypted_content` field.

## Observed behavior

- Endpoint: `/v1/responses`
- Model: `gpt-5.5`
- Upstream error: `invalid_encrypted_content`
- Message: `The encrypted content ... could not be verified. Reason: Encrypted content could not be decrypted or parsed.`
- sub2api logs: `Skip non-WSv2 invalid_encrypted_content retry because encrypted reasoning items are missing`

## Cause

The recovery path only treated `type: "reasoning"` input items as recoverable when they contained an `encrypted_content` field. In production, OpenAI can still reject stale reasoning inputs with `invalid_encrypted_content` even when the encrypted field is absent and only summary/metadata remains. Because the recovery check returned false, sub2api skipped the one-shot cleanup retry and surfaced the upstream 400.

## Expected behavior

When OpenAI explicitly returns `invalid_encrypted_content`, the retry path should drop stale `type: "reasoning"` input items and retry once while preserving non-reasoning input items.

zhengyuliu047-rgb:
## Summary

- Drop stale Responses `reasoning` input items during `invalid_encrypted_content` recovery retries.
- Preserve non-reasoning input items while retrying once after the upstream 400.
- Update HTTP and WS regression tests to cover the cleanup behavior, including reasoning items without `encrypted_content`.

Fixes #3138

## Tests

- `go test ./internal/service -run "TestOpenAIGatewayService_Forward_HTTPRetryRecovery|TestOpenAIGatewayService_Forward_HTTPIngressRetries.*InvalidEncryptedContent|TestOpenAIGatewayService_Forward_WSv2InvalidEncryptedContent"`

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -2999,7 +2999,7 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 			upstreamMsg := strings.TrimSpace(extractUpstreamErrorMessage(respBody))
 			upstreamMsg = sanitizeUpstreamErrorMessage(upstreamMsg)
 			upstreamCode := extractUpstreamErrorCode(respBody)
-			if !httpInvalidEncryptedContentRetryTried && resp.StatusCode == http.StatusBadRequest && upstreamCode == "invalid_encrypted_content" {
+			if !httpInvalidEncryptedContentRetryTried && resp.StatusCode == http.StatusBadRequest && isOpenAIEncryptedContextErrorCode(upstreamCode) {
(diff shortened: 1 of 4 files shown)


zhengyuliu047-rgb:

zhengyuliu047-rgb:
recheck

PR #3142, opened 2026-06-08
Fix last-used account sorting for unused accounts

CyrusAuyeung:
## Summary
- Ensure admin account sorting by `last_used_at` keeps accounts that have never been used at the end.
- Add an integration regression test covering both ascending and descending last-used sorting.

## Validation
- VS Code diagnostics: no errors for changed files.
- `git diff --check`: passed.
- Go tests were not run locally because the Go toolchain is not installed in this environment.


CyrusAuyeung:

PR #3150, opened 2026-06-09
fix(gateway): apply BetaPolicy filter / block on the Vertex Anthropic path

zhangtinglu:
## Symptom

Production Vertex Anthropic `service_account` accounts started returning HTTP 400 from upstream once clients upgraded to a recent Claude Code CLI:

```
HTTP 400 invalid_request_error
Unexpected value(s) `prompt-caching-scope-2026-01-05`, `redact-thinking-2026-02-12` for the `anthropic-beta` header.
(request_id: req_vrtx_...)
```

Configuring those tokens as `filter` rules in `BetaPolicySettings` had no effect — Vertex still received them verbatim.

## Root cause

`buildUpstreamRequestAnthropicVertex` (`backend/internal/service/gateway_service.go`) is a self-contained builder that bypasses the `computeFinalAnthropicBeta` machinery used on the Anthropic-direct / API-key / count-tokens paths:

- It copies `c.Request.Header` into the upstream `req.Header` via the `allowedHeaders` whitelist
- `anthropic-beta` is in that whitelist, so the client value is forwarded as-is
- Neither `evaluateBetaPolicy` (block) nor `getBetaPolicyFilterSet` / `mergeDropSets` / `stripBetaTokensWithSet` (filter) is ever invoked on this path

This was explicitly noted as a known limitation in PR #2833 (`fix(gateway): 按最终 anthropic-beta header 对 body.context_management 做能力维度 sanitize`):

> Vertex builder 仍然不应用 BetaPolicy filter / dropSet。这是该 builder 早 return 的既有架构决策, 本 PR 不动。

This PR closes that gap.

## Why it surfaced now

Empirical inspection of the bundled `@anthropic-ai/claude-code` CLI (`grep` over `cli.js` for the literal beta strings):

| CLI version  | `prompt-caching-scope-2026-01-05` | `redact-thinking-2026-02-12` |
| ------------ | --------------------------------- | ---------------------------- |
| `≤ 2.1.22`   | not sent                          | not sent                     |
| `2.1.23+`    | **sent**                          | not sent                     |
| `2.1.69+`    | **sent**                          | **sent**                     |

Cutoffs: `2.1.23` (2026-01-29) and `2.1.69` (2026-03-04). With Vertex rejecting any unknown beta token, every CLI version from `2.1.23` onward is broken against Vertex through the gateway.

## Fix

Reuse the existing helpers inside `buildUpstreamRequestAnthropicVertex`:

1. `evaluateBetaPolicy` → block check (symmetric to `resolveBedrockBetaTokensForRequest`)
2. `mergeDropSets(getBetaPolicyFilterSet(...))` → effective drop set
3. `stripBetaTokensWithSet(clientBeta, dropSet)` → final outgoing beta
4. `sanitizeAnthropicBodyForBetaTokens(body, finalBeta)` — use the **final** beta (not the raw client value) so `body.context_management` is stripped whenever the final header lacks `context-management-2025-06-27`. This keeps header / body symmetric on the filter path too.
5. At the end of the builder, `deleteHeaderAllForms("anthropic-beta")` then `setHeaderRaw` with the final value, overwriting the raw client value that the whitelist loop wrote in.

The diff is +31 lines / -7 lines in `gateway_service.go`; the existing whitelist passthrough, URL construction, body rewrite, and debug logging are all untouched.

## Tests

New file `backend/internal/service/gateway_beta_policy_filter_test.go` with four end-to-end cases:

| Test                                                       | What it proves                                                                              |
| ---------------------------------------------------------- | ------------------------------------------------------------------------------------------- |
| `StripsTokensOnVertexServiceAccount`                       | Reproduces the prod 400 — FAILS without this PR, PASSES with it                             |
| `PassesThroughWhenEmpty`                                   | Regression guard: when no filter is configured, every client token still reaches upstream   |
| `StripsMimicryInjectedTokenOnOAuth`                        | Confirms the OAuth mimicry path already worked (and still does) — boundary check            |
| `ContextManagementBodySanitizeUsesFinalBeta`               | Verifies body sanitize is now driven by the *final* outgoing beta, not the raw client value |

All existing `gateway_anthropic_vertex_service_account_test.go` cases still pass (no regression).

```
go test ./internal/service/ -count=1   # 46s, all green
go build ./...                          # 0 errors
go vet ./...                            # 0 issues
```

## Scope deliberately not expanded

- `count_tokens` Vertex path is not touched here (it uses a different builder; same fix would apply if reported, but no symptom yet).
- The `context_management` ↔ `context-management-2025-06-27` body sanitize already lives in this builder; this PR only changes which beta value it consults (final instead of raw client).
- The default `DefaultBetaPolicySettings()` rule set is not changed — operators should add `prompt-caching-scope-2026-01-05` and `redact-thinking-2026-02-12` themselves if they want them filtered, since these tokens are required for true Anthropic OAuth accounts to avoid third-party demotion.


zhangtinglu:
撤回，需要重新整理提交范围

PR #2937, opened 2026-06-01
fix(openai): inject WS Codex image bridge tool

Kanshan03:
## Summary
- inject image_generation tool and bridge instructions for Codex CLI WS ingress requests when image generation is allowed
- reuse image-generation permission result for WS image intent checks
- add targeted ingress WS test for the Codex image bridge

## Test
- docker run --rm -v "$PWD/backend:/src:ro" -v sub2api-go-build-cache:/root/.cache/go-build -v sub2api-go-mod-cache:/go/pkg/mod -w /app -e GOPROXY=https://goproxy.cn,direct -e GOSUMDB=sum.golang.google.cn golang:1.26.3-alpine sh -c 'cp -a /src/. /app/ && go test -v -timeout=2m ./internal/service -run TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_InjectsCodexImageBridge -count=1'

PR #3160, opened 2026-06-09
fix(openai): restore Codex image bridge in WS v2 passthrough

### LINKED ISSUE #3158: bug: WS v2 passthrough mode skips Codex image bridge
## Summary

Codex image generation works on the normal WS v2 ingress path after #2937, but it is still skipped when an OpenAI account is configured to use WS v2 `passthrough` mode.

In passthrough mode, `ProxyResponsesWebSocketFromClient` returns early into `proxyResponsesWebSocketV2Passthrough` / `openai_ws_v2_passthrough_adapter.go`, so the Codex image bridge logic in `openai_ws_forwarder.go` is bypassed. The first upstream `response.create` frame is forwarded without the injected `image_generation` tool, even when:

- the request is from a Codex client,
- the group allows image generation,
- `codex_image_generation_bridge_enabled` is enabled,
- and the account has `openai_apikey_responses_websockets_v2_mode` or `openai_oauth_responses_websockets_v2_mode` set to `passthrough`.

## Impact

Codex users routed through WS v2 passthrough mode can ask for image generation, but the upstream request does not receive the `image_generation` tool bridge. This makes image generation unavailable or unreliable only for passthrough-mode accounts, while non-passthrough WS v2 accounts may work.

## Why this is different from #2937

#2937 injects the Codex image bridge in the regular WS ingress flow in `openai_ws_forwarder.go`.

Passthrough mode takes a separate path before that regular relay logic runs:

```go
case OpenAIWSIngressModePassthrough:
    return s.proxyResponsesWebSocketV2Passthrough(...)
```

That adapter path currently applies fast-policy filtering, but does not apply the same Codex image bridge mutation before forwarding `response.create` frames upstream.

## Reproduction shape

1. Configure an OpenAI account with WS v2 passthrough mode, for example:

```json
{
  "openai_apikey_responses_websockets_v2_mode": "passthrough",
  "codex_image_generation_bridge": true
}
```

2. Use a Codex client request such as:

```json
{
  "type": "response.create",
  "model": "gpt-5.1",
  "stream": false,
  "input": "generate an image"
}
```

3. Observe that the frame sent upstream by the passthrough adapter has no `tools[0].type == "image_generation"`.

## Expected behavior

WS v2 passthrough mode should apply the same Codex image bridge behavior as the regular WS v2 ingress path:

- inject `image_generation` tool for eligible Codex `response.create` frames,
- normalize explicit `image_generation` tools,
- preserve image usage/accounting metadata for generated image output.

## Local validation

I reproduced this with a regression test around `ProxyResponsesWebSocketFromClient` in passthrough mode. On current upstream main, the test fails because the passthrough result has `ImageCount == 0` and the captured upstream request lacks `tools.0.type == "image_generation"`.

After applying the local fix, these checks pass:

```sh
go test ./internal/service -run TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_PassthroughModeInjectsCodexImageBridge -count=1
go test ./internal/service -run 'TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_(PassthroughMode|PassthroughHeaders)|TestPassthroughUsageMeta|TestPassthroughBilling|TestApplyCodexImageGenerationBridgeInstructions|TestEnsureOpenAIResponsesImageGenerationTool|TestNormalizeOpenAIResponsesImageGenerationTools' -count=1
go test ./internal/service -count=1
```

A fix has been proposed in #3160.

WesleyZiwen:
## Summary

- apply the Codex image-generation bridge to WS v2 `passthrough` response.create frames
- reuse passthrough image-intent permission and billing metadata checks so generated image output is counted
- add a regression test covering Codex image bridge injection through the passthrough adapter

Fixes #3158.

## Background

#2937 fixed the regular WS v2 ingress path in `openai_ws_forwarder.go`, but accounts configured with `openai_*_responses_websockets_v2_mode=passthrough` return early into `proxyResponsesWebSocketV2Passthrough` and bypass that bridge path.

This PR applies the same bridge/policy path to the passthrough adapter before forwarding `response.create` frames upstream.

## Tests

```sh
go test ./internal/service -run TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_PassthroughModeInjectsCodexImageBridge -count=1
go test ./internal/service -run 'TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_(PassthroughMode|PassthroughHeaders)|TestPassthroughUsageMeta|TestPassthroughBilling|TestApplyCodexImageGenerationBridgeInstructions|TestEnsureOpenAIResponsesImageGenerationTool|TestNormalizeOpenAIResponsesImageGenerationTools' -count=1
go test ./internal/service -count=1
git diff --check
```

### DIFF
--- a/backend/internal/service/openai_fast_policy_ws_test.go
+++ b/backend/internal/service/openai_fast_policy_ws_test.go
@@ -994,7 +994,7 @@ func TestPassthroughUsageMeta_TracksReasoningEffortAcrossTurns(t *testing.T) {
 	firstOut, firstBlocked, firstErr := svc.applyOpenAIFastPolicyToWSResponseCreate(context.Background(), account, capturedSessionModel, firstFrame)
 	require.NoError(t, firstErr)
 	require.Nil(t, firstBlocked)
-	meta.initFromFirstFrame(firstOut)
+	require.NoError(t, meta.initFromFirstFrame(firstOut, nil))
 	require.NotNil(t, meta.reasoningEffort.Load())
 	require.Equal(t, "medium", *meta.reasoningEffort.Load())
 
@@ -1011,7 +1011,7 @@ func TestPassthroughUsageMeta_TracksReasoningEffortAcrossTurns(t *testing.T) {
 		out, blocked, policyErr := svc.applyOpenAIFastPolicyToWSResponseCreate(context.Background(), account, model, payload)
 		if policyErr == nil && blocked == nil &&
 			strings.TrimSpace(gjson.GetBytes(payload, "type").String()) == "response.create" {
-			meta.updateFromResponseCreate(out, requestModelForThisFrame)
+			require.NoError(t, meta.updateFromResponseCreate(out, requestModelForThisFrame, nil))
 		}
 		return out, blocked, policyErr
 	}
--- a/backend/internal/service/openai_ws_forwarder_ingress_session_test.go
+++ b/backend/internal/service/openai_ws_forwarder_ingress_session_test.go
@@ -711,6 +711,146 @@ func TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_PassthroughModeR
 	require.Len(t, upstreamConn.writes, 1, "passthrough 模式应透传首条 response.create")
 }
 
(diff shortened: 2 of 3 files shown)


WesleyZiwen:

PR #3168, opened 2026-06-09
feat(usage): record caller-supplied request metadata on usage_logs

jaxxjj:
## Summary

Adds an optional `metadata` (jsonb) column to `usage_logs` that captures **caller-supplied request metadata** for per-business / per-user / per-feature usage attribution.

## Motivation

The bound API key already identifies the calling integration coarsely (`api_key_id`), but operators frequently need finer, **queryable** attribution *within* a key — which end user, which feature/flow, a trace id — to break down spend and usage in the dashboard / SQL. Issuing a virtual key per dimension doesn't scale; per-request labels do.

This is a well-established gateway capability: Helicone custom properties, LiteLLM/Portkey request `metadata`, and Anthropic's own `metadata` field. It builds directly on the request-context plumbing recently added in #2865 (`client_request_id`).

## What changed

- **Capture** — a new `X-Usage-Metadata` middleware parses a JSON-object header and stashes it in the request context, propagated to the usage-record path the same way `client_request_id` is. Mounted on the same route groups (`/v1`, gemini, antigravity).
- **Persist** — recorded on `usage_logs.metadata` via the existing `buildRecordUsageLog` path; threaded through every `usage_logs` insert/scan site (single + both batch CTEs + best-effort), the `usageLogInsertArgTypes` table, and the ent schema.
- **Migration** — idempotent `150_add_usage_logs_metadata.sql` (`ADD COLUMN IF NOT EXISTS metadata jsonb`). Nullable, default NULL → zero impact on existing rows.

## Safety / bounds

Best-effort by design: a missing, oversized (>2KB), malformed, non-object, or too-large (>16 keys) header is **silently ignored** so a bad metadata header can never fail the underlying LLM request. Bounds are hardcoded constants (no config surface) to keep the column small and avoid write amplification / cardinality blowup.

## Tests

- Middleware parsing: valid object, missing, malformed, non-object, oversized, too-many-keys.
- Context propagation through `submitUsageRecordTask`.
- `scanUsageLog` round-trip including the new column (+ updated existing sqlmock insert expectations).

`go build ./...`, `gofmt`, and the touched unit packages (`internal/server/middleware`, `internal/handler`, `internal/repository`) all pass.

## Open to discussion

Happy to adjust the **naming** (`metadata` vs reusing the `extra` convention), the **header name** (`X-Usage-Metadata`), the **bounds**, or whether to also fold in body-level `metadata` (e.g. Anthropic's) — wanted to land the mechanism first.


jaxxjj:

PR #3169, opened 2026-06-09
feat(gateway): universal silent-refusal retry for streaming Claude responses

Arkptz:
## Summary

Add an optional, default-off **silent-refusal retry** for streaming
Claude-compatible responses. When an upstream turn ends as a refusal but
carries no meaningful content, the gateway transparently switches account and
retries *before* any bytes reach the client — so callers never receive an empty
or refused assistant turn.

This extends the existing OpenAI silent-refusal handling
(`openai_silent_refusal.go` + the `pendingSSE` buffered-hold) to the Claude
paths, reusing the same mechanism and the same failover loop. No new failover
infrastructure is introduced.

## Motivation

Some upstream Claude/Gemini-backed streams terminate with a refusal signal and
no usable content (a policy/safety block, not a model answer). Today that empty
turn is forwarded verbatim, which surfaces to end users as a blank or broken
response. Operators with multiple accounts would prefer the gateway to retry on
another account, exactly as it already does for OpenAI.

## What it does

Detection is keyed on the upstream's own refusal signal — **not** on
`output_tokens == 0` (which both false-positives on legitimate empty turns and
false-negatives on partial-content refusals). Three streaming paths are covered:

| Path | Detector |
|------|----------|
| Antigravity (Gemini-backed Claude) | Gemini `finishReason` in the configured blocking set (`SAFETY`, `RECITATION`, `PROHIBITED_CONTENT`, `BLOCKLIST`, `SPII`) **and** no text/thinking/tool_use |
| Native Anthropic OAuth | `stop_reason == "refusal"` **and** no meaningful content |
| Native Anthropic API-key passthrough | `stop_reason == "refusal"` **and** no meaningful content |

### Mechanism

- **Buffered hold.** While the detector is armed and no meaningful content has
  been seen, converted SSE events are buffered and the HTTP 200 / first byte is
  deferred. The hold is released on the first meaningful content
  (text/thinking/tool_use), a byte cap (`refusal_hold_max_bytes`, default
  64 KiB), or a wall-clock bound (`refusal_hold_timeout_ms`, default 10 s).
  Keepalive pings are suppressed during the hold so the client cannot observe a
  partial stream that we still intend to retry.
- **Failover.** On stream end, if no client bytes were written and the turn is a
  silent refusal, the path returns
  `UpstreamFailoverError{StatusCode: 502, RetryableOnSameAccount: false,
  SilentRefusal: true}`. The existing failover loop then switches account.
  `RetryableOnSameAccount: false` is intentional: a prompt-triggered refusal is
  deterministic on the same account, and we do not want to trip the 1-minute
  temp-ban path on an otherwise-healthy account.
- **Anti-amplification.** A dedicated cap, `FailoverState.MaxRefusalSwitches`
  (config `refusal_max_retries`, default 2, `0 = unbounded`), bounds
  refusal-driven account switches independently of the general `MaxSwitches`,
  so a deterministic policy refusal cannot burn through every account.
- A typed `UpstreamFailoverError.SilentRefusal` field replaces an earlier
  magic-string match, so the failover loop classifies refusals explicitly.

## Configuration (all under `gateway.`, default-off)

| Key | Default | Notes |
|-----|---------|-------|
| `refusal_retry_enabled` | `false` | Master switch. When off, behavior is byte-for-byte unchanged. |
| `refusal_finish_reasons` | built-in Gemini set | Antigravity-only; native Anthropic always uses `stop_reason=refusal` and ignores this. |
| `refusal_hold_timeout_ms` | `10000` | Kept within common client first-byte budgets (keepalive is suppressed during the hold). |
| `refusal_hold_max_bytes` | `65536` | Hold released unconditionally past this. |
| `refusal_max_retries` | `2` | Cap on refusal-driven switches; `0` = unbounded. |

Documented bilingually in `deploy/config.example.yaml`; the two operator-facing
env vars (`GATEWAY_REFUSAL_RETRY_ENABLED`, `GATEWAY_REFUSAL_MAX_RETRIES`) are in
`deploy/.env.example`.

## Testing

- Detector predicate matrices for native Anthropic (`anthropic_silent_refusal_test.go`)
  and antigravity (`antigravity_silent_refusal_test.go`).
- Antigravity stream-transformer signal tests — finishReason / meaningful-content
  tracking (`stream_transformer_refusal_test.go`).
- Refusal-cap behavior in the failover loop (`failover_refusal_test.go`).
- `go build ./...`, `go vet ./...`, the refusal test suite, and
  `golangci-lint run ./...` (v2.9, matching CI) all pass.

## Backward compatibility

Fully gated behind `refusal_retry_enabled` (default `false`). With the flag off,
the streaming paths, usage accounting, sticky sessions, and the
`c.Writer.Written()` failover guard behave exactly as before — the buffered hold
is never armed.

## Why mirror the OpenAI pattern

The OpenAI path already solved this problem with a buffered-hold + silent-refusal
detector + failover. Rather than invent a parallel mechanism, this PR reuses the
same shape (hold → detect → typed failover error → existing loop) for the Claude
paths, keeping the failover semantics and operator mental model consistent across
providers.

## Out of scope / known gaps

- **Non-streaming** refusal retry (this PR only touches the streaming paths).
- **Partial-content refusals** are intentionally *not* retried — once meaningful
  content has been emitted to the client, the hold is released and the turn is
  delivered as-is.
- No end-to-end integration test of the buffered hold itself yet; coverage is at
  the predicate/signal level plus the shared, already-proven OpenAI hold
  mechanism. An e2e streaming hold test is a reasonable follow-up.

## Notes

- `VERSION` is intentionally left unchanged — release tagging (`v*` →
  `release.yml`) owns versioning.
- No schema/`ent` changes; no frontend changes.

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -770,18 +770,19 @@ type GatewayConfig struct {
 	// MaxLineSize: 上游 SSE 单行最大字节数（0使用默认值）
 	MaxLineSize int `mapstructure:"max_line_size"`
 
-	// AntigravityRefusalRetry: 对 antigravity(Gemini-backed Claude) 静默拒绝
-	// (blocking finishReason 且无内容) 在切换账号后透明重试。默认关闭。
-	AntigravityRefusalRetryEnabled bool `mapstructure:"antigravity_refusal_retry_enabled"`
-	// AntigravityRefusalFinishReasons: 视为拒绝的 Gemini finishReason 列表
+	// RefusalRetryEnabled: 对上游静默拒绝（native Anthropic stop_reason=refusal
+	// 或 antigravity/Gemini blocking finishReason 且无内容）在切换账号后透明
+	// 重试。覆盖 native Anthropic 与 antigravity 两条流式路径。默认关闭。
+	RefusalRetryEnabled bool `mapstructure:"refusal_retry_enabled"`
+	// RefusalFinishReasons: antigravity(Gemini) 路径下视为拒绝的 finishReason 列表
 	// (留空使用默认 SAFETY/RECITATION/PROHIBITED_CONTENT/BLOCKLIST/SPII)
-	AntigravityRefusalFinishReasons []string `mapstructure:"antigravity_refusal_finish_reasons"`
-	// AntigravityRefusalHoldTimeoutMs: 缓冲保持上限(毫秒)，超时则释放(默认 15000)
-	AntigravityRefusalHoldTimeoutMs int `mapstructure:"antigravity_refusal_hold_timeout_ms"`
-	// AntigravityRefusalHoldMaxBytes: 缓冲字节上限，超过则释放(默认 65536)
-	AntigravityRefusalHoldMaxBytes int `mapstructure:"antigravity_refusal_hold_max_bytes"`
-	// AntigravityRefusalMaxRetries: 拒绝触发的账号切换次数上限(默认 2)
-	AntigravityRefusalMaxRetries int `mapstructure:"antigravity_refusal_max_retries"`
+	RefusalFinishReasons []string `mapstructure:"refusal_finish_reasons"`
+	// RefusalHoldTimeoutMs: 缓冲保持上限(毫秒)，超时则释放(默认 15000)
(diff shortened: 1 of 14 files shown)


Arkptz:

PR #3172, opened 2026-06-09
fix(openai-ws): keep passthrough downstream connections alive

### LINKED ISSUE #3171: [Bug] OpenAI WS v2 passthrough lacks downstream WebSocket keepalive for long Codex tasks
## Summary

OpenAI Responses WebSocket v2 `passthrough` mode can stay silent on the downstream client connection during long Codex tasks, for example Codex `image_generation`. When Sub2API is behind an idle-sensitive proxy/load balancer/CDN, the downstream WebSocket may be closed before `response.completed`, even though the upstream OpenAI WebSocket is still alive.

This is different from the existing SSE keepalive issues and from `ctx_pool` continuation issues. The affected path is the WS v2 passthrough relay.

## Observed behavior

A Codex client using Responses WebSocket v2 through Sub2API passthrough may report errors similar to:

```text
stream disconnected before completion: websocket closed by server before response.completed
```

This is easiest to reproduce with long-running image generation or other turns where the upstream can be busy for a while without producing downstream business frames.

## Root cause

The passthrough relay forwards client/upstream frames, but it did not actively keep the downstream WebSocket alive when there were no downstream business writes for a while.

If an intermediate proxy has an idle timeout, it can close the client-facing WS connection during the long upstream turn. Sub2API then sees a graceful client-side EOF/close and has to drain upstream only for usage/accounting.

## Expected behavior

For WS v2 passthrough mode, Sub2API should optionally send WebSocket Ping control frames to the downstream client after the first downstream business frame, when the downstream side has been idle for a configured interval.

Suggested behavior:

- Default interval around 20 seconds.
- Configurable timeout for waiting on Pong, around 5 seconds.
- `0` interval disables the behavior.
- If the downstream connection does not support active Ping, log/trace this and keep existing relay behavior.
- If Ping/Pong fails, treat it as a graceful client disconnect and preserve the existing upstream drain/usage capture behavior.
- Do not start downstream keepalive before the first downstream business write, so the relay does not keep never-started/failed handshakes alive artificially.

## Validation from a downstream deployment

After adding downstream WS Ping/Pong keepalive, long Codex image_generation requests behind a proxy showed repeated successful downstream keepalive traces during the same long session, for example:

```text
stage=downstream_ping_ok direction=downstream_keepalive graceful=true wrote_downstream=true
```

The same deployment continued to drain upstream usage correctly when the client connection closed.

## Proposed fix

I can open a PR with a minimal implementation that:

- Adds `Ping(ctx)` support to the passthrough downstream frame connection.
- Starts a downstream keepalive goroutine after the first downstream business write.
- Adds config keys:
  - `gateway.openai_ws.passthrough_downstream_ping_interval_seconds`
  - `gateway.openai_ws.passthrough_downstream_ping_timeout_seconds`
- Adds relay tests for:
  - ping starts only after the first downstream write;
  - successful Ping/Pong during idle periods;
  - Ping failure is handled as graceful client disconnect while preserving upstream drain;
  - idle timeout behavior is avoided while pings succeed.

## Related issues checked

Related but not exact duplicates:

- #2121: SSE keepalive before upstream response headers.
- #2031: image generation HTTP/Cloudflare timeout discussion.
- #1807: Codex WS disconnect symptom, but no downstream WS Ping/Pong root cause.
- #1769 / #2139: ctx_pool/continuation behavior, not passthrough downstream keepalive.

WesleyZiwen:
## Summary

Fixes #3171.

This adds downstream WebSocket Ping/Pong keepalive for OpenAI Responses WS v2 passthrough mode. Long Codex turns such as `image_generation` can otherwise leave the downstream client-facing WebSocket idle long enough for intermediate proxies to close it before `response.completed`.

The keepalive starts only after the first downstream business frame has been written, so failed handshakes or never-started relays are not kept alive artificially.

## Changes

- Add configurable passthrough downstream ping interval and timeout:
  - `gateway.openai_ws.passthrough_downstream_ping_interval_seconds` default `20`
  - `gateway.openai_ws.passthrough_downstream_ping_timeout_seconds` default `5`
- Add `Ping(ctx)` support to the passthrough downstream frame connection.
- Start a downstream keepalive loop for idle downstream passthrough sessions.
- Treat downstream Ping/Pong failure as a graceful client disconnect while preserving upstream drain/usage capture behavior.
- Add regression tests for successful ping, delayed start, ping failure/drain, and idle timeout behavior while pings succeed.

## Validation

```text
ok   github.com/Wei-Shaw/sub2api/internal/service/openai_ws_v2  4.314s
ok   github.com/Wei-Shaw/sub2api/internal/config                0.310s
ok   github.com/Wei-Shaw/sub2api/internal/service               10.451s
```

Commands run:

```bash
go test ./internal/service/openai_ws_v2 -count=1
go test ./internal/config -count=1
go test ./internal/service -run 'OpenAIWS|Passthrough' -count=1
```

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -917,6 +917,10 @@ type GatewayOpenAIWSConfig struct {
 	EventFlushBatchSize int `mapstructure:"event_flush_batch_size"`
 	// EventFlushIntervalMS: WS 流式写出最大等待时间（毫秒）；0 表示仅按 batch 触发
 	EventFlushIntervalMS int `mapstructure:"event_flush_interval_ms"`
+	// PassthroughDownstreamPingIntervalSeconds: passthrough 模式下游 WS Ping 间隔；0 表示关闭。
+	PassthroughDownstreamPingIntervalSeconds int `mapstructure:"passthrough_downstream_ping_interval_seconds"`
+	// PassthroughDownstreamPingTimeoutSeconds: passthrough 模式下游 WS Ping 等待 Pong 的超时时间。
+	PassthroughDownstreamPingTimeoutSeconds int `mapstructure:"passthrough_downstream_ping_timeout_seconds"`
 	// PrewarmCooldownMS: 连接池预热触发冷却时间（毫秒）
 	PrewarmCooldownMS int `mapstructure:"prewarm_cooldown_ms"`
 	// FallbackCooldownSeconds: WS 回退冷却窗口，避免 WS/HTTP 抖动；0 表示关闭冷却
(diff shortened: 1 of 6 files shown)


ryan4yin:
Any updates on this? We're running into the same errors and think this PR is a good fix.

PR #3234, opened 2026-06-11
fix(account-stats): avoid double counting image output tokens

hehe-wz:
## Summary
- split image output tokens from text output tokens when computing account stats cost
- keep image output tokens billed with the dedicated image output token price
- add regression coverage for model-file and custom account-stats pricing

## Why
GPT image usage reports output tokens that can be entirely image output. The old account stats formula multiplied the full output token count by text output price and then added image output cost again.

## Validation
- go test -tags=unit ./internal/service -run 'TestTryModelFilePricing_ImageOutputTokensAreNotDoubleCountedAsTextOutput|TestCalculateStatsCost_TokenBilling_WithImageOutput' -count=1


hehe-wz:

PR #3235, opened 2026-06-11
fix(openai): price image input tokens separately

hehe-wz:
## Summary
- parse OpenAI image input token usage from usage.input_tokens_details.image_tokens
- store image_input_tokens and image_input_cost on usage logs
- price image input tokens with model-file or channel image_input_price rates
- expose image input token/cost details in admin and user usage views
- add channel/account-stats image_input_price configuration support

## Why
OpenAI image edit responses can include image input tokens that have a different rate from text input tokens. Without persisting and pricing that field separately, GPT image edit cost diverges from upstream pricing.

## Stacking note
This branch is stacked on #3234 because it reuses the image-output token split introduced there. After #3234 merges, this PR should show only the image-input token changes.

## Validation
- go test -tags=unit ./internal/service ./internal/repository ./internal/handler/admin ./internal/handler ./internal/server -count=1
- git diff --check

Frontend typecheck could not be run in the local checkout because pnpm failed before invoking scripts with ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING and vue-tsc is missing from node_modules/.bin.


hehe-wz:
recheck

PR #3237, opened 2026-06-11
fix(openai): avoid failover on context-window errors

### LINKED ISSUE #3236: OpenAI context-window overflow errors should not trigger account failover
## Problem

When an OpenAI-compatible upstream rejects a very large request because it exceeds the model context window, sub2api currently treats the upstream response as failover-worthy if the status is 5xx, especially HTTP 502.

Observed upstream body shape:

```json
{
  "error": {
    "message": "Your input exceeds the context window of this model. Please adjust your input and try again.",
    "type": "upstream_error",
    "code": null
  }
}
```

This failure is deterministic for the request payload/model pair. Retrying other accounts does not help; it only consumes retry budget, delays the client response, and may mark otherwise healthy accounts as problematic.

## Expected behavior

For context-window overflow errors, sub2api should return the upstream error to the client without wrapping it as `UpstreamFailoverError` and without triggering account failover/retry side effects.

## Suggested priority

I strongly recommend treating this as high priority. In long-context scenarios, a single oversized request can otherwise burn through the configured fallback accounts before returning the same deterministic error.

## Proposed fix

I have opened a PR that:

- detects context-window overflow messages/codes with narrow markers;
- bypasses generic 5xx failover for those OpenAI upstream responses;
- applies the same non-failover classification to `response.failed` SSE events;
- avoids account scheduling/block side effects for OpenAI context-window bodies;
- preserves transient processing/capacity failures as failover-worthy.

## Validation

Local validation on Go 1.26.4:

- `gofmt -w` on modified files
- `git diff --check`
- `go test -tags=unit ./internal/service`
- `go test -tags=unit ./...`

dftian478:
## Summary

This PR makes OpenAI context-window overflow errors non-failoverable.

For errors such as:

```json
{
  "error": {
    "message": "Your input exceeds the context window of this model. Please adjust your input and try again.",
    "type": "upstream_error",
    "code": null
  }
}
```

retrying another account is deterministic waste: the same oversized payload/model pair will fail again. I strongly recommend handling this at highest priority for long-context traffic, because the current generic 5xx failover behavior can burn through many configured accounts before returning the same useful error to the client.

## Changes

- Add narrow OpenAI context-window error detection.
- Bypass generic 5xx failover for matching HTTP upstream responses.
- Treat matching `response.failed` SSE events as non-failoverable.
- Preserve the upstream context-window message in non-failover error responses.
- Skip OpenAI account scheduling/block side effects for context-window bodies.
- Keep transient OpenAI processing/capacity failures failoverable.

## Validation

Run locally with Go 1.26.4:

- `gofmt -w backend/internal/service/openai_gateway_service.go backend/internal/service/openai_gateway_chat_completions.go backend/internal/service/openai_account_runtime_block_fastpath.go backend/internal/service/error_passthrough_runtime_test.go backend/internal/service/openai_gateway_chat_completions_test.go backend/internal/service/openai_gateway_service_test.go backend/internal/service/openai_gateway_service_codex_cli_only_test.go`
- `git diff --check`
- `go test -tags=unit ./internal/service`
- `go test -tags=unit ./...`

Fixes #3236

### DIFF
--- a/backend/internal/service/error_passthrough_runtime_test.go
+++ b/backend/internal/service/error_passthrough_runtime_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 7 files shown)


dftian478:

dftian478:
Hi maintainers, this PR is currently blocked because the CI and Security Scan workflows are in `action_required` state for this forked PR.

Could someone please approve the workflow runs when convenient? The change is small and targets a context-window overflow case where retrying/failover is not useful.

Related issue: #3236

Thanks!

dftian478:
Adding context on how this relates to the already-merged #2572 / #2556 fix.

#2572 handled the case where a GPT-5.5 long-context overflow could come back as an empty successful stream (`finish_reason=stop`, no usage/error/content). In that shape, treating it as a failure and entering the existing failover path makes sense because the upstream did not provide a deterministic context-window error to surface.

This PR covers the complementary explicit-error case: when upstream clearly reports that the request exceeds the model context window, retrying or rotating to another account is deterministic waste because the same oversized payload/model pair will fail again. In that case, the context-window error should be returned to the client instead of triggering account failover.

dftian478:
已基于最新 main 重新整理为中文提交并重新提交，关闭旧 PR。

PR #3308, opened 2026-06-16
fix(gemini): clean unsupported tool schema fields

### LINKED ISSUE #3143: Gemini compat: cleanToolSchema doesn't flatten JSON-Schema `type` arrays or strip `$defs` → Gemini 400 INVALID_ARGUMENT on tool calls
## TL;DR / 摘要

通过 **Anthropic `/v1/messages` → Gemini** 兼容路径转发带工具(tools)的请求时，如果工具的 JSON Schema 含有联合类型 `"type": ["string","null"]` 或 `$defs`/`$ref`，Google 会返回 `400 INVALID_ARGUMENT`。同样的工具发到 OpenAI/Anthropic 上游都正常——只有 Gemini 更严格的 `function_declarations` 格式会拒绝，而 `cleanToolSchema` 没有把它们清理干净。

## Environment

- Image: `weishaw/sub2api:latest` (source inspected at `a39163`).
- Path: `gemini_messages_compat_service.go` (Anthropic `/v1/messages` → Gemini). The OpenAI `/v1/chat/completions` → Gemini path is affected too, since it funnels through the same `convertClaudeMessagesToGeminiGenerateContent`.
- Account: Gemini OAuth (`code_assist`, tier `gcp_standard`).
- Client: an agent (Goose) that emits JSON-Schema tool definitions with union types and `$defs`/`$ref`.

## Error (gateway log)

```
[Gemini] upstream error 400: Invalid JSON payload received.
Unknown name "type" at 'request.tools[0].function_declarations[0].parameters.properties[0].value': Proto field is not repeating, cannot start list.
Unknown name "$defs" at 'request.tools[0].function_declarations[8].parameters': Cannot find field.
status: INVALID_ARGUMENT
```

## Root cause

`backend/internal/service/gemini_messages_compat_service.go` → `cleanToolSchema` (~L3375):

1. **Union `type` arrays are not flattened.** It only normalizes when `type` is a string:
   ```go
   if typeVal, ok := cleaned["type"].(string); ok {
       cleaned["type"] = strings.ToUpper(typeVal)
   }
   ```
   A `[]any` like `["string","null"]` passes through unchanged → Gemini: *"Proto field is not repeating, cannot start list."*
2. **`$defs` / `definitions` are not dropped.** The skip-list drops `$ref` but not `$defs`/`definitions`, so a sibling `$defs` object survives → Gemini: *"Unknown name `$defs`: Cannot find field."*

## Suggested fix

```go
// cleanToolSchema 清理工具的 JSON Schema，移除 Gemini 不支持的字段
func cleanToolSchema(schema any) any {
	if schema == nil {
		return nil
	}
	switch v := schema.(type) {
	case map[string]any:
		cleaned := make(map[string]any)
		for key, value := range v {
			if key == "$schema" || key == "$id" || key == "$ref" ||
				key == "$defs" || key == "definitions" || // NEW: drop $defs/definitions
				key == "additionalProperties" || key == "patternProperties" || key == "minLength" ||
				key == "maxLength" || key == "minItems" || key == "maxItems" {
				continue
			}
			cleaned[key] = cleanToolSchema(value)
		}
		// Normalize `type`: Gemini wants a single uppercase scalar, not a union array.
		switch t := cleaned["type"].(type) {
		case string:
			cleaned["type"] = strings.ToUpper(t)
		case []any: // NEW: collapse ["string","null"] to the first non-null type
			picked := ""
			for _, item := range t {
				if s, ok := item.(string); ok && !strings.EqualFold(s, "null") {
					picked = strings.ToUpper(s)
					break
				}
			}
			if picked != "" {
				cleaned["type"] = picked
			} else {
				delete(cleaned, "type")
			}
		}
		return cleaned
	case []any:
		cleaned := make([]any, len(v))
		for i, item := range v {
			cleaned[i] = cleanToolSchema(item)
		}
		return cleaned
	default:
		return v
	}
}
```

These two changes map exactly to the two log errors above.

## Further hardening (optional)

- Dropping `$ref` without inlining leaves an empty `{}` schema (Gemini generally treats it as ANY; inlining the ref against the root `$defs` would preserve typing).
- Gemini's `function_declarations` also reject `anyOf` / `oneOf` / `allOf` / `const`. Worth collapsing/stripping those too so other agents' tool sets don't hit a fresh `INVALID_ARGUMENT`.

## Repro

1. Use a key bound to a Gemini (`code_assist`) account.
2. POST `/v1/messages` with a tool whose `input_schema.properties.X.type` is `["string","null"]`, **or** whose `input_schema` contains a `$defs` block → `400 INVALID_ARGUMENT`.
3. The same tool payload succeeds against OpenAI / Anthropic upstreams.

a0yark:
## Summary
- Closes #3143 by removing Gemini-unsupported `$defs` and `definitions` fields from converted tool schemas.
- Normalizes JSON Schema union `type` arrays, such as `["string", "null"]`, to the first non-null Gemini scalar type.
- Adds a focused unit test for the schema cleaner behavior.

## Test Plan
- [x] go test ./internal/service -run TestCleanToolSchema_NormalizesGeminiUnsupportedSchemaFields
- [x] go test ./internal/service -run "Test(ConvertClaudeToolsToGeminiTools|CleanToolSchema)"

### DIFF
--- a/backend/internal/service/gemini_messages_compat_service.go
+++ b/backend/internal/service/gemini_messages_compat_service.go
@@ -3416,6 +3416,7 @@ func cleanToolSchema(schema any) any {
 		for key, value := range v {
 			// 跳过不支持的字段
 			if key == "$schema" || key == "$id" || key == "$ref" ||
+				key == "$defs" || key == "definitions" ||
 				key == "additionalProperties" || key == "patternProperties" || key == "minLength" ||
 				key == "maxLength" || key == "minItems" || key == "maxItems" {
 				continue
@@ -3426,6 +3427,17 @@ func cleanToolSchema(schema any) any {
 		// 规范化 type 字段为大写
 		if typeVal, ok := cleaned["type"].(string); ok {
 			cleaned["type"] = strings.ToUpper(typeVal)
+		} else if typeValues, ok := cleaned["type"].([]any); ok {
+			for _, typeValue := range typeValues {
+				typeName, ok := typeValue.(string)
+				if ok && !strings.EqualFold(typeName, "null") {
+					cleaned["type"] = strings.ToUpper(typeName)
+					break
+				}
+			}
+			if _, ok := cleaned["type"].([]any); ok {
+				delete(cleaned, "type")
+			}
 		}
 		return cleaned
 	case []any:
--- a/backend/internal/service/gemini_messages_compat_service_test.go
+++ b/backend/internal/service/gemini_messages_compat_service_test.go
@@ -323,9 +323,18 @@ func TestCleanToolSchema_NormalizesGeminiUnsupportedSchemaFields(t *testing.T) {
 
 	properties, ok := cleaned["properties"].(map[string]any)
 	require.True(t, ok)
-	require.Equal(t, "STRING", properties["path"].(map[string]any)["type"])
-	require.Equal(t, "INTEGER", properties["count"].(map[string]any)["type"])
-	require.NotContains(t, properties["empty"].(map[string]any), "type")
+
+	pathSchema, ok := properties["path"].(map[string]any)
+	require.True(t, ok)
+	require.Equal(t, "STRING", pathSchema["type"])
+
+	countSchema, ok := properties["count"].(map[string]any)
+	require.True(t, ok)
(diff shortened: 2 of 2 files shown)


a0yark:

PR #3319, opened 2026-06-17
feat(payment): add balance packages and flexible redeem validity

laozuzhen:
## Summary
- Add balance-package support so wallet balance can be granted with an expiry and deducted before permanent wallet balance.
- Allow redeem-code validity to be represented with sub-day precision and apply it to subscription/balance package redemption flows.
- Extend payment catalog/order handling, XunhuPay provider support, tiered balance pricing, and related admin/user payment UI.

## Tests
- `go test ./internal/service ./internal/payment ./internal/handler ./internal/handler/admin ./internal/server/routes -run 'Payment|Redeem|BalancePackage|Subscription|Xunhu|Plan|Fulfillment|Order'`

## Notes
- Frontend build was not run in the isolated worktree because `frontend/node_modules` is not installed there.


laozuzhen:
recheck

PR #3324, opened 2026-06-17
[codex] defer admin tour until compliance is ready

xwil1:
## Summary
- defer the admin onboarding tour until admin compliance status has loaded
- prevent the tour from starting while the compliance acknowledgement dialog is required
- add regression coverage for first-login compliance gating

## Root cause
The admin tour scheduled its auto-start as soon as `AppLayout` mounted. On first admin login, the compliance acknowledgement status is loaded asynchronously, so the tour could call `drive(0)` before the required compliance dialog finished.

## Validation
- `cd frontend && ./node_modules/.bin/vitest run src/composables/__tests__/useOnboardingTour.spec.ts`
- `cd frontend && ./node_modules/.bin/vue-tsc --noEmit`
- `cd frontend && ./node_modules/.bin/eslint src/composables/useOnboardingTour.ts src/composables/__tests__/useOnboardingTour.spec.ts`
- `git diff --check`

### DIFF
--- a/frontend/src/composables/__tests__/useOnboardingTour.spec.ts
+++ b/frontend/src/composables/__tests__/useOnboardingTour.spec.ts
@@ -1,0 +1,153 @@
+import { mount } from '@vue/test-utils'
+import { createPinia, setActivePinia } from 'pinia'
+import { defineComponent, nextTick } from 'vue'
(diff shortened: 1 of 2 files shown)


xwil1:

xwil1:
recheck

PR #3326, opened 2026-06-17
fix(deploy): add :Z SELinux labels to bind mounts in compose files

tairan:
see description


tairan:

PR #3414, opened 2026-06-22
feat: improve model marketplace display

laozuzhen:
## Summary
- Reworks the available channels page into a model marketplace with group-first browsing, channel fallback, model icons, copy actions, and responsive layout.
- Extends available-channel responses with group image pricing tiers and displays original/effective prices using recharge and channel multipliers.
- Includes account-mapped/direct group models when channel model restrictions allow it, with tests for the expanded model listing behavior.

## Test Plan
- [x] pnpm build
- [x] go test ./internal/service ./internal/handler ./internal/server -count=1
- [x] go build ./cmd/server


laozuzhen:
我已阅读CLA文件，特此签署CLA

PR #3442, opened 2026-06-23
fix(openai): fail over on chat completions transport errors

hsn8086:
## Summary
- Route OpenAI chat completions transport errors through the existing upstream transport-error handler.
- Return UpstreamFailoverError instead of writing a hard 502, so the caller can fail over to another account.
- Add unit coverage for raw chat completions transport-error failover.

## Why
/responses already converts transport errors into failover behavior. The chat completions paths still wrote 502 directly, which bypassed account failover even when other accounts were available.

## Testing
- GOCACHE=/tmp/sub2api-go-cache GOMODCACHE=/tmp/sub2api-go-mod go test ./internal/service -run 'Test(HandleOpenAIUpstreamTransportError|ForwardAsRawChatCompletions_TransportErrorFailsOver)' -tags unit


hsn8086:

PR #3494, opened 2026-06-26
docs: note admin account wizard requirement in source-compile install

devnomad-byte:
Method 3 (source compile) instructs users to pre-create config.yaml, but does not mention that doing so disables the setup wizard — the only code path that creates the initial admin account. Users following the README end up with a running server and an empty users table, and login fails with "invalid email or password".

The default.admin_email / default.admin_password fields in config.yaml are loaded into the config struct but never read by any code path (setup.SetupConfig.Admin is tagged yaml:"-"), so they cannot serve as an alternative way to seed the admin.

Add a "Creating the Admin Account" subsection to Method 3 in all three language READMEs (EN/CN/JA), explaining the wizard dependency and two workarounds: let the wizard generate config.yaml, or temporarily move config.yaml aside to trigger the wizard on first run.

Related: #521 (docker variant of the same symptom), #2617, #2350.


devnomad-byte:

PR #3537, opened 2026-06-28
Add persistent reset credit expiry display

rick147:
## Summary
- Add a reset credit expiry column for OpenAI OAuth account usage.
- Query and cache reset credit count plus up to five expiry times when clicking Credits.
- Preserve cached reset credit count and expiry times across page refreshes.
- Keep reset credit expiry values sorted by expiration time.

## Notes
- Cached values are only refreshed by clicking the Credits button.
- Reset-credit cache fields are treated as scheduler-neutral extra fields.


rick147:

PR #3577, opened 2026-06-30
fix: remove Chinese strings from English UI

### LINKED ISSUE #1205: English locale shows Chinese text "倍率" on Keys page
In the user `Keys` page, English locale still shows Chinese text like `1x 倍率`.

This appears to come from a hardcoded string in:

`frontend/src/components/common/GroupOptionItem.vue`

Expected: English UI should not display Chinese text.

Image

heathermhuang:
## Summary
- Fix Chinese text leaking into English UI while preserving Chinese UI localization.
- Localize the affected Groups, account, channel pricing, login agreement, risk control, custom page, ops, DingTalk, and preset-label paths through locale keys.
- Address review-blocking regressions by localizing Ops System Logs and channel interval validation instead of hardcoding English.
- Remove the remaining English fallback strings introduced by this PR in channel and OAuth paths.

Fixes #1205

## Verification
- frontend/node_modules/.bin/vue-tsc --noEmit
- CI=1 frontend/node_modules/.bin/vitest run src/components/admin/channel/__tests__/types.spec.ts --cache=false --pool=threads --poolOptions.threads.singleThread=true --fileParallelism=false --reporter=verbose
- git diff --check
- Source scan: no added hardcoded Ops System Logs labels from the review remain in the PR diff.
- Source scan: no added English fallback arguments remain in the channel pricing, channel submit, or OAuth paths touched by this PR.


Wei-Shaw:
### Blocking i18n regressions

  1. `frontend/src/views/admin/ops/components/OpsSystemLogTable.vue`
     - Many Chinese UI strings were replaced with hardcoded English strings, such as `System Logs`, `Runtime Log
     Configuration`, `Search`, `Reset`, `Clean current filters`, confirmation dialogs, and toast messages.
     - This component does not import `useI18n()` and no corresponding locale keys were added.
     - As a result, Chinese users will see an almost entirely English System Logs page, which is the opposite of this
     PR’s stated i18n goal.

  2. `frontend/src/components/admin/channel/types.ts`
     - `validateIntervals()` now returns hardcoded English validation messages.
     - These messages are user-visible: `frontend/src/views/admin/ChannelsView.vue:1501-1505` passes them directly to
     `appStore.showError(...)`.
     - As a result, channel pricing interval validation errors will be shown in English even in Chinese sessions.

  3. Channel pricing components still rely on English fallbacks
     - `IntervalRow.vue` and `PricingEntryCard.vue` changed many calls from `t(key, 'Chinese fallback')` to `t(key,
     'English fallback')`.
     - If the locale key is missing or locale messages are not available yet, the Chinese UI will fall back to English.
     - A safer approach is to add complete `en` and `zh` locale keys and avoid replacing user-visible Chinese fallback
     text with English-only fallback text.

heathermhuang:
@Wei-Shaw I addressed the blocking i18n regressions in the follow-up commits.

Summary:
- Ops System Logs now uses i18n keys in both en/zh instead of hardcoded user-facing labels.
- Channel interval validation now requires the translator and returns localized messages.
- Channel pricing/channel submit/OAuth paths touched by this PR no longer rely on the English fallback strings introduced earlier; the missing en/zh locale keys were added.

Verification:
- frontend/node_modules/.bin/vue-tsc --noEmit
- CI=1 frontend/node_modules/.bin/vitest run src/components/admin/channel/__tests__/types.spec.ts --cache=false --pool=threads --poolOptions.threads.singleThread=true --fileParallelism=false --reporter=verbose
- git diff --check
- GitHub CI is green on the latest commit.

Could you please re-review when you have a chance?

PR #3596, opened 2026-07-01
fix: source-build version resolution for release tags

JRBaggins:
## Summary

- prefer the exact Git release tag over `backend/cmd/server/VERSION` when resolving the source-build version
- wire the same version resolution into the main backend/deploy build entrypoints so manual source builds, Docker builds, and embedded frontend builds report the tagged version consistently
- update the EN/CN/JA source-build README snippets to use the resolved build version instead of relying on the stale file value inside older tags

When building directly from a release tag checkout, the embedded `VERSION` file can lag by one release because the sync back to `main` happens after the tag is cut. For example, `v0.1.140` still contains `backend/cmd/server/VERSION=0.1.139`, which makes `--version` and the sidebar version badge display the previous release for source builds from that tag.

This keeps the existing fallback order for non-tagged checkouts:

- explicit build arg / ldflags
- exact Git tag
- `backend/cmd/server/VERSION`

## Testing

```bash
cd backend
./scripts/resolve-version.sh
make build
./bin/server --version
```

```bash
# tag checkout reproduction
cd backend
./scripts/resolve-version.sh
VERSION="$(./scripts/resolve-version.sh)"
CGO_ENABLED=0 go build -ldflags="-s -w -X main.Version=${VERSION}" -o /tmp/sub2api-v0140 ./cmd/server
/tmp/sub2api-v0140 --version
```

Verified locally:

- `origin/main` resolves/builds as `0.1.141`
- `v0.1.140` resolves/builds as `0.1.140` even though the tag's `backend/cmd/server/VERSION` file is still `0.1.139`
- `golangci-lint v2.9.0` reports `0 issues` for `./internal/pkg/anthropicfp` after rebasing onto latest `main`

## Notes

- this does not rewrite historical tags or change the release workflow itself
- release CI keeps its existing explicit `VERSION` / ldflags precedence
- the README changes only keep the documented manual source-build commands aligned with the new resolution path
- after rebasing onto the latest `main`, this branch also includes a small lint-only fix in `backend/internal/pkg/anthropicfp/dateline.go`; that file came from upstream `main` during rebase and would otherwise keep this PR red even though it is unrelated to the tagged-version logic

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,6 +119,8 @@ tests
 CLAUDE.md
 .claude
 scripts
+!backend/scripts/
+!backend/scripts/resolve-version.sh
 .code-review-state
 #openspec/
 code-reviews/
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,9 +67,9 @@ COPY backend/ ./
 COPY --from=frontend-builder /app/backend/internal/web/dist ./internal/web/dist
 
 # Build the binary (BuildType=release for CI builds, embed frontend)
-# Version precedence: build arg VERSION > cmd/server/VERSION
+# Version precedence: build arg VERSION > exact git tag > cmd/server/VERSION
 RUN VERSION_VALUE="${VERSION}" && \
-    if [ -z "${VERSION_VALUE}" ]; then VERSION_VALUE="$(tr -d '\r\n' < ./cmd/server/VERSION)"; fi && \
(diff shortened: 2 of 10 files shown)


Wei-Shaw:
backend/internal/pkg/anthropicfp/dateline.go 同步一下main分支

JRBaggins:

JRBaggins:
> backend/internal/pkg/anthropicfp/dateline.go 同步一下main分支

done

PR #3609, opened 2026-07-01
fix: Handle invalid arguments correctly for Gemini reasoning models

phoenixikkifullstack:
### Summary
This PR resolves `INVALID_ARGUMENT` errors when proxying requests to Gemini reasoning/thinking models (e.g., `gemini-3.1-pro-high`) through the Antigravity gateway. 

Gemini reasoning models strictly reject requests that include generation configs with sampling parameters (`temperature`, `top_p`, `top_k`), `stop_sequences`, or empty `toolConfig` objects.
Previously, passing these fields would cause the upstream API to respond with a 400 Bad Request.

### Changes Made
- Introduced an `IsReasoning` boolean flag to the internal `modelDef` structs inside `claude_types. go` to explicitly tag thinking models in a structured way.
- Added a new helper function `IsGeminiReasoningModel(modelID string) bool` to perform model capability checks without relying on hardcoded name string matching across the codebase.
- Refactored `request_transformer.go` to intercept and strip out `Temperature`, `TopP`, `TopK`, `StopSequences`, and empty forced `ToolConfig`s during payload conversion when a reasoning model is detected.

### Impact
Ensures that all current and future Gemini reasoning models properly generate responses through the Antigravity gateway without triggering `INVALID_ARGUMENT` exceptions from the upstream API.

### DIFF
--- a/backend/internal/pkg/antigravity/claude_types.go
+++ b/backend/internal/pkg/antigravity/claude_types.go
@@ -1,6 +1,9 @@
 package antigravity
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"strings"
+)
 
 // Claude 请求/响应类型定义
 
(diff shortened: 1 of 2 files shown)


phoenixikkifullstack:

PR #3653, opened 2026-07-02
fix: use chatgpt_user_id for OpenAI Codex import matching

lovisnd:
## Summary
- prefer `chatgpt_user_id` when building Codex session import identity keys
- only fall back to `chatgpt_account_id` when `chatgpt_user_id` is missing
- add regression coverage for multiple users in the same ChatGPT team/account

## Why
Users in the same ChatGPT team can share `chatgpt_account_id`. The previous import matching included that shared account key before the user key, so importing another team member could update the first account instead of creating a separate account.

## Test
- `go test -count=1 ./internal/handler/admin`


lovisnd:

Wei-Shaw:
https://github.com/Wei-Shaw/sub2api/commit/a5638a4e54088192f5be408ae1b36d9083626c06

PR #3695, opened 2026-07-03
Support Anthropic model-scoped weekly limits (Fable 5): usage display + scoped 429 handling

tc-imba:
## Summary

Anthropic moved model-scoped weekly limits (e.g. the **Fable 5 / Opus-tier** window) out of the deprecated top-level `seven_day_opus` / `seven_day_sonnet` fields (now always `null`) and into a `limits[]` array on `/api/oauth/usage`, with matching `anthropic-ratelimit-unified-*` response headers on 429s. This PR teaches sub2api about that shape in two places.

### 1. Show model-scoped 7d windows on the Accounts page (`feat`)
- Parse `limits[]` `weekly_scoped` entries into a new `seven_day_scoped[]` (`ScopedUsageWindow{label, utilization, resets_at}`), labelled by `scope.model.display_name` (e.g. "Fable").
- Render each as a `7d {label}` bar (e.g. **7d Fable**) in `AccountUsageCell`.
- Future-proof: any model Anthropic scopes shows up automatically, not just Fable.

### 2. Scope model-limited 429s to the requested model (`fix`)
- Previously a Fable 429 (scoped window exhausted, general 5h/7d still `allowed`) set an **account-wide** rate limit — pausing scheduling for *every* model on the account, and mis-timed to the 5h reset.
- Now detect that case from the 429 headers (`unified-status: rejected` while neither 5h nor 7d is rejected/exceeded) and rate-limit **only the requested model** via the existing per-model mechanism (`Extra.model_rate_limits`, already consulted by `IsSchedulableForModelWithContext`). General 5h/7d exhaustion keeps the account-wide behavior.
- Model key uses `GetMappedModel` so the stored key matches what the scheduler recomputes. Detection keys off the stable 5h/7d headers, not the internal `7d_oi` window name.

## Verification
- New Go unit tests built against the **real** `/api/oauth/usage` + 429 headers captured from `api.anthropic.com` (fable→model-only limit; general 7d→account-wide; no-model→fallback; detection/reset helpers). `go build ./...` + existing Anthropic rate-limit tests pass.
- Live check: a Fable request 429s while Sonnet/Opus on the same account return 200; the "7d Fable" bar renders in the admin UI after an active usage query.

## Notes
- Passive usage sampling can't show scoped windows (Anthropic doesn't emit them in the per-request rate-limit headers) — they appear via the active "Query" only, matching the existing 5h/7d passive behavior.

## Image
image

### DIFF
--- a/backend/internal/repository/claude_usage_service_test.go
+++ b/backend/internal/repository/claude_usage_service_test.go
@@ -41,7 +41,13 @@ func (s *ClaudeUsageServiceSuite) TestFetchUsage_Success() {
 		_, _ = io.WriteString(w, `{
   "five_hour": {"utilization": 12.5, "resets_at": "2025-01-01T00:00:00Z"},
   "seven_day": {"utilization": 34.0, "resets_at": "2025-01-08T00:00:00Z"},
-  "seven_day_sonnet": {"utilization": 56.0, "resets_at": "2025-01-08T00:00:00Z"}
+  "seven_day_sonnet": {"utilization": 56.0, "resets_at": "2025-01-08T00:00:00Z"},
+  "seven_day_opus": null,
+  "limits": [
+    {"kind": "session", "group": "session", "percent": 12, "severity": "normal", "resets_at": "2025-01-01T00:00:00Z", "scope": null, "is_active": false},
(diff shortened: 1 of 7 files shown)


tc-imba:

Wei-Shaw:
https://github.com/Wei-Shaw/sub2api/pull/3698

PR #3734, opened 2026-07-06
fix(i18n): correct misleading "irreversible" description for revoke action

Yinr:
## Summary
Updated the i18n descriptions for the revoke action in both `en.ts` and `zh.ts` to accurately reflect that the subscription is restorable.

## Impact
- UI text update only.
- No business logic or functional changes.


Yinr:

PR #3838, opened 2026-07-08
fix(gateway): fail over on Anthropic transport-layer errors + add connect timeouts

hongheshan-svg:
## Problem

The Anthropic-family forward paths — native OAuth, API-key passthrough, Bedrock, `ForwardAsChatCompletions`, `ForwardAsResponses` — write a hard **502** to the client when `Do`/`DoWithTLS` returns a transport-layer error (proxy / DNS / TCP / TLS failure, no HTTP status code) and return a plain error. They neither retry nor fail over, **even when healthy accounts remain**. A single `connection reset` / `EOF` / proxy blip becomes a user-visible 502.

The OpenAI `/responses` path already handles this correctly (`handleOpenAIUpstreamTransportError`, commit 217f8599). This PR brings the Anthropic paths to parity by reusing the same mechanism.

Separately, the main upstream `Transport` had **no dial or TLS-handshake timeout**, and the TLS-fingerprint dialers used a context-unaware SOCKS5 `Dial` plus a deadline-less HTTP `CONNECT` read — so a blackholed upstream/proxy could hang a connection indefinitely (the request `ctx` usually carries no deadline).

## Changes

**Transport-error failover** — new `handleAnthropicUpstreamTransportError` (mirrors the OpenAI helper, reuses `classifyOpenAITransportError`):
- transport errors → `*UpstreamFailoverError` so the handler switches accounts;
- durable faults (rejected proxy creds / connection refused / DNS failure) → `SetTempUnschedulable` for 10m;
- `context.Canceled` (client gone) → no failover, no eviction;
- does **not** write the response (the handler owns it — failover, or a protocol-correct error once exhausted).
- Wired into all 5 Anthropic call sites.

**Connection-establishment timeouts** (bound connection setup only; do **not** affect streaming):
- `buildUpstreamTransport`: `DialContext` 15s + KeepAlive 30s + `TLSHandshakeTimeout` 15s, plus explicit `ForceAttemptHTTP2` so a custom `DialContext` doesn't silently disable H2 on the direct / HTTP-proxy paths (SOCKS5 stays HTTP/1.1 as before);
- `tlsfingerprint` dialers: 30s establishment envelope; SOCKS5 switched from `Dial` to `DialContext`; HTTP `CONNECT` phase bounded by a conn deadline cleared before the handshake;
- `proxyutil` SOCKS5 tunnel: 30s establishment envelope.

## Tests

- transport-error failover behaviour across all 5 paths;
- classify/handle helper: persistent → evict + fail over; transient → fail over without eviction; `context.Canceled` → no failover;
- Transport field assertions (dial/handshake timeouts, H2 preservation);
- real TCP-blackhole listeners verifying a hung SOCKS5 / HTTP proxy fails within the `ctx` deadline.

`go build ./...`, `go test -tags=unit ./...`, and the targeted transport tests all pass on top of current `main`.

---

## Note on unrelated CI reds (not caused by this PR)

When CI runs, two checks may go red for reasons pre-existing on `main`, independent of this change — happy to adjust however you prefer:

- **`backend-security` (govulncheck):** `main` pins Go **1.26.4**, which govulncheck now flags for **GO-2026-5856** (`crypto/tls`, stdlib, fixed in **go1.26.5**). This affects `main` as-is; the fix is bumping the `go` directive + the CI `go1.26.x` version pins. I kept it out of this PR to stay focused, but can include it if you'd like.
- **`golangci-lint` (staticcheck):** adding this PR's tests to the very large `service` package can tip staticcheck's fact-loading on CI (linux) so it loses the `testing.T.Fatal` noreturn fact and emits nondeterministic **SA5011** false positives on the `if x == nil { t.Fatal }; x.Field` test idiom (local runs report 0 issues). If it surfaces, the clean fix is a scoped `.golangci.yml` exclusion (`SA5011` in `*_test.go`) rather than per-line `//nolint`. I can add that here if you want.


hongheshan-svg:

feiyuan201991-maker:
Production data point in support of this PR — we hit exactly this on `0.1.155` yesterday (2026-07-16) and it caused a ~1h user-visible outage.

### What happened

Attaching a **TLS fingerprint profile** (`extra.tls_fingerprint_profile_id`) to an Anthropic OAuth account makes every request through that account fail at the transport layer:

```
upstream request failed: Post "https://api.anthropic.com/v1/messages?beta=true":
net/http: HTTP/1.x transport connection broken: malformed HTTP response
"\x00\x00\x12\x04\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00d\x00\x04\x00\x01\x00\x00\x00\x05\x00\xff\xff\xff..."
```

`\x00\x00\x12\x04` is an HTTP/2 SETTINGS frame being read by the HTTP/1.x parser. The profile we used declares `alpn_protocols: ["http/1.1", "h2"]`; the server selects `h2`, but the fingerprint dialer's transport then speaks HTTP/1.1 over it.

### Why this PR matters (the failover half)

This is the part that turned a broken account into a **customer-facing outage**, and it's exactly what the PR description calls out:

> They neither retry nor fail over, **even when healthy accounts remain**.

The affected account was one of two active Anthropic accounts in the group. The other one was healthy the entire time and served requests fine when tested directly. But every request routed to the broken account returned a hard **502** to the client instead of failing over. From the client's side (Claude Code) it looked like a total upstream outage:

```
502 Upstream request failed · Retrying in 3s · attempt 3/10
```

All 10 client-side retries hit the same 502 — sticky session kept re-selecting the same broken account, and the transport error never triggered failover.

### Isolation steps (each ruled out, in this order)

| Changed | Result |
|---|---|
| Proxy protocol `http` → `socks5` (port `33335` → `22228`) | identical error |
| Switched to a different upstream proxy / different exit IP | identical error |
| **Removed the proxy entirely** (`proxy_id = NULL`, direct egress) | **identical error** |
| **Removed `tls_fingerprint_profile_id`** | **fixed — 3/3 HTTP 200 immediately** |

The third row is the important one: the failure reproduces with **no proxy at all**, so this is the TLS-fingerprint transport path, not a proxy issue.

Worth noting `curl` through the exact same proxy negotiates `h2` cleanly and returns real certs — so a `curl` check does not reproduce it. Only the Go client path does.

### Contrast with OpenAI

OpenAI accounts on the same instance — including one bound to the **same proxy** as an affected Anthropic account — were unaffected throughout. That matches this PR's framing: `handleOpenAIUpstreamTransportError` already covers the OpenAI path, and the Anthropic paths are the gap.

### Suggestions

1. **This PR (transport-error failover on Anthropic paths)** would have contained the blast radius entirely — the healthy account would have taken over and no user would have seen a 502. Would be great to see it land.
2. Separately, the TLS-fingerprint transport looks like it needs either `ForceAttemptHTTP2` on the fingerprint dialer's transport, or `NextProtos: ["http/1.1"]` pinned so ALPN can't negotiate `h2` the transport won't speak. Happy to open a separate issue for that if useful — it's arguably the root cause, while this PR is the safety net.
3. Minor, unrelated: proxy credentials (`user:pass`) are logged in plaintext on the failure path, which makes logs unsafe to share for debugging. Happy to file separately.

Env: `weishaw/sub2api:0.1.155`, Docker, Postgres 18, Anthropic OAuth + setup-token accounts. Reproduced consistently; resolved by detaching the fingerprint profile.

hongheshan-svg:
你分析的比我透彻，你抓包了吗
feiyuan201991-maker ***@***.***>于2026年7月17日 周五03:26写道：

feiyuan201991-maker:
是的 吗的死了三个号

hongheshan-svg:
Closing this — upstream已经在 44003d7f6 (`fix(gateway): Anthropic/Bedrock 传输层错误转 failover + 持久故障临时摘除账号`, v0.1.184) 中独立实现了同一修复，覆盖面比本 PR 更完整。

对比下来 upstream 的实现是本 PR 的超集：

- 同样覆盖 Anthropic/Bedrock 五条转发路径，返回 `*UpstreamFailoverError(502)` 由 handler failover，service 不再直写 502
- 持久故障临时摘除账号 10 分钟、`context.Canceled` 不换号不摘号
- **额外**用 `errors.Is(err, context.DeadlineExceeded) && errors.Is(ctx.Err(), context.DeadlineExceeded)` 区分「客户端超时」与「上游超时」——本 PR 只判 `Canceled`，会把客户端超时也当成可 failover，这一点 upstream 的处理更正确
- 分类器改名为平台无关的 `classifyUpstreamTransportError`，OpenAI/Anthropic 两侧共用，比本 PR 复用 `classifyOpenAITransportError` 更干净

我们 fork 侧已删除自有实现、切到 upstream 的 `handleUpstreamTransportError`。感谢 @feeeei 的实现 🙏

PR #3969, opened 2026-07-10
fix(grok): preserve compatible reasoning effort

### LINKED ISSUE #3968: fix(grok): /v1/responses usage log misses flat reasoning_effort
## Problem

The Grok-specific `/v1/responses` forwarding path does not reuse the existing unified reasoning-effort parser. As a result, OpenAI-compatible clients that send the flat Chat Completions field:

```json
{
  "model": "grok-4.3",
  "reasoning_effort": "high"
}
```

can successfully invoke the model, but the resulting `usage_logs.reasoning_effort` is `NULL`/empty instead of `high`.

The nested Responses form works:

```json
{
  "reasoning": {
    "effort": "high"
  }
}
```

This is visible with clients such as OpenCode using `@ai-sdk/openai-compatible`: its provider option `reasoningEffort: "high"` is serialized as `reasoning_effort: "high"`.

## Expected behavior

The Grok Responses path should apply the same reasoning-effort compatibility rules as the regular OpenAI paths:

- `reasoning.effort`
- `reasoning_effort`
- supported model suffix fallback where applicable

Both request shapes should produce `usage_logs.reasoning_effort = high`.

## Root cause

Grok accounts are intentionally dispatched early to `forwardGrokResponses`:

```go
if account.Platform == PlatformGrok {
    return s.forwardGrokResponses(ctx, c, account, body, originalModel, reqStream, startTime)
}
```

At the end of `forwardGrokResponses`, the usage metadata is currently extracted only from the nested Responses field:

```go
ReasoningEffort: ptrStringOrNil(
    normalizeOpenAIReasoningEffort(
        gjson.GetBytes(patchedBody, "reasoning.effort").String(),
    ),
),
```

However, the existing unified parser already supports both request forms and model suffix fallback:

```go
extractOpenAIReasoningEffortFromBody(body, requestedModel)
```

The raw Grok `/v1/chat/completions` path already uses this unified parser, so the inconsistency is specific to the Grok `/v1/responses` path.

## History

- Unified reasoning-effort parsing was introduced before Grok support:
  - `53ee6383` (`feat(usage): add reasoning effort column`, 2026-02-03)
  - `61a2bf46` added the body-level fast parser, 2026-02-12
  - `57099a6a` fixed the raw Chat Completions path to preserve `reasoning_effort`, 2026-05-02
- Grok subscription forwarding was added later:
  - `39be1ec9` (`feat: add grok subscription support`, 2026-06-16)
  - PR #3310

The Grok implementation added a new local extraction expression instead of reusing the parser that already existed.

## Suggested fix

In `backend/internal/service/openai_gateway_grok.go`, parse once with the shared helper:

```go
reasoningEffort := extractOpenAIReasoningEffortFromBody(patchedBody, originalModel)
```

Then use:

```go
ReasoningEffort: reasoningEffort,
```

A regression test should cover a Grok Responses request containing:

```json
{"model":"grok-4.3","reasoning_effort":"high"}
```

and assert that `OpenAIForwardResult.ReasoningEffort` is non-nil and equals `high`.

## Scope

This does not propose removing the Grok-specific upstream dispatch. That dispatch is needed for xAI OAuth, base URL handling, quota headers, model mapping, and Grok-specific request sanitization. The change only makes common request metadata parsing consistent across platform adapters.

jinfeijie:
## Summary

- reuse the shared OpenAI reasoning-effort parser in the Grok Responses path
- recognize both nested `reasoning.effort` and flat OpenAI-compatible `reasoning_effort` inputs
- preserve the parsed value in Grok usage logs
- add parser and end-to-end Grok Responses regression coverage

## Problem

`forwardGrokResponses` previously read only `reasoning.effort` when building `OpenAIForwardResult`. OpenAI-compatible clients such as OpenCode serialize their `reasoningEffort` provider option as the flat `reasoning_effort` request field, so Grok requests succeeded but `usage_logs.reasoning_effort` remained empty.

The regular OpenAI and raw Chat Completions paths already use `extractOpenAIReasoningEffortFromBody`, which supports both request shapes and model suffix fallback.

## Fix

Use the same shared parser in the Grok Responses path:

```go
reasoningEffort := extractOpenAIReasoningEffortFromBody(patchedBody, originalModel)
```

The Grok-specific dispatch remains unchanged; this only aligns common request metadata parsing across platform adapters.

Closes #3968

## Validation

- `gofmt` applied
- `git diff --check` passed
- focused Go tests could not complete locally because the repository requires Go 1.26.5 and automatic toolchain download repeatedly timed out; the added unit and forwarding regression tests are included for CI

### DIFF
--- a/backend/internal/service/openai_gateway_grok.go
+++ b/backend/internal/service/openai_gateway_grok.go
@@ -122,13 +122,14 @@ func (s *OpenAIGatewayService) forwardGrokResponses(
 	if usage == nil {
 		usage = &OpenAIUsage{}
 	}
+	reasoningEffort := extractOpenAIReasoningEffortFromBody(patchedBody, originalModel)
 	return &OpenAIForwardResult{
 		RequestID:       firstNonEmpty(resp.Header.Get("x-request-id"), resp.Header.Get("xai-request-id")),
 		ResponseID:      responseID,
 		Usage:           *usage,
 		Model:           originalModel,
 		UpstreamModel:   upstreamModel,
-		ReasoningEffort: ptrStringOrNil(normalizeOpenAIReasoningEffort(gjson.GetBytes(patchedBody, "reasoning.effort").String())),
+		ReasoningEffort: reasoningEffort,
 		Stream:          reqStream,
 		OpenAIWSMode:    false,
 		ResponseHeaders: resp.Header.Clone(),
--- a/backend/internal/service/openai_gateway_grok.go
+++ b/backend/internal/service/openai_gateway_grok.go
@@ -694,10 +694,3 @@ func (s *OpenAIGatewayService) tempUnscheduleGrok(ctx context.Context, account *
 		_ = s.accountRepo.SetTempUnschedulable(stateCtx, account.ID, until, reason)
 	}
 }
-
-func ptrStringOrNil(value string) *string {
-	if strings.TrimSpace(value) == "" {
-		return nil
-	}
-	return &value
-}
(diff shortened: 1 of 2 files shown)


SwallOwDili:

PR #3975, opened 2026-07-10
feat(openai): add configurable Fable family mapping

ncwn:
## Summary

- classify Claude Fable requests as a first-class messages-dispatch family
- add an optional `fable_mapped_model` group setting before Opus, Sonnet, and Haiku
- keep exact-model mappings as higher-priority overrides
- leave Fable unmapped when the family setting is blank

## Problem

Fable was missing from the shared Claude family classifier. A group-level exact mapping could resolve a target, but downstream OpenAI forwarding only applied the messages-dispatch target to a recognized Claude family. The original Fable model ID could therefore reach a Codex account and be rejected as unsupported.

A family setting is more reliable for versioned Fable model IDs while retaining exact mappings for genuine exceptions. The target model remains deployment configuration; this change does not add a built-in Fable target.

## Tests

- `make test-unit`
- `make test-integration`
- `make test-frontend`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.9.0 run ./...`


ncwn:

PR #3986, opened 2026-07-10
fix(pricing): support above-272k token rates

pkeran:
## Summary
- derive the existing long-context pricing policy from the remote model-price repository's `*_above_272k_tokens` fields
- preserve explicit `long_context_*` values when both formats are present
- cover parsing through actual GPT-5.6 Terra input, output, cache-read, and cache-write billing

## Why

The bundled fallback price file has `long_context_*` fields, but the runtime default source at `Wei-Shaw/model-price-repo` publishes GPT-5.6 long-context prices as `input_cost_per_token_above_272k_tokens` and `output_cost_per_token_above_272k_tokens`. Those fields were ignored after a successful remote sync, so requests over 272K context were billed at short-context rates.

## Test

`go test -tags unit ./internal/service -count=1`


pkeran:

PR #4040, opened 2026-07-11
fix(billing): apply service tier multipliers to channel pricing

AixLau:
## Summary
- treat flat and interval channel prices as Standard tier prices only
- apply existing priority 2x and flex 0.5x multipliers to final channel pricing
- add regression coverage for exact channel component prices and tier ratios

## Test plan
- `GOMAXPROCS=2 go test -p=1 -tags=unit ./internal/service -count=1`
- `GOMAXPROCS=2 go test -p=1 ./internal/service -run "^$"`


AixLau:

PR #4065, opened 2026-07-11
fix: retry OpenAI model sync after TLS EOF

GitClound:
## Summary
- Retry OpenAI API-key model-list sync with a Chrome TLS fingerprint when the standard upstream request ends in TLS EOF.
- Retry the Chrome request once for transient EOF or connection-reset errors.
- Keep the fallback scoped to OpenAI API-key accounts; other platforms and account types keep their current behavior.

## Validation
- `go vet ./internal/service`
- Live integration test with the affected account and an injected EOF completed successfully.

The live test is behind the `integration` build tag and requires `SUB2API_LIVE_OPENAI_BASE_URL` and `SUB2API_LIVE_OPENAI_API_KEY`; credentials are not included in this change.


GitClound:

The-Ladder-of-Rrogress:
Hi maintainers, just a friendly nudge on this PR from a self-hoster running weishaw/sub2api:latest.

**Why it matters for us:** We operate OpenAI-compatible API-key upstreams (e.g. https://api.b.ai, an OpenAI-format aggregator) where the standard Go http.Transport model-list sync gets rejected at the TLS layer — upstream returns 401 "Invalid token" only to the Go client while plain curl (including `Go-http-client/2.0` UA and HTTP/2) succeeds. The Chrome TLS-fingerprint fallback in this PR is exactly the fix we need for OpenAI API-key accounts (not just Anthropic OAuth), and currently `IsTLSFingerprintEnabled()` hard-gates TLS fingerprinting to Anthropic OAuth/SetupToken only.

**Status request:** The PR currently shows merge conflicts and 0 reviews (last updated 2026-07-17). Could you let us know:
1. Is this PR still planned to be merged?
2. Would you accept help resolving the merge conflicts / splitting out the large unrelated upstreamstation changes from the TLS-fingerprint core fix (the ~60-line `doUpstreamModelsRequest` + `req/v3` ImpersonateChrome fallback)?

Happy to submit a rebased, minimal PR scoped to just the TLS fix if that would help it land faster. Thanks!

PR #4066, opened 2026-07-11
feat(grok): support API Key + custom base_url upstream

Go1c:
## Summary

- Add Grok `type=apikey` so admins can attach OpenAI-compatible upstreams with **Base URL + API Key** (official `api.x.ai` or third-party hosts).
- OAuth path still uses the xAI host allowlist; apikey path uses OpenAI-compatible URL assembly for `/v1/responses`, `/v1/chat/completions`, and media endpoints.
- Frontend Create/Edit account modals: Grok dual mode OAuth | API Key, with i18n.

## Context

Upstream/majority forks only support Grok OAuth. We adopted a lighter approach than `youxuanxue/sub2api` (no edge/TokenKey stack): native `AccountTypeAPIKey` + existing OpenAI-compatible gateway forwarding.

## Test plan

- [x] `go test ./internal/service/ -run 'Grok|APIKey|...'`
- [x] Frontend `vue-tsc --noEmit`
- [x] Third-party OpenAI-compatible upstream smoke focused on **`grok-4.5`**:
  - models list includes `grok-4.5`
  - `/v1/responses` non-stream + stream → 200
  - `/v1/chat/completions` non-stream + stream → 200
  - bare `model=grok` may 503 on some hosts; gateway alias maps to `grok-4.5`
- [ ] Create Grok API Key account in admin UI and call through local gateway (optional follow-up)

## Notes

- Do not commit production API keys. Knowledge record: `.spec/knowledge/records/grok-apikey-upstream-20260711.md`.
- Base branch: `dev` (topic PR).


Go1c:
Opened against upstream by mistake; this change targets Go1c/sub2api base=dev.

PR #4071, opened 2026-07-11
fix(grok): strip unsupported reasoning parameters for Composer

JD13703202:
## Summary

- remove unsupported reasoning parameters from Grok Composer Responses requests
- preserve reasoning configuration for Grok models that support it
- add regression coverage for Composer aliases and provider-prefixed model names

## Problem

The Anthropic Messages to Responses conversion adds a default `reasoning.effort` value. `grok-composer-2.5-fast` does not support this parameter and rejects the request:

```text
400 invalid-argument:
Model grok-composer-2.5-fast does not support parameter reasoningEffort.
```

This affects `/v1/messages` requests converted to `/v1/responses`, including Claude Code Composer side requests.

## Fix

Before forwarding a Grok Responses request, remove:

- `reasoning`
- `reasoning_effort`
- `reasoningEffort`

when the effective upstream model is `grok-composer-2.5-fast`, `composer-2.5`, or a provider-prefixed equivalent.

Other Grok models retain their reasoning configuration.

## Validation

- added regression coverage for the canonical model, alias, and provider-prefixed form
- added regression coverage confirming `grok-4.5` preserves reasoning parameters
- `git diff --check` passed
- focused Go tests were not run locally because the installed Snap Go toolchain lacks the required `cap_dac_override` capability

### DIFF
--- a/backend/internal/service/openai_gateway_grok.go
+++ b/backend/internal/service/openai_gateway_grok.go
@@ -146,6 +146,10 @@ func patchGrokResponsesBody(body []byte, upstreamModel string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	out, err = sanitizeGrokResponsesModelCapabilities(out, upstreamModel)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


JD13703202:

PR #4085, opened 2026-07-12
[codex] add FIFO account wait queue

xtyooo:
## What changed

- Add an optional Redis-backed FIFO wait queue for account slots
- Only the queue head can acquire the next account slot
- Encode wait deadlines in tickets and remove expired or cancelled heads
- Keep the legacy counter and polling path for non-queue cache implementations
- Use the configured WaitPlan max waiting and timeout across Claude, Gemini, and OpenAI account paths
- Include integration coverage for FIFO ordering, capacity limits, and cancelled waiters

## Why

The previous account wait path only counted waiters and let every request compete to acquire a slot. That allowed later requests to overtake earlier requests and caused avoidable slot-acquisition pressure during bursts.

This first slice makes admission order deterministic and bounds stale waiters. Waiters still use the existing exponential-backoff polling loop, but only the FIFO head performs the account-slot acquisition attempt; event-driven wakeups can be added in a follow-up without changing the queue contract.

## Validation

- `go test ./internal/handler ./internal/service ./internal/repository -count=1`
- `go test -tags integration ./internal/repository -run TestConcurrencyCacheSuite/TestAccountWaitQueue -count=1`
- `git diff --check`

The queue is optional so existing cache implementations and test doubles retain the legacy behavior.

### DIFF
--- a/backend/internal/handler/gateway_handler.go
+++ b/backend/internal/handler/gateway_handler.go
@@ -391,10 +391,11 @@ func (h *GatewayHandler) Messages(c *gin.Context) {
 					}
 				}
 
-				accountReleaseFunc, err = h.concurrencyHelper.AcquireAccountSlotWithWaitTimeout(
+				accountReleaseFunc, err = h.concurrencyHelper.AcquireAccountSlotWithWaitTimeoutQueued(
 					c,
 					account.ID,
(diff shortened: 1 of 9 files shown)


xtyooo:

PR #4095, opened 2026-07-12
fix(openai): support variable quota window usage stats

emailck:
## Summary

- preserve the upstream Codex window duration in account usage responses
- ignore explicit zero-minute placeholder windows instead of rendering a synthetic 5h window
- aggregate local request, token, and cost stats using the actual upstream window duration
- derive OpenAI quota labels from the real duration, including 30d windows, while retaining legacy fallbacks

## Tests

- `go test ./internal/service -run TestBuildCodexUsageProgressFromExtra -count=1`
- `pnpm test:run -- src/components/account/__tests__/AccountUsageCell.spec.ts`
- `NODE_OPTIONS=--max-old-space-size=1536 pnpm build`

## Compatibility

Accounts without `window_minutes` retain the existing 5h/7d fallback behavior.

### DIFF
--- a/backend/internal/service/account_usage_service.go
+++ b/backend/internal/service/account_usage_service.go
@@ -144,10 +144,11 @@ type WindowStats struct {
 
 // UsageProgress 使用量进度
 type UsageProgress struct {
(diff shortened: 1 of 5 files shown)


emailck:

emailck:
recheck

PR #4111, opened 2026-07-12
feat: reduce upstream quota consumption

sparkjokerben:
## What changed

- add group and API-key controls for quota-first sticky scheduling, session model stability, and unified retry budgets
- keep session-to-model decisions in a separate versioned Redis binding with availability-preserving fallback
- share retry limits across account failover, HTTP retries, WebSocket reconnects, and request rectification retries
- detect retry-capable clients through an explicit header plus versioned two-factor client rules
- add signed rule updates, administrator overrides, rollback, ETag caching, and management UI
- coalesce capability and usage probes in-process and across instances with Redis leases

## Why

Layered retries, account switching, model remapping, and duplicate probes can increase upstream quota consumption and reduce prompt-cache reuse. These changes reduce duplicate submissions while preserving the existing upstream request profile.

All behavior-changing group policies default to disabled. Existing OAuth, TLS/proxy, User-Agent, originator, session identifiers, payload transforms, and streaming protocol behavior remain unchanged.

## User and operator impact

- administrators can set group defaults and control whether users may override sticky routing per API key
- users can choose inherit/enabled/disabled when the group permits overrides
- administrators can inspect and override client retry rules and manage signed updates
- remote rule updates require an Ed25519 public key supplied through `CLIENT_RETRY_RULES_PUBLIC_KEY` or build-time injection; failures safely retain the last-known-good or embedded rules

## Validation

- `go test ./... -run '^$'`
- targeted retry-budget, sticky-policy, model-stability, and rule-manifest tests
- migration and repository tests
- `vue-tsc --noEmit`
- ESLint on changed frontend files
- `git diff --check`


sparkjokerben:
Closing this draft; I’ll maintain the change in my fork instead.

PR #4139, opened 2026-07-13
fix(grok): support upstream model sync for API keys

heathermhuang:
## Summary

- support live upstream model synchronization for Grok API-key accounts
- request the OpenAI-compatible `/models` endpoint using the configured base URL and bearer API key
- keep Grok OAuth model synchronization explicitly unsupported because it uses a different subscription transport
- add regression coverage for request construction, response parsing, authentication, deduplication, and OAuth rejection

## Root cause

The model whitelist UI enabled upstream synchronization for Grok, but the backend request dispatcher did not route Grok accounts. The request therefore returned HTTP 400 before contacting the configured upstream.

## Validation

- `go test ./internal/service ./internal/handler/admin -count=1`
- `go vet ./internal/service ./internal/handler/admin`
- `git diff --check`

## Test plan

- [x] Grok API-key accounts build the configured OpenAI-compatible model-list URL
- [x] Bearer authentication is applied without exposing credentials
- [x] duplicate model IDs are normalized
- [x] Grok OAuth accounts return an account-type-specific unsupported error

### DIFF
--- a/backend/internal/service/upstream_models.go
+++ b/backend/internal/service/upstream_models.go
@@ -131,6 +131,8 @@ func (s *AccountTestService) buildUpstreamModelsRequest(ctx context.Context, acc
 	switch {
 	case account.Platform == PlatformAntigravity:
 		return s.buildAntigravityAPIKeyModelsRequest(ctx, account)
+	case account.IsGrok():
+		return s.buildGrokUpstreamModelsRequest(ctx, account)
 	case account.IsOpenAI():
 		return s.buildOpenAIUpstreamModelsRequest(ctx, account)
 	case account.IsGemini():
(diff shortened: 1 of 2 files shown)


heathermhuang:
@Wei-Shaw This is ready for review.

What changed:
- Route Grok API-key upstream model sync through the OpenAI-compatible `/models` flow.
- Keep Grok OAuth accounts explicitly unsupported for this operation.
- Add regression coverage for bearer authentication, custom base URL handling, model filtering/deduplication, and safe error behavior using `grok-4.5`.

Validation:
- Focused Go tests and `go vet` pass.
- All hosted test, lint, frontend, CLA, and security checks are green.
- Private live canary passed with CLIProxyAPI v7.2.71: `grok-4.5` appeared in model discovery and a Grok 4.5 chat completion returned HTTP 200. Both services remained healthy with zero restarts.

Could you please review when convenient?

heiyuan0801:
@Wei-Shaw 作者大大，这个可以修复合并一下

PR #4165, opened 2026-07-13
fix(apicompat): preserve max effort for GPT-5.6 Anthropic bridge

jeffry88:
## Summary

Anthropic `/v1/messages` requests can specify `output_config.effort=max`.
GPT-5.6 Responses models support `max` as a distinct reasoning tier, but the
current Anthropic bridge always converts it to `xhigh`.

This change preserves `max` for GPT-5.6 models while keeping the existing
`max → xhigh` fallback for earlier models.

## Behavior

- GPT-5.6 variants: `max → max`
- Earlier models such as GPT-5.2 and GPT-5.4: `max → xhigh`
- Other effort levels remain unchanged

## Tests

- Added coverage for bare GPT-5.6 and sol/terra/luna variants
- Added provider-prefixed model coverage
- Preserved legacy GPT-5.2/GPT-5.4 fallback coverage
- Added a GPT-5.60 boundary case to prevent false-positive matching

`go test -tags=unit ./internal/pkg/apicompat -count=1`

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_responses_test.go
+++ b/backend/internal/pkg/apicompat/anthropic_responses_test.go
@@ -1124,19 +1124,37 @@ func TestAnthropicToResponses_OutputConfigHigh(t *testing.T) {
 }
 
 func TestAnthropicToResponses_OutputConfigMax(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jeffry88:

PR #4168, opened 2026-07-13
feat(api-key): add per-key concurrency limits

adamglin0:
## Summary

- add a configurable `concurrency_limit` to individual API keys
- enforce key-level concurrency atomically with the existing Redis slot infrastructure across HTTP, Gemini, and Responses WebSocket turns
- expose the setting in the API key edit dialog with Chinese and English copy

## Compatibility

- use `0` as unlimited and default existing rows to `0` in migration 175
- bump the API key auth snapshot version so cached authentication data reloads the new field
- treat missing frontend values as `0` during rolling upgrades

## Validation

- `go test ./...`
- `go test -tags=unit ./internal/service ./internal/handler`
- `pnpm test:run` (956 tests)
- `pnpm build`
- targeted frontend ESLint checks


adamglin0:
#4312

PR #4173, opened 2026-07-13
fix(grok): keep OAuth media on the official API

### LINKED ISSUE #4170: Grok 0.1.152 版本图生视频失败
Grok 0.1.152 版本图生视频失败,回滚到0.1.151正常
```
{
  "error": {
    "message": "xAI upstream returned status 400",
  }
}

{
  "error": "Failed to read request body: length limit exceeded"
}
```
@heathermhuang  佬可以看看

heathermhuang:
## Summary

- split Grok upstream selection by traffic contract: OAuth text/Responses traffic continues to use the CLI subscription proxy, while Grok Imagine media traffic uses the official `api.x.ai` API
- preserve API-key custom base URLs and explicit unsafe development overrides
- recognize normalized official API and CLI base URL variants before selecting the media upstream
- add a route-level regression test with a 2 MiB base64 image payload

## Root cause

The OAuth routing change introduced for text compatibility made `GetGrokBaseURL()` return the CLI proxy for every Grok request. `ForwardGrokMedia()` reused that generic selector, so image-to-video payloads were sent to the CLI proxy as well. Its smaller request-body limit rejected large base64 media inputs with HTTP 400 before xAI could create the video request.

This change makes the distinction structural: text and media call separate base URL selectors.

## Regression coverage

The new matrix verifies:

- OAuth text still routes to the CLI proxy
- OAuth image/video routes to the official media API
- a 2 MiB image-to-video body is preserved end-to-end
- stored official/CLI URL variants cannot accidentally change the media route
- API-key custom bases keep their existing behavior
- untrusted OAuth custom bases remain pinned to a first-party xAI host
- explicit unsafe development overrides remain available behind the existing environment flag

## Validation

- `make -C backend test-unit`
- `make -C backend test-integration`
- `go vet ./internal/pkg/xai ./internal/service`
- `git diff --check`

## Live verification

- deployed as an app-only hotfix on the existing v0.1.152 production lineage; the deployed candidate differs from this PR head only in base ancestry
- public `https://api.byok.chat/health` returned `200 {"status":"ok"}` after cutover
- sent a real Grok OAuth image-to-video request with a 1,609,937-byte JSON body containing a 1,207,278-byte PNG as base64
- `POST /v1/videos/generations` returned HTTP 200 and a request ID; polling reached `status=done`
- completed response contained a video URL with 8-second duration at 480p
- the production usage row confirmed `platform=grok`, `account_type=oauth`, `billing_mode=video`, and the expected requested model
- production container remained healthy with zero restarts after the canary

Fixes #4170

### DIFF
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -1251,6 +1251,9 @@ func (a *Account) GetOpenAIRefreshToken() string {
 	return a.GetCredential("refresh_token")
 }
 
+// GetGrokBaseURL selects the upstream used by Grok text and Responses traffic.
+// Grok media traffic has a different transport contract and must use
+// GetGrokMediaBaseURL instead.
 func (a *Account) GetGrokBaseURL() string {
 	if !a.IsGrok() {
 		return ""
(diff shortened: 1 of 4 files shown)


put-go:
@Wei-Shaw 佬,这个问题直接影响Grok图生视频,可以优先安排下PR 😊

PR #4176, opened 2026-07-13
fix(openai): pass deterministic upstream 4xx through instead of 502

bestony:
## Problem

On the OpenAI Responses path, `handleErrorResponse` and `mapUpstreamError` flatten every unrecognized upstream status — including deterministic request-shaped 4xx — into a generic client **502 "Upstream request failed"**:

- invalid `response_format` schemas (400)
- context-window exceeded (400)
- plan-gated models on ChatGPT accounts (400)
- unknown `previous_response_id` / model (404)

Clients such as Codex CLI auto-retry 5xx with backoff, so one deterministic 400 becomes a **sustained retry storm** (observed in production: ~2200 app-generated 502s in 3h from a handful of codex-tui users), and the actionable upstream message is lost — users never learn why their request failed.

The Chat Completions compat path (`handleCompatErrorResponse`) already passes upstream status + message through; this PR aligns the Responses path with that in-repo precedent.

## Fix

New shared helpers in `internal/service/openai_upstream_request_errors.go`:

- `IsOpenAIUpstreamRequestError`: request-shaped statuses = **{400, 404, 413, 422}**. 409 is deliberately excluded (OpenAI uses it for time-dependent conflicts where retry-later can succeed).
- `ShouldPassthroughOpenAIUpstreamRequestError`: two classes stay on the opaque, retryable 502 even though they ride a request-shaped status:
  - **transient processing failures** (`server_is_overloaded`, `slow_down`, "selected model is at capacity", "an error occurred while processing your request" — reusing the existing `isOpenAITransientProcessingError` classifier, i.e. exactly the errors the failover path already treats as retryable);
  - **account-shaped 400s** ("organization has been disabled", "credit balance", "identity verification is required" — the same markers `RateLimitService.HandleUpstreamError` disables accounts on), which describe the gateway's account, not the client's request, and must not leak.
- `OpenAIUpstreamRequestErrorType`: `invalid_request_error` / `not_found_error`, mirroring the compat path.
- `OpenAIUpstreamRequestErrorMessage`: sanitized upstream message with generic fallback.

Wired into:
- `handleErrorResponse` final mapping (non-failover upstream errors),
- `OpenAIGatewayHandler.mapUpstreamError` + `handleFailoverExhausted` / `handleAnthropicFailoverExhausted` (failover-exhausted path; message enriched from `UpstreamFailoverError.ResponseBody`),
- `mapResponsesErrorCode`: new `not_found_error` → `not_found` so streaming 404s emit a canonical code in `response.failed` SSE frames.

**Unchanged**: 401/402/403 stay mapped to 502 (gateway-account credential problems must not look like client auth errors), 429/529/5xx mappings untouched, error passthrough rules still take precedence, `GatewayHandler.mapUpstreamError` (Gemini/antigravity flows) untouched.

## Tests

- Helper table tests (status list, error types, message extraction/sanitization, transient + account-shaped guards).
- `handleErrorResponse`: request-shaped 4xx passthrough (422 schema / 400 plan-gated / 404 / empty body fallback), transient-400 stays 502, default mapping for unrecognized 5xx preserved.
- Handler: `mapUpstreamError` table, `handleFailoverExhausted` 400-passthrough + transient-400 + account-shaped-400 + 401 regression, Anthropic-format variant.
- Updated `TestOpenAIGatewayService_Forward_LogsInstructionsRequiredDetails` (asserted the old 502 flattening for an upstream 400) to the new passthrough behavior.
- `go test -tags=unit ./...` passes; `golangci-lint run --new-from-rev=upstream/main` reports 0 new issues.

## Out of scope (deliberate)

- 410/415 remain on the 502 mapping (conservative status list).
- Context-window errors arriving on a 5xx carrier status keep today's behavior (502 + real message).
- The upstream message exposure matches what the Chat Completions compat path already exposes today.

Related: #4169 (plan-gated model cooldown + failover) — complementary: #4169 stops selecting incapable accounts; this PR stops misreporting whatever deterministic 4xx still reaches the client. With both merged, a plan-gated request that exhausts failover now returns an honest 400 with the upstream explanation instead of a retryable 502.

### DIFF
--- a/backend/internal/handler/openai_gateway_handler.go
+++ b/backend/internal/handler/openai_gateway_handler.go
@@ -1094,6 +1094,16 @@ func (h *OpenAIGatewayHandler) anthropicStreamingAwareError(c *gin.Context, stat
 // handleAnthropicFailoverExhausted maps upstream failover errors to Anthropic format.
 func (h *OpenAIGatewayHandler) handleAnthropicFailoverExhausted(c *gin.Context, failoverErr *service.UpstreamFailoverError, streamStarted bool) {
 	status, errType, errMsg := h.mapUpstreamError(failoverErr.StatusCode)
+	if service.IsOpenAIUpstreamRequestError(failoverErr.StatusCode) {
+		if service.ShouldPassthroughOpenAIUpstreamRequestError(failoverErr.StatusCode, failoverErr.ResponseBody) {
+			errMsg = service.OpenAIUpstreamRequestErrorMessage(failoverErr.ResponseBody)
+		} else {
+			// Transient overloads and account-shaped 400s ride request-shaped
+			// statuses but are not the client's fault; keep the retryable,
+			// opaque mapping.
+			status, errType, errMsg = http.StatusBadGateway, "upstream_error", "Upstream service temporarily unavailable"
+		}
+	}
 	h.anthropicStreamingAwareError(c, status, errType, errMsg, streamStarted)
 }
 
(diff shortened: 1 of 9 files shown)


bestony:
图片
you can use this feature to make all error show to user.
图片

PR #4179, opened 2026-07-13
fix(openai): preserve exact messages dispatch mappings

### LINKED ISSUE #4177: bug: exact /v1/messages model mappings are dropped for unknown Claude families
## Summary

An OpenAI group can resolve an exact `/v1/messages` model mapping, but forwarding drops the selected model when the requested Claude model is not classified as Opus, Sonnet, or Haiku.

For example, an exact mapping from `claude-fable-5` to `gpt-5.6-sol` is resolved correctly by the group, but the original `claude-fable-5` is ultimately sent upstream.

## Request path

1. `Group.ResolveMessagesDispatchModel` resolves the exact mapping to `gpt-5.6-sol`.
2. Forwarding calls `resolveOpenAIForwardModel`.
3. `resolveOpenAIForwardModel` only accepts `defaultMappedModel` when `claudeMessagesDispatchFamily(requestedModel)` is non-empty.
4. `claudeMessagesDispatchFamily` only recognizes Opus, Sonnet, and Haiku, so `claude-fable-5` falls back to the original model.

## Expected behavior

A non-empty model already selected by eligible `/v1/messages` dispatch should be used when account-level model mapping does not match, without reclassifying the original Claude family. Account exact and wildcard mappings should retain higher priority.

## Compatibility boundaries

- Ordinary `/v1/chat/completions` and other OpenAI request paths must continue to pass an empty messages-dispatch model and must not use the group default.
- Non-OpenAI groups must not gain OpenAI messages dispatch behavior.
- Existing Opus, Sonnet, and Haiku family defaults should remain unchanged.

## Reproduction

Configure an OpenAI group with:

```json
{
  "exact_model_mappings": {
    "claude-fable-5": "gpt-5.6-sol"
  }
}
```

Send a `/v1/messages` request using `claude-fable-5` through an account without a matching account-level model mapping. The upstream request currently contains `claude-fable-5`; it should contain `gpt-5.6-sol`.

Observed on current upstream `main` at `adb5106c1`.

zz80900:
## Summary

- treat a non-empty `/v1/messages` dispatch model as a caller-approved fallback after account-level mapping
- stop reclassifying exact mappings through the Opus/Sonnet/Haiku-only family helper
- preserve account exact, wildcard, and pass-through mapping precedence
- keep ordinary GPT and `/v1/chat/completions` requests outside messages dispatch by passing an empty dispatch model

## Root cause

`Group.ResolveMessagesDispatchModel` correctly resolved exact mappings such as `claude-fable-5 -> gpt-5.6-sol`, but `resolveOpenAIForwardModel` rejected the result because `claudeMessagesDispatchFamily` does not recognize Fable. The original Claude model was then forwarded upstream.

## Tests

- added Fable exact mapping coverage beside the existing Sonnet case
- added account exact, wildcard, pass-through, nil-account, and whitespace fallback cases
- added a full `/v1/messages` forwarding regression asserting the upstream body uses `gpt-5.6-sol`
- retained `/v1/chat/completions` and non-OpenAI group boundary coverage

```text
go test ./internal/service ./internal/handler -count=1
ok  github.com/Wei-Shaw/sub2api/internal/service  57.156s
ok  github.com/Wei-Shaw/sub2api/internal/handler  29.384s
```

Fixes #4177

### DIFF
--- a/backend/internal/handler/openai_gateway_handler_test.go
+++ b/backend/internal/handler/openai_gateway_handler_test.go
@@ -415,11 +415,13 @@ func TestResolveOpenAIMessagesDispatchMappedModel(t *testing.T) {
 					SonnetMappedModel: "gpt-5.2",
 					ExactModelMappings: map[string]string{
 						"claude-sonnet-4-5-20250929": "gpt-5.4-mini-high",
+						"claude-fable-5":             "gpt-5.6-sol",
 					},
 				},
 			},
 		}
 		require.Equal(t, "gpt-5.4-mini", resolveOpenAIMessagesDispatchMappedModel(apiKey, "claude-sonnet-4-5-20250929"))
+		require.Equal(t, "gpt-5.6-sol", resolveOpenAIMessagesDispatchMappedModel(apiKey, "claude-fable-5"))
 	})
 
 	t.Run("uses_family_default_when_no_override", func(t *testing.T) {
--- a/backend/internal/service/openai_compat_model_test.go
+++ b/backend/internal/service/openai_compat_model_test.go
@@ -124,6 +124,55 @@ func TestApplyOpenAICompatModelNormalization(t *testing.T) {
 	})
 }
 
+func TestForwardAsAnthropic_UsesExactFableMessagesDispatchModel(t *testing.T) {
+	t.Parallel()
+	gin.SetMode(gin.TestMode)
+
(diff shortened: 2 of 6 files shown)


zz80900:
recheck

zz80900:
recheck

zz80900:
recheck

zz80900:

PR #4182, opened 2026-07-13
fix: fail over grok token refresh failures

dishenglee:
## Summary\n- Convert Grok access-token refresh failures into failover errors so the gateway can switch accounts instead of returning an immediate upstream failure.\n- Apply the behavior to Grok Responses forwarding and raw Chat Completions forwarding.\n\n## Test\n- go test -tags=unit ./internal/service/ -run 'TestForwardAsChatCompletionsForGrok|TestBuildGrokResponsesRequest|TestGetGrokCLIHeaders'

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,3 +747,7 @@ Copyright (c) 2026 Wesley Liddick
 **If you find this project useful, please give it a star!**
 
 </div>
+
+---
+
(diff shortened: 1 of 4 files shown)


dishenglee:
Closing: changes were intended for dishenglee/sub2api-grok-cli, not upstream.

PR #4218, opened 2026-07-13
fix(openai): complete streamed image generation results

### LINKED ISSUE #4051: [Bug] Codex Desktop receives image result with status=generating and does not render it
## Summary

When using an OpenAI OAuth account through the Codex `/v1/responses` path, image generation succeeds and the complete PNG result is returned, but the emitted `image_generation_call` remains `status: "generating"`.

Codex Desktop therefore treats the call as unfinished and does not render the generated image.

## Environment

- Sub2API: `v0.1.151`
- Commit: `deff3123ded1d14e51df1fd1286e3d43ed9ec9bd`
- Codex Desktop: `0.144.0-alpha.4`
- Account type: OpenAI OAuth
- Endpoint: `POST /v1/responses`
- Requested model: `gpt-image-2`
- Upstream model after normalization: `gpt-5.4-mini`
- Request mode: streaming

## Reproduction

1. Configure an OpenAI OAuth account and allow image generation for the group.
2. Configure Codex Desktop to use the Sub2API `/v1/responses` endpoint.
3. Ask Codex Desktop to generate an image with `gpt-image-2`.
4. Wait until the request finishes successfully.
5. Observe that no final image is rendered in Codex Desktop.

## Observed behavior

The gateway reports successful image generation:

- HTTP status is `200`.
- The response transferred to Codex Desktop is approximately 11-12 MB.
- The usage record contains `image_count = 1` and a valid output size.
- There is no reverse-proxy timeout or truncated response on these requests.

The local Codex rollout contains an item equivalent to:

```json
{
  "type": "image_generation_call",
  "id": "ig_REDACTED",
  "status": "generating",
  "revised_prompt": "...",
  "result": "iVBORw0KGgo..."
}
```

The `result` is a valid PNG Base64 payload (approximately 3.4 MB), but the status is still `generating`.

## Expected behavior

Once `response.output_item.done` contains a non-empty image `result`, or the terminal `response.completed` event is received, the downstream item should be normalized to:

```json
{
  "type": "image_generation_call",
  "status": "completed",
  "result": "iVBORw0KGgo..."
}
```

Codex Desktop should then render the image as a completed image-generation call.

## Suspected cause

The Responses streaming path collects an `image_generation_call` from `response.output_item.done` when `result` is non-empty, but preserves the upstream item's stale `status: "generating"` value.

The terminal response normalization can also retain the stale status when `response.completed.response.output` is already non-empty.

## Suggested fix

Normalize image items at both boundaries:

1. When collecting `response.output_item.done`, if `type == "image_generation_call"` and `result` is non-empty, set `status = "completed"`.
2. When forwarding a terminal `response.completed`/`response.done` payload, normalize every `response.output[]` image item with a non-empty `result` to `status = "completed"`.

For example:

```go
if item.Get("type").String() == "image_generation_call" &&
    strings.TrimSpace(item.Get("result").String()) != "" {
    itemJSON, _ = sjson.SetBytes(itemJSON, "status", "completed")
}
```

A regression test should assert that an upstream `response.output_item.done` containing `status: "generating"` plus a non-empty `result` is emitted downstream as `status: "completed"`.

## Additional note

This is distinct from proxy timeout failures: the affected requests return `200` and deliver the complete image payload to Codex Desktop. The rendering failure is caused by the non-terminal item status.

chensunlai:
## Summary

- Normalize final `image_generation_call` items from `generating`/`in_progress` to `completed` only when a non-empty result is present.
- Cover regular Responses SSE, passthrough SSE, and SSE-to-JSON reconstruction.
- Preserve Base64 data, event types/order, failed items, result-less items, and non-terminal events.

## Scope

This is intentionally limited to the HTTP/SSE path reproduced in #4051; non-streaming JSON and WebSocket paths are unchanged.

## Tests

Fixes #4051

### DIFF
--- a/backend/internal/service/openai_gateway_passthrough.go
+++ b/backend/internal/service/openai_gateway_passthrough.go
@@ -914,6 +914,11 @@ func (s *OpenAIGatewayService) handleStreamingResponsePassthrough(
 				trimmedData = strings.TrimSpace(string(normalizedData))
 				line = "data: " + string(normalizedData)
 			}
+			if normalizedData, normalized := normalizeCompletedImageGenerationStatus(dataBytes); normalized {
+				dataBytes = normalizedData
+				trimmedData = strings.TrimSpace(string(normalizedData))
+				line = "data: " + string(normalizedData)
+			}
 			eventType := strings.TrimSpace(gjson.Get(trimmedData, "type").String())
 			if eventType == "response.failed" {
 				failedMessage = extractOpenAISSEErrorMessage(dataBytes)
--- a/backend/internal/service/openai_gateway_response_handling.go
+++ b/backend/internal/service/openai_gateway_response_handling.go
@@ -281,6 +281,11 @@ func (s *OpenAIGatewayService) handleStreamingResponse(ctx context.Context, resp
 				forceFlushFailedEvent = true
 				sawFailedEvent = true
 			}
+			if normalizedData, normalized := normalizeCompletedImageGenerationStatus(dataBytes); normalized {
+				dataBytes = normalizedData
+				data = string(normalizedData)
+				line = "data: " + data
+			}
 			imageCounter.AddSSEData(dataBytes)
 
 			// Correct Codex tool calls if needed (apply_patch -> edit, etc.)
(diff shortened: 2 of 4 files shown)


chensunlai:

PR #4219, opened 2026-07-13
fix(images): add opt-in non-stream JSON keepalive

zh239ns:
## Problem

Non-streaming OpenAI Images requests can remain completely silent while image generation runs. Deployments behind Nginx, OpenResty, Cloudflare, or another proxy with an idle/read timeout can lose the downstream connection before the image is ready.

The observed failure was deterministic:

- a simple image completed in about 25 seconds;
- more complex image requests consistently lost the client connection at about 121.8 seconds;
- Sub2API later logged `status=200`, `latency_ms=140423`, and `write: broken pipe`.

## Confirmed root cause

For OpenAI OAuth accounts, the public non-streaming `/v1/images/generations` request is internally converted to a streaming Responses request:

- `buildOpenAIImagesResponsesRequest` sets `stream: true`;
- the upstream request sends `Accept: text/event-stream`;
- `handleOpenAIImagesOAuthNonStreamingResponse` buffers the complete upstream SSE body before writing the final Images JSON response;
- `detachUpstreamContext` uses `context.WithoutCancel`, so generation continues after the downstream proxy disconnects.

The proxy therefore sees no downstream response bytes during generation, closes the apparently idle request, and Sub2API later writes the completed image to a closed socket.

API-key Images providers can have the same downstream-silence problem because many return only a final JSON response, so the keepalive is applied at the public non-streaming Images handler rather than only to OAuth accounts.

## Fix

Add an opt-in JSON keepalive for non-streaming Images requests:

- introduce `gateway.image_nonstream_keepalive_interval`, default `0` (disabled), valid range `5-60` seconds;
- start the keepalive only after request parsing, moderation, billing checks, account selection, and concurrency-slot acquisition have succeeded;
- write JSON-standard leading whitespace and flush it at the configured interval;
- send `Content-Type: application/json`, `Cache-Control: no-cache`, and `X-Accel-Buffering: no` before the first heartbeat;
- serialize heartbeat and terminal response writes with a wrapped Gin `ResponseWriter`;
- exclude heartbeat bytes from written-size checks so same-account retry and account failover remain available;
- preserve an OpenAI-compatible JSON error body if an error arrives after the first heartbeat;
- keep `stream=true` behavior unchanged.

## Why body whitespace instead of HTTP 102

#2047 attempted periodic `102 Processing` responses. Its production feedback showed that Cloudflare Tunnel could still terminate the request around 120 seconds because informational responses were not a reliable application-data heartbeat on that path.

This PR sends actual response-body bytes. JSON permits leading whitespace before the document, and the regression tests validate that the complete response remains valid JSON. This follows the production-tested approach in #2976 while additionally preserving retry/failover accounting through adjusted writer-size checks.

## Compatibility and behavior

| Scenario | Behavior |
| --- | --- |
| Default configuration | No goroutine, writer wrapper, status change, or response-body change |
| `stream=true` | Unchanged; existing SSE keepalive remains in use |
| Non-stream request finishes before first heartbeat | Original HTTP status and JSON body are preserved |
| Long successful request after heartbeat | HTTP 200 with valid JSON preceded by whitespace |
| Retryable upstream failure after heartbeat | Heartbeat bytes do not block same-account retry or account failover |
| Terminal failure after heartbeat | HTTP status is already committed as 200; body is a valid OpenAI-style JSON error |

The final row is an unavoidable HTTP constraint once body bytes have been sent. The feature is therefore disabled by default and must be explicitly enabled by operators who prefer connection survival over late HTTP status preservation.

## Configuration

```yaml
gateway:
  # 0 disables the feature; non-zero values must be between 5 and 60 seconds.
  image_nonstream_keepalive_interval: 15
```

Environment variable:

```text
GATEWAY_IMAGE_NONSTREAM_KEEPALIVE_INTERVAL=15
```

## Regression coverage

- default-disabled path is a complete no-op;
- environment-variable loading and configuration range validation;
- fast errors preserve their original HTTP status;
- delayed OAuth SSE-to-JSON success flushes heartbeat bytes and remains valid JSON;
- late errors remain valid OpenAI error JSON;
- heartbeat-only writes do not disable retry/failover detection;
- writer nil guards and idempotent stop behavior;
- race-detector coverage for concurrent heartbeat/final writes;
- existing OAuth/API-key Images, error, failover, and handler tests remain passing.

## Validation

```text
go test -tags=unit ./... -count=1
go test -race ./internal/config ./internal/service ./internal/handler -run 'Test(LoadImageNonstreamKeepaliveFromEnv|Load_DefaultGatewayImageStreamConfig|OpenAIImagesJSONKeepalive|OpenAIGatewayServiceForwardImages_OAuth|ImagesOAuthNonStreaming)' -count=1
go vet ./internal/config ./internal/service ./internal/handler
git diff --check
```

## Residual risk and rollout

- This addresses idle/read timeouts. It cannot bypass a proxy that enforces an absolute maximum request duration regardless of traffic.
- An intermediary that buffers all tiny chunks despite `X-Accel-Buffering: no` may still defeat the heartbeat. Operators should verify their actual proxy path.
- Recommended rollout: enable at `15` seconds for one API key or one Sub2API instance, verify long image requests and error reporting, then expand.
- Async jobs/polling would eliminate synchronous proxy lifetime limits entirely, but would require a new API contract and is intentionally outside this compatibility-focused change.

Refs #2031 #2047 #2976 #3906 #4019

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -818,6 +818,8 @@ type GatewayConfig struct {
 	ImageStreamDataIntervalTimeout int `mapstructure:"image_stream_data_interval_timeout"`
 	// ImageStreamKeepaliveInterval: 图片流式 keepalive 间隔（秒），0表示禁用
 	ImageStreamKeepaliveInterval int `mapstructure:"image_stream_keepalive_interval"`
+	// ImageNonstreamKeepaliveInterval: 图片非流式 JSON keepalive 间隔（秒），0表示禁用
+	ImageNonstreamKeepaliveInterval int `mapstructure:"image_nonstream_keepalive_interval"`
 	// MaxLineSize: 上游 SSE 单行最大字节数（0使用默认值）
 	MaxLineSize int `mapstructure:"max_line_size"`
 
@@ -2032,6 +2034,7 @@ func setDefaults() {
 	viper.SetDefault("gateway.stream_keepalive_interval", 10)
 	viper.SetDefault("gateway.image_stream_data_interval_timeout", 900)
 	viper.SetDefault("gateway.image_stream_keepalive_interval", 10)
+	viper.SetDefault("gateway.image_nonstream_keepalive_interval", 0)
 	viper.SetDefault("gateway.max_line_size", 500*1024*1024)
 	viper.SetDefault("gateway.scheduling.sticky_session_max_waiting", 3)
 	viper.SetDefault("gateway.scheduling.sticky_session_wait_timeout", 120*time.Second)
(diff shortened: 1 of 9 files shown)


zh239ns:

zh239ns:
recheck

PR #4254, opened 2026-07-14
fix(grok): probe-gate Free account recovery

superman2003:
## Summary

- mark Free or unclassified Grok OAuth accounts as probe-managed after an upstream 429
- treat the structured `subscription:free-usage-exhausted` error as authoritative Free evidence even when stale billing metadata looks paid, while retaining compatibility with the legacy hyphenated code
- block the in-memory scheduler before persistence and exclude pending accounts from database, cache, group, and dashboard availability paths
- probe pending accounts with Grok 4.5 no more than once every 5 minutes and clear the latch only after a healthy 2xx response
- preserve the latch across generic account tests while allowing the recovery probe to refresh expired OAuth credentials safely
- refresh the affected account row immediately after a manual quota or connection test so the 429 state is visible without reloading the page

## Recovery safety

- explicit SuperGrok/paid evidence and API-key accounts keep the existing reset-based behavior
- pending Free accounts remain out of scheduling even after the ordinary reset timestamp expires
- a 10-minute scheduling lease covers each probe; a one-minute scanner respects the persisted five-minute per-account next-probe time
- a singleton leader lock and three-worker limit avoid multi-instance probe storms
- compare-and-clear checks the probe generation and rate-limit observation, so a stale successful probe cannot clear a newer 429
- normal OAuth requests still reject pending accounts; only the recovery credential context may bypass the pending/rate-limit guard, while active, manual schedulable, expiry, overload, temporary block, proxy, and credential-version checks remain enforced

## UI and counts

- account badges show `429` / recovery probing while pending instead of returning to `normal`
- manual quota and connection tests refresh the account row immediately, including the account badge inside an already-open connection-test dialog
- account filters, group availability counts, and dashboard normal/rate-limited counts use the same pending semantics as the scheduler

## Validation

- focused Grok recovery, quota, runtime-block, token-refresh, CAS, scheduler-cache, DTO, and Wire tests
- Free Messages cache and Messages stream-error regression tests from merged PR #4333
- repository, handler DTO, and server package tests
- frontend focused tests (44), typecheck, lint, and production-path checks
- relevant `go vet`, `gofmt`, and `git diff --check`

No schema migration is required. Based on current `main` and includes merged PR #4333.

Related: #4189


superman2003:
Superseded by #4483, which is rebased onto the latest main, resolves the current conflicts, regenerates Wire from the current source, and preserves the latest scheduler/OAuth/billing behavior.

This remains an important Grok Free pool availability fix: without the persistent recovery gate, accounts that have hit HTTP 429 can re-enter scheduling before upstream quota recovery, repeatedly cycle through failover, and make the Free account pool effectively unavailable.

Please continue review on #4483.

PR #4266, opened 2026-07-14
fix(frontend): update Codex API key snippets for image generation

### LINKED ISSUE #4253: Update Codex config snippets for Responses Lite image generation
## Context

PR #4218 normalized completed hosted image-generation results, while PR #4240 correctly stopped injecting the hosted `image_generation` tool into Responses Lite requests.

As a result, `gpt-5.6-sol` image generation now depends on Codex/Desktop's client-side `image_gen` executor. However, the Codex snippets currently generated by `UseKeyModal.vue` use `requires_openai_auth = true` and provide an `auth.json`. For a custom API-key provider, that configuration does not activate the actor-authorized client image executor.

## Minimal config change

Apply only the following changes to the active custom provider:

```diff
- requires_openai_auth = true
+ requires_openai_auth = false
+ http_headers = { "x-openai-actor-authorization" = "local-image-extension" }
```

Enable the extension in the existing features table:

```diff
+ image_generation = true
```

`requires_openai_auth` controls whether Codex treats this provider as requiring its first-party login/account path; it does not disable upstream authentication. Current Codex versions can continue using the API key already loaded from `auth.json` when this value is `false`. The client must be fully restarted and a new thread created so its tool registry is rebuilt.

## UI request

Please update the normal and WebSocket Codex snippets in `UseKeyModal.vue` to include the minimal settings above while keeping the existing `auth.json` flow. The UI should briefly mention the required restart/new-thread step, label the option along the lines of **`Codex (API Key Mode)`**, and clarify that force injection is a hosted image bridge for non-Responses-Lite requests only.

## Verification

This flow was verified against the same Sub2API endpoint: once the client executor was registered, `image_gen` generated an image, saved it locally, and emitted a displayable image item.

Related: #4218, #4240

StarryKira:
## Summary

- keep the existing Codex configuration as the default Compatibility mode (`requires_openai_auth = true`, without the actor header)
- add an explicit API Key Mode for both normal and WebSocket Codex snippets
- in API Key Mode, set only `requires_openai_auth = false` and `x-openai-actor-authorization = "local-image-extension"`
- preserve `auth.json`, existing feature flags, and WebSocket settings without adding `env_key` or `image_generation`
- add English and Chinese mode guidance plus regression coverage for defaults, switching, and reset behavior

## Root cause

The previous revision changed the generated Codex configuration by default, which could break older Codex clients. The image-executor authorization now stays opt-in so existing users keep the legacy-compatible setup.

## User impact

Existing Codex users see the same configuration as before. Users who need the client-side image executor can select API Key Mode, and the selection carries across the normal and WebSocket tabs while the modal is open.

## Validation

- `pnpm --dir frontend exec vitest run src/components/keys/__tests__/UseKeyModal.spec.ts` (9 tests)
- targeted ESLint
- `pnpm --dir frontend typecheck`
- `make test-frontend` (91 tests)

Closes #4253


skadiUnderTides:
那个issue有些字段不用添加，与问题无关，实测只需要修改这两个即可
requires_openai_auth = false
http_headers = {
  "x-openai-actor-authorization" = "local-image-extension"
}
@StarryKira

StarryKira:
> 那个issue有些字段不用添加，与问题无关，实测只需要修改这两个即可 requires_openai_auth = false http_headers = { "x-openai-actor-authorization" = "local-image-extension" } @StarryKira

ok

skadiUnderTides:
你是直接改了默认的配置吗，那怎么兼容旧版本的调用呢

PR #4296, opened 2026-07-14
fix(openai): repair concatenated Responses stream events

wp-a:
## Problem

An upstream Responses transport can occasionally place multiple complete JSON documents in one SSE `data:` line or one WebSocket v2 message. Downstream OpenAI SDKs parse each event as a single JSON document, so they fail with errors such as `Unexpected non-whitespace character after JSON` even though the first document is complete.

The native HTTP, passthrough HTTP, and WSv2 forwarding paths currently preserve that invalid document boundary.

## Fix

- Detect only the narrow corruption shape: a payload that is invalid as one JSON value but decodes completely into 2-16 typed JSON objects.
- Expand repaired HTTP payloads into separate SSE events in native and passthrough forwarding.
- Queue repaired WSv2 documents in order, and do not return a connection to the pool when a terminal event has trailing documents.
- Cap repair attempts at 16 MiB and reject oversized lines before the string-to-byte conversion.
- Leave valid single JSON values and malformed, truncated, untyped, or debug-tailed payloads unchanged.

This does not change legal multi-line SSE parsing or event flush policy, so it is separate from #1479 and #4275.

## Tests

- Reproduces a 68,106-byte first document followed immediately by a second JSON object.
- Covers native HTTP synchronous and asynchronous scanners, HTTP passthrough, WSv2 ordering, usage preservation, terminal handling, the repair-size cap, and ambiguous WS connection disposal.
- `go test -race ./internal/service -run ConcatenatedJSONDocuments -count=1`
- `go test -tags=integration ./... -count=1`
- `go vet ./...`
- `golangci-lint v2.9`: 0 issues

### DIFF
--- a/backend/internal/service/openai_gateway_passthrough.go
+++ b/backend/internal/service/openai_gateway_passthrough.go
@@ -883,6 +883,7 @@ func (s *OpenAIGatewayService) handleStreamingResponsePassthrough(
 	scanBuf := getSSEScannerBuf64K()
 	scanner.Buffer(scanBuf[:0], maxLineSize)
 	defer putSSEScannerBuf64K(scanBuf)
+	documentScanner := newOpenAISSEJSONDocumentScanner(scanner)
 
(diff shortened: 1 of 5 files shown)


wp-a:

wp-a:
recheck

wp-a:

PR #4302, opened 2026-07-14
fix(apicompat): serialize interleaved parallel tool_use blocks

Avilianb:
## Summary

Fix Anthropic SSE `Content block not found` errors when an OpenAI Responses-compatible upstream interleaves multiple parallel function calls.

The converter previously tracked one global current block. Starting tool 2 closed tool 1, so a later arguments delta for tool 1 could target an already-stopped Anthropic block. Late text/reasoning completion events could also stop whichever block happened to be current.

## What changed

- Track pending tools independently by Responses `output_index` and preserve their original order.
- Keep exactly one active Anthropic `tool_use` block.
- Stream arguments for the active tool immediately; buffer later tools until the active block stops.
- Reconcile final arguments with the already-streamed prefix, so `response.completed` can safely append a missing suffix without duplicating JSON.
- Only let text/reasoning done events stop a matching block type and `output_index`.
- Flush unfinished tools in order before `message_delta` / `message_stop`.
- Keep thinking and text output streaming in real time before tool execution begins.

This intentionally serializes only overlapping tool blocks; it does not buffer the whole response.

## Why this differs from #4294

#4294 prevents one ghost delta by skipping arguments that target an already-closed block. This PR prevents the premature close in the first place and retains every queued tool call's ID, name, and arguments under genuinely interleaved events. It also covers late thinking/text done events and terminal-response argument recovery.

Related: #4193, #4286, #4294.

## Tests

Added lifecycle regression coverage for:

- tool 1 added → tool 2 added → interleaved arguments → ordered tool blocks;
- queued tool arguments arriving before the active tool completes;
- late thinking/text done and output-item done events;
- final arguments available only from `response.completed`;
- every delta/stop referencing an open block and no open block before `message_stop`.

Local verification:

- `go test ./internal/pkg/apicompat -count=1`
- `go test ./... -count=1`
- `go test -tags=integration ./... -count=1`
- `go vet ./internal/pkg/apicompat`

### DIFF
--- a/backend/internal/pkg/apicompat/responses_to_anthropic.go
+++ b/backend/internal/pkg/apicompat/responses_to_anthropic.go
@@ -3,6 +3,7 @@ package apicompat
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


Avilianb:

Avilianb:
## Rebased onto latest `main` (ready for review)

This PR was ~186 commits behind and sitting idle after #4294 landed. I rebased onto current `main` and re-verified.

### Status
- Rebased cleanly onto `main` (includes #4294)
- No merge conflicts
- Force-pushed: `7fb77016` → `93c6d04d`

### Why this is still needed after #4294
#4294 only skips ghost `content_block_delta` targeting an already-closed block. That stops Claude Code from crashing on one failure mode, but it can still **drop** interleaved tool arguments when tool 2 starts and prematurely closes tool 1.

This PR fixes the root cause:
1. Track pending tools by Responses `output_index`
2. Keep exactly one active Anthropic `tool_use` block
3. Buffer later tools until the active block stops (serialize overlapping tool blocks only)
4. Guard late thinking/text done events so they only close a matching block
5. Recover final args from `response.completed` without duplicating JSON

### Verification (post-rebase)
```
go test ./internal/pkg/apicompat -count=1   # PASS
go vet ./internal/pkg/apicompat            # clean
```
Also PASS for the #4294 ghost-delta regressions:
- `TestStreamingParallelToolUseNoGhostDelta`
- `TestStreamingParallelToolUseSecondToolPackedArgsDone`
- `TestStreamingThreeParallelToolsAllPackedDone`
plus the new interleaved queue lifecycle tests in this PR.

@Wei-Shaw please take another look when you have a moment — happy to adjust if you prefer a smaller follow-up on top of #4294.

PR #4333, opened 2026-07-15
fix(grok): cache Free Messages function tools

superman2003:
## Summary

- route known Grok Free `/v1/messages` function-tool requests through xAI's cache-capable mixed-tools shape
- preserve client function declarations and missing/`auto` tool choice while appending `web_search` and `x_search`
- keep paid/unknown tiers, direct Responses, WebSocket, non-auto choices, and unsupported tool shapes unchanged
- fail closed when no tenant-isolated cache identity can be derived

## Why

Grok Free routes function-only requests to `grok-4.5-build-free`, which always reports zero cached tokens. The cache-capable route requires native search tools. xAI accepts native and function tools together with `tool_choice:auto`, preserving structured function calls while enabling prompt cache reads.

## Upstream validation

Using stable identities and repeated prefixes against the Grok CLI proxy:

- native tools + `none`: `cached_tokens=3584`
- function tools + `auto`: model `grok-4.5-build-free`, `cached_tokens=0`
- raw Chat function tools: model `grok-4.5-build-free`, `cached_tokens=0`
- mixed native + function tools + `auto`: second request `cached_tokens=1664`
- mixed tools with an explicit `lookup` request: returned `function_call lookup`; second request `cached_tokens=6144`
- mixed `allowed_tools`: HTTP 422, so this PR does not use it

## Safety boundaries

The mixed route is applied only by the Anthropic Messages bridge, only for accounts whose billing/quota/credential tier is known as Free, and only when a stable cache identity exists. Function tools must use the Responses top-level shape and tool choice must be absent or the string `auto`.

Native search tools are eligible under `auto` and may be selected by the model. The scope is deliberately narrow; existing Responses-to-Anthropic handling already converts native web-search output.

## Tests

- `go test -tags=unit ./internal/service -run 'Test(ApplyGrokCacheIdentity|GrokFreeMessagesFunctionToolCacheRoute|ForwardAsAnthropicForGrok)' -count=1`
- `go test -tags=unit ./internal/pkg/apicompat -run 'Test(AnthropicToResponses|ResponsesToAnthropic)' -count=1`
- full `internal/service` unit package passes with the unrelated flaky `TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig` skipped; that test passes when rerun alone
- `git diff --check`

### DIFF
--- a/backend/internal/service/openai_gateway_grok_cache.go
+++ b/backend/internal/service/openai_gateway_grok_cache.go
@@ -15,6 +15,7 @@ const (
 	grokConversationIDHeader        = "X-Grok-Conv-Id"
 	grokFreeCacheNativeToolsJSON    = `[{"type":"web_search"},{"type":"x_search"}]`
 	grokFreeCacheDisabledToolChoice = "none"
+	grokFreeRolling24hTokenLimit    = int64(2_000_000)
 )
 
 // resolveGrokCacheIdentity derives one stable, tenant-isolated routing identity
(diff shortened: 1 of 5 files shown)


superman2003:
Follow-up from real Free-account deployment:

- xAI leaves billing.plan empty for Free; only paid tiers get SuperGrok / monthly-limit signals. The original gate therefore treated normal Free accounts as unknown and skipped the mixed-tools route.
- The classifier now accepts a successful blank monthly billing observation or the observed 2,000,000-token rolling quota as positive Free evidence. Explicit paid signals always win, including over a stale 2M snapshot.
- Messages streaming now treats a bare Responses type:error event as terminal. Before downstream output it remains failover-safe; after output it emits an Anthropic error event instead of ending as a misleading missing-terminal disconnect.

Additional verification:
- focused service cache/routing/stream-error tests pass
- apicompat Anthropic/Responses tests pass
- go vet ./internal/service passes
- git diff --check passes

Issue #4249, opened 2026-07-14
最新的版本v0.1.153更新后出现了几个报错


viccong:
问题 1：
<img width="2008" height="1764" alt="Image" src="https://github.com/user-attachments/assets/220f03da-3b62-40d8-b9aa-e7ca5bf86874" />

问题2：upstream connect error or disconnect/reset before headers. reset reason: connection termination

<img width="2014" height="1246" alt="Image" src="https://github.com/user-attachments/assets/6d4be7a0-3f90-4f50-8e66-818207bfc4b2" />

问题 3："Invalid Value: 'tools'. Function 'image_gen.imagegen' conflicts with a hosted tool in the same request.",

<img width="2034" height="1422" alt="Image" src="https://github.com/user-attachments/assets/67b98850-919a-473f-b2ef-6aa94c3e67d1" />

ye740891000:
ok 的话 踢我一脚 刚更新了v0.1.155
 观望一下

Vicsosoa:
look....

SatellaPoi:
基本上全中，都遇到了

MenXianWe:
更新不行

Wei-Shaw:
504问题参考 https://github.com/Wei-Shaw/sub2api/issues/4232

PR #4104, opened 2026-07-12
[codex] avoid duplicate hosted image tool injection

gebdalaoli-arch:
## Summary

- reuse the existing comprehensive image-tool detector before the Codex bridge injects a hosted `image_generation` tool
- preserve Codex `namespace/image_gen` declarations in both top-level `tools` and Responses Lite `input[].additional_tools`
- add helper-level and HTTP forwarding regression coverage

## Root cause

`hasOpenAIImageGenerationTool` already recognized hosted `image_generation`, top-level `namespace/image_gen`, and Responses Lite `additional_tools`. However, `ensureOpenAIResponsesImageGenerationTool` used a separate, narrower check that only looked for a top-level `type=image_generation` tool.

When a Codex request already carried `image_gen.imagegen`, the bridge therefore appended a hosted `image_generation` tool to the same request. OpenAI rejects that combination as a conflicting hosted tool.

The injection path now uses the shared detector, keeping HTTP and WebSocket behavior consistent and skipping hosted-tool injection whenever either supported image-tool representation is already present.

## Validation

- `go test ./internal/service -run 'ImageGeneration|CodexBridge|ImageGenNamespace' -count=1`
- `go test ./internal/service -count=1`
- `go test ./... -count=1`
- `go test -tags=unit ./... -count=1`
- `go test -tags=integration ./... -count=1`
- `golangci-lint run ./... --timeout=30m` (`0 issues`)

Fixes #4022

PR #4346, opened 2026-07-15
fix(openai): preserve Codex image function tools during bridge injection

xiaohei210509:
## Summary

- recognize Codex client image tools serialized as flat or nested `image_gen.imagegen` functions
- skip hosted `image_generation`, automatic `tool_choice`, and hosted-bridge instructions when that client function is already present
- add helper, HTTP forwarding, and WebSocket ingress regression coverage

## Root cause

PR #4104 made bridge injection reuse the image-tool detector for hosted tools and `namespace/image_gen` declarations. Some Codex clients and MCP adapters instead serialize the same client-side tool as one of these function shapes:

```json
{"type":"function","name":"image_gen.imagegen"}
{"type":"function","function":{"name":"image_gen.imagegen"}}
```

Those forms are not OpenAI hosted tools and are intentionally not part of `hasOpenAIImageGenerationTool`. As a result, the bridge still appended a hosted `image_generation` tool to the same request, which OpenAI rejects with:

```text
Function 'image_gen.imagegen' conflicts with a hosted tool in the same request.
```

This change reuses the existing exact-name function parser before applying any hosted bridge additions. Similar names such as `image_gen.imagegenerator` continue through the existing injection path.

This is a follow-up to #4104. Refs #4022 and #4249.

## Validation

- targeted helper, HTTP, and WebSocket regression tests
- `go test ./... -count=1`
- `go test -tags=integration` across all packages except `internal/pkg/tlsfingerprint`
- `golangci-lint run ./... --timeout=30m` (`0 issues`)

The full integration command was also run. Its only failure was the unrelated external TLS fingerprint endpoint intermittently returning `unexpected EOF`; all remaining integration packages passed locally.

### DIFF
--- a/backend/internal/service/openai_codex_transform.go
+++ b/backend/internal/service/openai_codex_transform.go
@@ -83,6 +83,8 @@ type codexOAuthTransformOptions struct {
 	PreserveToolCallIDs     bool
 }
 
+const codexImageGenerationFunctionToolName = "image_gen.imagegen"
+
 const (
 	codexImageGenerationBridgeMarker = "<sub2api-codex-image-generation>"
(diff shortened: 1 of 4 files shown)


xiaohei210509:

PR #4349, opened 2026-07-15
fix(openai): allow Agent Identity Codex JSON import

catoncat:
## Problem

The Codex JSON importer recognizes `auth_mode=agentIdentity`, but then applies the OAuth access-token expiry policy unconditionally. Agent Identity has neither an OAuth `refresh_token` nor an access-token expiry, so a valid Agent Identity JSON is rejected before an account can be created.

## Fix

Skip OAuth expiry and auto-pause handling for recognized Agent Identity imports. Runtime/task lifecycle remains handled by the existing Agent Identity task recovery path.

## Verification

- Added an import-flow test that creates an Agent Identity account from Codex auth JSON without OAuth credentials.
- `go test ./internal/handler/admin -count=1`

### DIFF
--- a/backend/internal/handler/admin/account_codex_agent_identity_import_test.go
+++ b/backend/internal/handler/admin/account_codex_agent_identity_import_test.go
@@ -1,6 +1,7 @@
 package admin
 
 import (
+	"context"
 	"crypto/ed25519"
 	"crypto/rand"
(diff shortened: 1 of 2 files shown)


catoncat:
Closing this draft while the end-to-end account import experience is validated on a deployed instance.

PR #4368, opened 2026-07-15
fix(gateway): bound default SSE max_line_size to 40MiB (#4365)

Doone-skser:
## Summary

- Lower the built-in `gateway.max_line_size` default from **500 MiB** to **40 MiB**, matching `deploy/config.example.yaml`.
- Keep `service.defaultMaxLineSize` and the viper default in sync so Docker installs that omit the key no longer allow multi-hundred-MiB SSE scanner growth per concurrent stream.
- Add a regression test so the default cannot silently climb above 64 MiB again.

Fixes / mitigates https://github.com/Wei-Shaw/sub2api/issues/4365

### Why

`bufio.Scanner` grows its token buffer up to `MaxLineSize`. With the old 500 MiB default, a few concurrent Codex/long-context streams could allocate multi-GiB private anonymous memory. Restart temporarily frees RSS, but normal traffic recreates the growth within minutes.

### Notes for operators

- Existing configs that already set `gateway.max_line_size` are unchanged.
- Hosts that intentionally need larger single SSE lines can still raise the config value (minimum validated size remains 1 MiB).
- This is a defensive default bound; further request-body lifecycle work may still help on very small hosts.

## Test plan

- [x] Unit test `TestDefaultMaxLineSizeIsBoundedForLowMemoryHosts`
- [ ] `go test ./internal/service/ -run TestDefaultMaxLineSize`
- [ ] Deploy on a low-memory host previously hit by #4365 and confirm RSS no longer climbs to multi-GiB under light Codex traffic
- [ ] Confirm large tool-result SSE lines under 40 MiB still stream successfully

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -838,7 +838,7 @@ type GatewayConfig struct {
 	ImageStreamKeepaliveInterval int `mapstructure:"image_stream_keepalive_interval"`
 	// ImageNonstreamKeepaliveInterval: 图片非流式 JSON keepalive 间隔（秒），0表示禁用
 	ImageNonstreamKeepaliveInterval int `mapstructure:"image_nonstream_keepalive_interval"`
-	// MaxLineSize: 上游 SSE 单行最大字节数（0使用默认值）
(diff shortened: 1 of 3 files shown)


Doone-skser:

PR #4372, opened 2026-07-15
[codex] preserve typed Responses SSE event boundaries

wp-a:
## Problem

Responses transports do not always preserve one logical event per physical SSE frame. In production we observed large `response.in_progress` payloads near 68 KiB arrive with another typed event, an SSE prefix, or `[DONE]` attached without the expected blank boundary. Official OpenAI SDKs parse each `data:` value as one JSON document, so the stream fails with `Unexpected non-whitespace character after JSON` before application code can recover.

PR #4296 repairs the narrow `{json1}{json2}` shape, but the scanner still forwards several equivalent boundary corruptions unchanged:

- multiple complete typed `data:` lines in one frame;
- a missing blank line before the next `event:`;
- `data:` or `event: ... data:` text attached after a complete JSON document;
- a typed JSON document immediately followed by `[DONE]`;
- the same frame-state transitions after a data line exceeds the 16 MiB repair limit.

## Root cause

The document scanner was line-local. After forwarding one complete typed payload, it did not retain enough frame state to recognize that a following typed line or event header belonged to a new logical event. Oversized valid typed payloads bypassed repair without updating that state at all.

## Fix

- Track the current typed event and whether its complete JSON payload has already been seen.
- Buffer ambiguous post-data SSE fields as one bounded group, and restore a missing boundary only when another typed payload or `[DONE]` proves that a new logical event started.
- Recognize bounded embedded SSE prefixes and `[DONE]` tails after a complete typed JSON document.
- Preserve legal `event:` overrides and exact blank-line / EOF behavior.
- Bound candidate field groups to 256 lines and 16 MiB; fail closed instead of allowing ambiguous unbounded buffering.
- Preserve the existing 16-document and 16 MiB concatenated-document repair bounds.
- Keep data-only open-event behavior compatible with the current first-output staging and flush logic.

This is a focused follow-up to #4296. It does not change downstream flush scheduling (#4275), first-output staging (#4307), or the configured scanner line limit (#4368).

## Tests

New regression coverage includes multiple typed data lines, missing blank boundaries, embedded `data:` / `event:` prefixes, JSON plus `[DONE]`, legal event-type overrides, event/id/comment field groups, EOF replay, field-group safety limits, and oversized typed events followed by another typed event or terminator.

Verification:

- `go vet ./internal/service`
- focused SSE boundary tests
- `git diff --check`

### DIFF
--- a/backend/internal/service/openai_sse_concatenated_json_test.go
+++ b/backend/internal/service/openai_sse_concatenated_json_test.go
@@ -28,6 +28,235 @@ func TestOpenAIStreamingPassthroughRepairsConcatenatedJSONDocumentsInSingleDataL
 	testOpenAIStreamingRepairsConcatenatedJSONDocuments(t, true, 0)
 }
 
+func TestOpenAISSEScannerRepairsMultipleTypedDataLinesInSingleFrame(t *testing.T) {
+	largeInProgress, outputItemAdded, completed := openAIConcatenatedJSONTestEvents(t)
+	input := strings.Join([]string{
+		"event: response.in_progress",
+		"data: " + largeInProgress,
+		"data: " + outputItemAdded,
+		"",
+		"event: response.completed",
+		"data: " + completed,
+		"",
+	}, "\n")
+
+	assertOpenAISSEFrames(t, scanOpenAISSEJSONDocumentsForTest(t, input), []string{
+		"response.in_progress",
+		"response.output_item.added",
+		"response.completed",
+	})
+}
+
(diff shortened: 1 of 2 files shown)


wp-a:
CI note: the `test` job reached the integration suite and failed only in `internal/pkg/tlsfingerprint/TestAllProfiles` because `https://tls.peet.ws/api/all` reset the TLS connection. The same external failure occurred on both #4372 and #4373; the changed packages and local full test suites pass. I attempted to rerun the failed job, but fork contributors do not have repository admin permission to rerun it. Please rerun the failed job when convenient.

PR #4373, opened 2026-07-15
[codex] preserve usage ranges in the user timezone

wp-a:
## Problem

Usage dashboards currently mix browser-local dates, server-local timestamps, and fixed 24-hour arithmetic. This causes visible date drift for users outside the server timezone and incorrect ranges across daylight-saving transitions, where a local calendar day can be 23 or 25 hours.

Examples include seven-day and thirty-day ranges computed with `86400000`, daily trend buckets formatted directly from `created_at`, and inclusive end dates recovered with `EndTime.Add(-24h)`.

## Root cause

The frontend did not consistently send the user's IANA timezone, while the backend parsed and grouped several usage queries without an explicit timezone contract. Fixed-duration subtraction was also used where the product semantics are calendar days.

## Fix

- Compute local date ranges with calendar-day operations instead of fixed milliseconds.
- Preserve an explicitly supplied timezone in the GET client interceptor.
- Send the browser's IANA timezone for dashboard and key-usage requests.
- Parse gateway and user usage ranges from civil dates in the requested timezone, including zones whose first valid instant is after midnight.
- Reject a directly requested civil date that does not exist because the zone skipped the entire day, while safely finding the next valid boundary for surrounding ranges.
- Group PostgreSQL daily trends with a parameterized `AT TIME ZONE` expression.
- Use calendar-day subtraction for inclusive end dates across DST boundaries.
- Reuse the same timezone for daily usage aggregation.
- Pass the same civil-day lower bound through API-key Today summaries so cards, trends, models, and recent usage share one timezone contract.

The API remains backward compatible: callers that omit `timezone` retain the existing default behavior.

## Tests

Regression coverage includes 23-hour and 25-hour DST days, Santiago's midnight DST gap, Apia's skipped civil date, explicit timezone preservation, browser timezone propagation, timezone-aware backend range parsing and trend grouping, API-key Today boundaries, and inclusive end-date behavior.

Verification:

- `go test ./internal/handler ./internal/repository ./internal/service -count=1`
- targeted frontend tests: 5 files, 27 tests
- `pnpm typecheck`
- `pnpm lint:check`
- `git diff --check`

### DIFF
--- a/backend/internal/handler/gateway_handler.go
+++ b/backend/internal/handler/gateway_handler.go
@@ -1259,7 +1259,7 @@ func (h *GatewayHandler) Usage(c *gin.Context) {
 	}
 
 	// Best-effort: 获取用量统计（按当前 API Key 过滤），失败不影响基础响应
-	usageData := h.buildUsageData(ctx, apiKey.ID)
+	usageData := h.buildUsageData(ctx, apiKey.ID, apiKeyTodayStart(c, time.Now()))
 	dailyUsage := h.buildAPIKeyDailyUsage(c, subject.UserID, apiKey.ID, days)
 
 	// Best-effort: 获取模型统计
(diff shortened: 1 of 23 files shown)


wp-a:
CI note: the `test` job reached the integration suite and failed only in `internal/pkg/tlsfingerprint/TestAllProfiles` because `https://tls.peet.ws/api/all` reset the TLS connection. The same external failure occurred on both #4372 and #4373; the changed packages and local full test suites pass. I attempted to rerun the failed job, but fork contributors do not have repository admin permission to rerun it. Please rerun the failed job when convenient.

PR #2096, opened 2026-04-29
feat: multi-dimensional service quota + Advisor Tool rectifier

touwaeriol:
## 背景 / Background

网关层目前缺少两块能力，外加渠道监控编辑表单的一处 UX 漏洞：
1. 细粒度限流 —— 现有 RPM 限流粒度太粗，不能区分用户/分组/账号/渠道/路径，也不能按 token、daily-USD 等多维度配额限制；
2. Advisor Tool 整流 —— 部分对齐 Anthropic 官方 SDK 的上游开始拒绝 advisor-tool-2026-03-01 这条 anthropic-beta header 以及 tools[type=advisor_*]，导致客户端透传请求失败；
3. 渠道监控 API Key 字段被浏览器/密码管理器自动填充 —— 使用者可能在不感知的情况下用旧的"假填"值覆盖已保存的 key。

The gateway is missing two capabilities, plus a UX vulnerability in the channel monitor edit form:
1. Fine-grained rate limiting — existing RPM limits cannot be scoped per user/group/account/channel/path or expressed as multi-dimension quotas (rpm/tpm/tpd/concurrency/daily-USD).
2. Advisor Tool rectifier — some upstream providers (those tracking the official Anthropic SDK strictly) reject the `advisor-tool-2026-03-01` anthropic-beta token and `tools[type=advisor_*]`, breaking pass-through requests.
3. Channel monitor API Key input is auto-filled by browsers/password managers, so users may unwittingly resubmit a stale browser-saved value and overwrite the stored key.

---

## 目的 / Purpose

一次性补齐这两块能力，并保持与现有 signature/budget rectifier、billing pipeline 的复用；同时修掉渠道监控编辑表单的密码自动填充问题。

Add both capabilities in a single PR — reusing existing signature/budget rectifier and billing pipeline infrastructure — and fix the channel monitor edit form's password-autofill issue.

---

### 后端 / Backend

- **Service Quota（服务配额）**
  - 新增 `service_quota_*` package：rule CRUD、两阶段 precheck（PreCheckSelect + PreCheckAcquire）pipeline、deferred-batch limiter（fixed-window / rolling-window / concurrency）、residual counter cleanup、admin & user handlers、metrics、migrations 135-145。
  - 复用 `BillingTicket` 把 quota lease 嵌入现有 pre-flight / post-flight 链路。
  - `SnapshotForAccounts` 批量查询 + 账号可调度性预过滤（200ms 超时 + fail-open）。
[...]

PR #4425, opened 2026-07-16
feat(admin): batch update user concurrency and RPM

AdrianZhaoDev:
## Background

The admin users page only supported editing limits one user at a time. This PR adds controlled multi-select and a single bulk endpoint for overwriting existing user concurrency and RPM limits without issuing one request per user.

## API contract

Adds `POST /api/v1/admin/users/batch-limits`:

```json
{
  "user_ids": [1, 2, 3],
  "all": false,
  "concurrency": 10,
  "rpm_limit": 60
}
```

- `concurrency` and `rpm_limit` are optional pointer fields; at least one is required.
- Explicit `0` values are preserved. `rpm_limit: 0` means unlimited.
- `all: false` requires 1-500 user IDs; `all: true` targets all non-deleted users.
- Both fields use overwrite semantics only.
- The repository performs one SQL update and returns `affected`.
- Auth cache invalidation runs per selected user only after the repository update succeeds.

## Admin UI

- Adds current-page select all and controlled row selection to `DataTable`, including mobile cards.
- Keeps selected IDs while paging through `admin/users`.
- Shows the bulk limits action only when users are selected.
- Adds independent concurrency and RPM toggles, validation, explicit RPM unlimited copy, and overwrite confirmation.
- Prevents submission when more than 500 users are selected.
- Refreshes the user list and clears selection after success.

## Compatibility

- Keeps `POST /api/v1/admin/users/batch-concurrency` unchanged.
- Adds no schema migration and does not change concurrency adjustment records.
- Preserves the current user RPM global-hard-cap behavior discussed in #3288; this PR only changes how admins write the existing field.
- This is a focused bulk editor for existing fields, not the multi-dimensional quota system proposed in #2096.
- PR #2202 remains the precedent for the existing bulk concurrency endpoint; this endpoint has a separate overwrite-only contract covering both fields.

## Verification

Passed:

- `go1.26.5 test -tags=unit ./internal/handler/admin ./internal/service -run BatchUpdateLimits -count=1`
- `go1.26.5 test -tags=integration ./internal/repository -run TestUserRepoSuite/TestBatchUpdateLimits -count=1 -v`
- `golangci-lint run ./... --timeout=30m` (`0 issues`)
- `pnpm run lint:check`
- `pnpm run typecheck`
- Feature Vitest suites: 4 files, 22 tests passed
- `pnpm run build`

Known upstream baseline issues encountered while running the full suites:

- Full Go unit/integration runs fail at `TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig` with `Condition never satisfied`. The same focused command fails on a clean `upstream/main` worktree.
- `pnpm run test:run` passes the feature suites but cannot resolve the undeclared direct dependency `@intlify/message-compiler` imported by `localesMessageCompile.spec.ts`.

Refs #2202, #3288, #2096.


AdrianZhaoDev:

PR #4443, opened 2026-07-16
fix(openai): strip unsupported input namespace fields

margetrp-hub:
## Summary
Fix OAuth passthrough requests that contain unsupported `namespace` fields inside the Responses `input` payload.

## Behavior
- Recursively removes `namespace` only below the top-level `input` field.
- Preserves namespaces in `tools` and other top-level fields.
- Uses `json.Decoder.UseNumber()` so large JSON numbers are not rounded during normalization.
- Leaves the request unchanged when no input namespace is present.

This targets the OAuth passthrough path and complements the native Responses namespace handling.

## Validation
- `go test -p 2 -tags=unit ./internal/service -run TestNormalizeOpenAIPassthroughOAuthBody -count=1`

### DIFF
--- a/backend/internal/service/openai_gateway_request_body.go
+++ b/backend/internal/service/openai_gateway_request_body.go
@@ -563,6 +563,13 @@ func normalizeOpenAIPassthroughOAuthBody(body []byte, compact bool) ([]byte, boo
 	normalized := body
 	changed := false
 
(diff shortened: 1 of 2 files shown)


margetrp-hub:

margetrp-hub:
recheck

PR #4463, opened 2026-07-17
fix(auth): avoid binding sessions to untrusted proxy IPs

### LINKED ISSUE #4452: 我的sub2api 登不上了,一登录进后台就秒过期退出登录
刚刚自动更新到最新版以后就出现这个问题.回退部署到老版本也不行,一进后台就被秒踢.

aomsir:
## Summary

- include the client IP in the session fingerprint only when Gin successfully configures an explicit trusted proxy chain
- fall back to User-Agent-only session binding when `server.trusted_proxies` is empty or invalid
- reuse the request-context binding for both token issuance and validation
- add regression tests for trusted proxy resolution, proxy peer changes, and User-Agent-only fallback

## Root cause

When `server.trusted_proxies` is empty, Gin returns the direct peer address instead of a verified client address.

Behind a distributed reverse proxy such as Cloudflare, consecutive requests may arrive through different edge nodes. The login request can therefore issue a token using one edge address, while the first authenticated request computes the session fingerprint using another. This causes an immediate `SESSION_BINDING_MISMATCH` and revokes the newly created session family.

A similar inconsistency could occur when `server.trusted_proxies` is non-empty but Gin rejects the configuration. The configuration previously remained non-empty even though the trusted proxy chain was not active.

## Behavior

| Trusted proxy state | Session binding |
| --- | --- |
| Successfully configured | Resolved client IP + User-Agent |
| Empty | User-Agent only |
| Invalid or rejected by Gin | User-Agent only |

If trusted proxy configuration fails, the server explicitly disables Gin's trusted proxy chain before falling back to User-Agent-only binding.

This change does not directly trust `CF-Connecting-IP`, `X-Forwarded-For`, or any other forwarding header without a successfully configured Gin trusted proxy chain.

Existing behavior for genuine binding mismatches is unchanged: the session family is revoked, a security audit event is recorded, and the request returns `SESSION_BINDING_MISMATCH`.

## Related work

This complements #4462. That PR unifies security-sensitive client IP selection behind the existing forwarded-IP trust toggle. With the toggle disabled by default, the direct proxy peer can still be unstable. This PR provides the safe fallback for that state by excluding IP from the session fingerprint unless Gin has successfully configured an explicit trusted proxy chain.

## Validation

Passed:

- `cd backend && make test-unit`
- `cd backend && go test -tags=unit ./internal/server -run '^TestConfigureTrustedProxies' -count=1`
- `cd backend && go test -tags=unit ./internal/server/middleware -run 'SessionBinding' -count=1`
- `cd backend && GOTOOLCHAIN=go1.26.5 go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.9.0 run --timeout=30m`
- `pnpm install --frozen-lockfile`
- `make test-frontend`
- `git diff --check`

Integration tests were attempted but could not complete because Docker Hub image pulls for `redis:8.4-alpine` and `postgres:18.1-alpine3.23` failed with TLS handshake timeouts. The failures occurred while Testcontainers was obtaining its dependencies, before the affected assertions could run.

Fixes #4452

### DIFF
--- a/backend/internal/server/http.go
+++ b/backend/internal/server/http.go
@@ -47,14 +47,8 @@ func ProvideRouter(
 
 	r := gin.New()
 	r.Use(middleware2.Recovery())
-	if len(cfg.Server.TrustedProxies) > 0 {
-		if err := r.SetTrustedProxies(cfg.Server.TrustedProxies); err != nil {
-			log.Printf("Failed to set trusted proxies: %v", err)
-		}
-	} else {
-		if err := r.SetTrustedProxies(nil); err != nil {
-			log.Printf("Failed to disable trusted proxies: %v", err)
-		}
+	includeSessionBindingIP := configureTrustedProxies(r, cfg.Server.TrustedProxies)
+	if len(cfg.Server.TrustedProxies) == 0 {
 		if cfg.Server.Mode == "release" {
 			log.Printf("Warning: server.trusted_proxies is empty in release mode; client IP trust chain is disabled")
 		}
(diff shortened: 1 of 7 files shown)


aomsir:

PR #4468, opened 2026-07-17
fix(apicompat): emit content_part events and full output in Responses stream

docooler:
## Problem

The Anthropic→Responses streaming converter omits two things the OpenAI Responses wire format requires. Clients that only read `output_text.delta` are fine; clients that **reconstruct the response from the event stream** break.

### 1. `response.content_part.added` is never emitted

A message item is opened with `content: []`. The OpenAI SDK's accumulating stream helper (`client.responses.stream`) only appends a content part when it sees `content_part.added`. Without it, the next `output_text.delta` indexes into an empty list:

```
File "openai/lib/streaming/responses/_responses.py", line 352, in accumulate_event
    content = output.content[event.content_index]
IndexError: list index out of range
```

Raw iteration (`responses.create(stream=True)`) doesn't accumulate and is unaffected — which is why this went unnoticed.

### 2. `response.completed` carries `Output: []ResponsesOutput{}`

`get_final_response()` and tracing integrations parse the terminal event's `response` directly, so callers get an empty `output_text` even though the deltas streamed correctly. This one is invisible when watching the stream render — the text shows up fine, only the reconstructed object is empty.

## Reproduce

Against any Anthropic-platform group, with `openai` 2.46.0:

```python
with client.responses.stream(model="claude-sonnet-4-5", input="hi") as s:
    for ev in s: pass
    print(s.get_final_response().output_text)
```

Before: `IndexError: list index out of range`.
After: the response text.

Arize Phoenix's playground hits this same path (its OpenAI provider defaults to the Responses API), which is where I ran into it.

## Changes

- Emit `response.content_part.added` when a message text part opens, before its first delta.
- Accumulate output items over the stream and carry them in the terminal event.
- Carry full text on `output_text.done` / `content_part.done` (deltas carry increments, done events carry the whole part).
- Fill in `content` / `arguments` / `summary` on `output_item.done`, which had the same empty-payload issue.

All additive to the event stream — clients that ignore the new events are unaffected.

## Tests

Adds `anthropic_to_responses_stream_test.go` covering event ordering (`content_part.added` precedes the first delta), done-event payloads, and the terminal event's output for both text and tool calls.

`go test ./internal/pkg/apicompat/... ./internal/service/... ./internal/handler/...` passes.

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
+++ b/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
@@ -152,10 +152,24 @@ type AnthropicEventToResponsesState struct {
 
 	// For message output: accumulate text parts
 	ContentIndex int
+	// TextAccum accumulates the current text part so that output_text.done and
+	// content_part.done can carry the full text (deltas carry increments only).
+	TextAccum string
 
 	// For function_call: track per-output info
 	CurrentCallID string
 	CurrentName   string
+
+	// Content of the currently open item, folded into Outputs when it closes.
+	CurrentContent []ResponsesContentPart // message
+	CurrentArgs    string                 // function_call
+	CurrentSummary string                 // reasoning
+
+	// Outputs accumulates every closed output item so that response.completed
+	// can carry the full output list. The OpenAI SDK's get_final_response()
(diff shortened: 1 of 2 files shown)


docooler:

Issue #4471, opened 2026-07-17
GROK授权登录的接口生图偏离根本对不上啊，是工具问题还是sub平台问题


Yangshengyi035717:
官方生图没问题，下面官方网页版的

<img width="1360" height="846" alt="Image" src="https://github.com/user-attachments/assets/3ddec291-80fb-48b3-9d58-ffc0d700fa03" />

这是用的中转，这给我生的是啥，图2也看出调用image模型
1.
<img width="827" height="1131" alt="Image" src="https://github.com/user-attachments/assets/73fe2825-43af-48ef-8681-63753ef871fe" />
2.

<img width="1618" height="176" alt="Image" src="https://github.com/user-attachments/assets/1ab639b6-0f99-4710-8681-7a85dcd9d50f" />

Yangshengyi035717:
有没有人遇到同样的问题

Rotor1996:
一样的bro

cshxql:
同样遇到了，头大！

Yangshengyi035717:
> 同样遇到了，头大！

没更新前还能用，就是今天更新了就用不了了

cshxql:
> > 同样遇到了，头大！
> 
> 没更新前还能用，就是今天更新了就用不了了

应该不是更新导致的，我最近都没更新过，突然就这样了，应该是官方动了啥

Yangshengyi035717:
> > > 同样遇到了，头大！
> > 
> > 
> > 没更新前还能用，就是今天更新了就用不了了
> 
> 应该不是更新导致的，我最近都没更新过，突然就这样了，应该是官方动了啥

那还是寄托sub2来做优化适配吧

heathermhuang:
This appears to be the same API-only prompt-divergence symptom tracked in #4420: generation works in the official web UI, while the API returns an unrelated image.

#4474 addresses two relevant gateway-side failure modes: it normalizes legacy `image_url` references to xAI's current `url` payload shape, and it prevents OAuth accounts with a recorded weekly or monthly billing `403` from receiving new media-generation requests. Accounts that return HTTP 200 but are known to produce fallback media can be quarantined explicitly with `extra.grok_media_eligible=false`.

The PR is open and CI is green, but it has not been merged or released upstream yet.

cshxql:
这个PR，没有跟最新版本发布，啥时候发布这个呀？

cshxql:
[...]

PR #4433, opened 2026-07-16
fix(grok): repair OAuth media and compatibility regressions

heathermhuang:
## Summary

Repair the Grok regressions reported after the recent OAuth routing changes:

- route OAuth image/video traffic through the official xAI API so large Base64 image-to-video bodies no longer hit the CLI proxy body limit;
- preserve the CLI proxy as the primary text/account-test route, but retry only the exact replayable OAuth `403 Access denied` compatibility response against the official API;
- advertise Grok 4.5 reasoning-effort capabilities in `/v1/models` for Grok Build clients.

## Root cause and scope

- #4412 regressed when OAuth media was pinned back to `cli-chat-proxy.grok.com`; the proxy rejects large Base64 request bodies. Media now uses `api.x.ai` while explicit non-default custom media hosts remain supported.
- #4420 began in the same media-routing window. Restoring the official media route avoids the affected proxy behavior for image generation as well as video.
- #4419 and #4421 report the CLI proxy returning `403 Access denied` for credentials that can remain valid on the official API. The fallback is deliberately narrow: exact CLI host, OAuth bearer request, CLI identity header, replayable body, exact 403 body marker, and one successful 2xx retry. Other 403s and failed fallbacks preserve the original response.
- #4409 was a model-discovery contract gap: Grok model entries did not expose the reasoning-effort fields consumed by Grok Build.

## Validation

- focused Grok service/repository/handler tests
- `go test -tags=unit ./... -count=1`
- `go vet ./...`
- `git diff --check`
- private production canary on both Grok OAuth and API-key groups:
  - `/v1/models` capability metadata
  - Responses API with `reasoning.effort=high`
  - image generation
  - large-Base64 image-to-video generation and completion
  - usage-log account/group attribution
  - disposable API-key deletion and 401 confirmation

Fixes #4409
Fixes #4412
Fixes #4419
Fixes #4421
Refs #4420

Issue #4421, opened 2026-07-16
super grok账号在sub2上面403，但是在网页上面又能用什么问题


Yangshengyi035717:
重新授权登录也是403
<img width="890" height="618" alt="Image" src="https://github.com/user-attachments/assets/71aa9beb-8f44-4060-b64a-93e98a5111d2" />

<img width="283" height="149" alt="Image" src="https://github.com/user-attachments/assets/74beded5-51fb-4951-bf4e-317c8f674ab1" />

Peanut13:
你是试用账号吗？好像现在试用账号直登grok build也不行

Yangshengyi035717:
订阅的和试用的

Peanut13:
我试了目前7天订阅的不行，正常订阅的可以用，可能xai那边在调整？

Yangshengyi035717:
今天才开始的好像，我昨天还能用7天的，今天突然就用不了了

Yangshengyi035717:
<img width="482" height="571" alt="Image" src="https://github.com/user-attachments/assets/ae73b8ff-e6f3-45ed-b3d7-e8d38d5b9a37" />

Yangshengyi035717:
<img width="468" height="158" alt="Image" src="https://github.com/user-attachments/assets/485eb420-a459-445d-b79b-23b4479cc8e6" />

cczuishuai233:
回退到 0.151可以用  别用最新版

Yangshengyi035717:
> 回退到 0.151可以用 别用最新版

回退那其他平台的是不是也会出bug

CoolLiuzw:
> 回退到 0.151可以用 别用最新版

真的吗？不是 grok 限制吗？

Yangshengyi035717:
> > 回退到 0.151可以用 别用最新版
> 
> 真的吗？不是 grok 限制吗？

只有少数部分能用，但是大部分都有问题

heathermhuang:
We need to distinguish a genuine xAI account or subscription restriction from this Sub2API compatibility regression.

#4433 handles one narrow case: an OAuth request receives `403 Access denied` from the CLI proxy, but the same credential remains valid against the official API. A successful fallback is only returned when the request can be safely replayed, the response matches that marker, and the official API retry succeeds. Expired accounts, missing entitlements, and other 403 responses still fail normally, so this does not bypass genuine subscription restrictions.

[...]

Issue #4420, opened 2026-07-16
我的grok生图感觉有自己的想法


Rotor1996:
说小猫，给灯塔，关键三天两头给同样的图。 今天上午开始出现的

heathermhuang:
The timing of this report matches the OAuth media-routing regression window: a later OAuth security change sent image requests back through the CLI proxy instead of the official Imagine API.

#4433 restores official-API routing for OAuth media while preserving API-key and custom-host behavior. In a private production canary, both OAuth and API-key image requests returned HTTP 200. The requests used different prompts, produced distinguishable outputs with different SHA-256 hashes, and both subsequent image-to-video requests reached `done`.

This addresses the routing regression. If repeated fixed images still occur after the fix is merged, please provide a new prompt, timestamp, and request ID so an upstream generation issue can be investigated separately from routing.

Fix PR: #4433

put-go:
我今天也遇到了，我通过规律发现是那些便宜的周卡订阅会有这个问题，换成正价账号就没有这个问题了

put-go:
@heathermhuang  我使用最新版本复测后确认，该问题会在部分特定订阅账号上稳定出现。

例如提示词明确要求“小猫跳舞”，但任务几乎瞬间生成成功，返回的内容却与提示词和原图明显不符。多次测试后发现，结果似乎是在少量已有视频中随机选取或拼接，更像是触发了某种降级或兜底逻辑，而非正常执行视频生成任务。

我已经通过邮件发送了详细的复现过程，以及可用于复现问题的 Grok 测试账号。

pxj2026:
补充：同样遇到相同问题，这部分账号在网页上生视频是正常的，但走API生成就会出现生成完全无关的视频。唯一异常是 weekly/monthly billing 查询都返回 403 Access denied。但网页上确实是supergrok。

heathermhuang:
I opened #4474 with two focused fixes:

1. Grok media requests now normalize legacy `image_url` references to xAI's current `url` payload shape for `image`, `images`, `reference_images`, and `mask`.
2. New image/video generation uses a media-only eligibility gate. OAuth accounts with a recorded weekly or monthly billing `403` are excluded automatically, while chat and existing video status lookups remain available.

[...]

PR #4484, opened 2026-07-17
fix(grok): preserve media eligibility in scheduler cache

heathermhuang:
## Summary

- Preserve `grok_media_eligible` and `grok_billing_snapshot` in the scheduler's slim Redis account metadata.
- Keep Grok media eligibility decisions consistent between fresh database reads and cached scheduler reads.
- Add regression coverage for an explicit media override and for a forbidden weekly/monthly billing observation.

## Why

#4474 added Grok media eligibility filtering, but `filterSchedulerExtra` still dropped the two fields consumed by `GrokMediaGenerationEligibility`. After the scheduler cache warmed, an OAuth account could therefore become eligible for media again even though the database record still marked it ineligible.

This follow-up keeps only the two small routing fields needed by the scheduler; unrelated account metadata remains filtered out.

## Verification

- `go test -tags=unit ./internal/repository ./internal/service`
- `go vet -tags=unit ./internal/repository ./internal/service`
- `git diff --check origin/main..HEAD`
- `gitleaks git --log-opts='origin/main..HEAD'`
- Controlled cache-path E2E: text remained available, ineligible OAuth image/video requests were rejected before upstream dispatch, API-key media remained routable, and account/cache state was restored afterward.

Follow-up to #4474.

Related reports: #4420, #4471, #4421, #4419, #4412, #4433.

### DIFF
--- a/backend/internal/repository/scheduler_cache.go
+++ b/backend/internal/repository/scheduler_cache.go
@@ -961,6 +961,8 @@ func filterSchedulerExtra(extra map[string]any) map[string]any {
 		"auto_pause_7d_disabled",
 		"model_rate_limits",
 		service.UpstreamBillingProbeExtraKey,
+		service.GrokMediaEligibleExtraKey,
+		"grok_billing_snapshot",
 	}
 	filtered := make(map[string]any)
 	for _, key := range keys {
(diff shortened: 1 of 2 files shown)


Yangshengyi035717:
希望推送下更新

PR #4506, opened 2026-07-17
fix(deploy): make the redis command flags take effect

coo1white:
In `deploy/docker-compose.yml`, the redis `command` is one quoted script given to the inner `sh -c`. Docker compose keeps the newlines inside the quoted string, so `redis-server` on the first line runs as a complete command with no flags at all — the `--save` / `--appendonly` / `--appendfsync` lines after it are silently never applied.

You can see it on any deploy of this file: `redis-cli CONFIG GET appendonly` says "no", though `--appendonly yes` is right there in the command.

The fix is trailing `\` line continuations, which fold the script back into one command. Checked with redis:7-alpine: `appendonly` is now "yes" and `save` is "60 1".

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -241,12 +241,16 @@ services:
         hard: 100000
     volumes:
       - redis_data:/data
+    # The command is one quoted script for the inner `sh -c`. Compose keeps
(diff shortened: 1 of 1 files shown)


coo1white:

PR #4507, opened 2026-07-17
fix(docker): cross-compile the image instead of running Go under QEMU

coo1white:
Building the image for another arch (e.g. `--platform linux/amd64` on an Apple-Silicon Mac) runs the whole Node and Go toolchain under QEMU emulation, because the builder stages have no `--platform`. That is slow, and the emulated networking kept dropping `go mod download` with `unexpected EOF` on big module zips.

Two changes, both build-time only:

1. Pin `frontend-builder` and `backend-builder` to `$BUILDPLATFORM` and cross-compile: add `TARGETOS` / `TARGETARCH` args and set them on `go build`. The binary is `CGO_ENABLED=0`, so this is a clean pure-Go cross-compile with no emulation on the heavy stages. The frontend output is JS (arch-neutral), so building it on the host arch is safe too.

2. Add go module and build cache mounts, so a retry after a network blip goes on from where it stopped instead of re-fetching every zip.

The runtime image and app behavior are unchanged. Same-arch builds are unchanged too — `BUILDPLATFORM` equals the target there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,9 @@ ARG NPM_CONFIG_REGISTRY=
 # -----------------------------------------------------------------------------
 # Stage 1: Frontend Builder
 # -----------------------------------------------------------------------------
-FROM ${NODE_IMAGE} AS frontend-builder
+# --platform=$BUILDPLATFORM: the frontend output is JS (arch-neutral), so build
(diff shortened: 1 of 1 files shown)


coo1white:

PR #4537, opened 2026-07-18
fix(monitor): refresh Anthropic text-block extraction

heathermhuang:
## Summary
- extract Anthropic monitor text from all `content[]` text blocks instead of assuming `content[0]` is text
- preserve the newer OpenAI Responses reasoning-aware extractor already on `main`
- cover thinking-before-text, single-text, thinking-only, and multi-text responses

## Why
Anthropic Messages responses may place a `thinking` block before the answer. The monitor treated those successful responses as empty, producing challenge mismatches or replace-mode empty-text failures.

## Attribution and supersession
This refresh adopts #1953 on current `main`. The original commit author, @mark-ly-wang, is preserved in commit `1a855d3e`, including the cherry-pick source reference to `76cf1ee`.

#1953 is now stale and conflicts with the later OpenAI Responses monitor work. This PR resolves that conflict without changing the original Anthropic behavior or taking ownership of the other #4527 requests.

Addresses #4527 item 5 only.

## Test coverage
- AI-assessed changed-path coverage: 88% (14/16 paths)
- focused regression tests cover Anthropic thinking-before-text extraction and challenge validation
- existing tests cover empty-response failure and OpenAI Responses reasoning-before-text coexistence

## Pre-landing review
No issues found. Backend, security, provider behavior, trust boundaries, regression risk, and scope drift were reviewed against current `main`.

## Test plan
- [x] `go test -tags=unit ./internal/service -run 'TestExtractAnthropicMonitorText|TestValidateChallenge_AnthropicTextAfterThinking|TestRunCheckForModel' -count=1`

### DIFF
--- a/backend/internal/service/channel_monitor_checker.go
+++ b/backend/internal/service/channel_monitor_checker.go
@@ -149,11 +149,11 @@ func pingEndpointOrigin(ctx context.Context, endpoint string) *int {
 	return &ms
 }
 
-// providerAdapter 描述某个 provider 在 challenge 检测中需要的 4 件事：
+// providerAdapter 描述某个 provider 在 challenge 检测中需要的几件事：
 //   - 拼出请求路径（含 model 占位）
 //   - 序列化请求体
 //   - 构造鉴权头
-//   - 从响应 JSON 中按 path 提取文本（gjson path）
+//   - 从响应 JSON 中提取文本（默认按 gjson path；需要时可自定义）
 //
 // 加新 provider 只需要在 providerAdapters 里增加一个条目，无需触碰 callProvider / validateProvider。
 type providerAdapter struct {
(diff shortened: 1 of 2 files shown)


heathermhuang:
Live pre-review verification completed on the exact PR head `1a855d3e` as part of the combined current-main canary.

- Focused Anthropic monitor regressions pass, including `thinking` arriving before the first text block.
- Combined backend service/repository/handler/routes tests and vet passed.
- The live OAuth and API-key lanes each passed text, image generation, video generation, protected video download, and ownership isolation.
- Public browser smoke returned 200 with no console errors.
- The canary remained healthy with zero restarts and clean fatal/panic/migration logs; disposable keys were deleted and returned 401 afterward.

Production had no scheduled channel-monitor tasks available, so I did not manufacture a live monitor failure on shared accounts. The exact thinking-before-text behavior is covered by the focused regression; the live matrix proves no gateway regression.

Issue #4525, opened 2026-07-18
grok账号授权授权forbidden


Yangshengyi035717:
sub提示授权成功“

<img width="327" height="74" alt="Image" src="https://github.com/user-attachments/assets/f077243b-30f2-4e69-86fb-ec2aff9e3fe0" />

xai端也还在：

<img width="729" height="475" alt="Image" src="https://github.com/user-attachments/assets/cd689d18-f022-4c94-a0d6-796c473f60b2" />

账号还是403：

<img width="555" height="218" alt="Image" src="https://github.com/user-attachments/assets/c6817615-fddd-4513-a053-bfdd4a4b0a72" />

Yangshengyi035717:
<img width="497" height="604" alt="Image" src="https://github.com/user-attachments/assets/2dcb3347-4af0-4737-9da9-4e4fb632e343" />

PR #4540, opened 2026-07-18
fix(grok): retry CLI chat permission denial

heathermhuang:
## Summary

- recognize the structured Grok CLI 403 reported in #4525 as the newer compatibility-denial wording
- reuse the existing authenticated, replayable fallback from the CLI proxy to the official xAI API
- keep unrelated entitlement denials on the original CLI response

## Root cause

The account-test path already routes OAuth traffic to the CLI subscription endpoint and the shared transport already supplies the required CLI identity headers. The existing compatibility fallback, however, only recognized the older contiguous phrase `access denied`.

The reported response uses `code: permission_denied` with `Access to the chat endpoint is denied...`, so the fallback was never attempted even though this is the same compatibility boundary.

This change parses that structured response and requires the observed code and message prefix. It does not broaden fallback to generic permission or subscription failures. The existing transport still retries only authenticated, replayable CLI-proxy 403 requests and only substitutes the official API response when that response is successful.

## Version diagnosis

No CLI version bump is included. The current pinned version may be overridden by operators, but the issue does not prove that version drift caused this permission denial.

## Validation

- regression reproduces the exact #4525 response wording and confirms the official API retry
- entitlement and near-miss permission denials do not retry
- non-CLI, unauthenticated, non-403, and non-replayable requests cannot trigger credential replay
- `go test -tags=unit ./internal/repository -count=1`
- `git diff --check`
- independent changed-path review found no actionable issues

Live success with the reporter's credential remains unverified because no affected OAuth credential was available to this change.

Addresses #4525

### DIFF
--- a/backend/internal/repository/http_upstream.go
+++ b/backend/internal/repository/http_upstream.go
@@ -7,6 +7,7 @@ import (
 	"compress/gzip"
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -323,7 +324,7 @@ func (t *grokAccessDeniedFallbackTransport) RoundTrip(req *http.Request) (*http.
 	}
 
 	body, ok := bufferSmallResponseBody(resp, grokFallbackBodyLimit)
-	if !ok || !bytes.Contains(bytes.ToLower(body), []byte("access denied")) {
+	if !ok || !isGrokCLICompatibilityAccessDenied(body) {
 		return resp, nil
 	}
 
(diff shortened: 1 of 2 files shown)


heathermhuang:
Live pre-review verification completed on the exact PR head `2946281a` as part of the combined current-main canary.

- Focused repository tests cover the permission-denied fallback, one-retry limit, header behavior, and non-403 exclusions.
- Combined backend service/repository/handler/routes tests and vet passed.
- The live OAuth lane passed text, image generation, video generation, protected video download, and usage attribution; the API-key lane passed the same no-regression matrix.
- Public browser smoke returned 200 with no console errors; the canary stayed healthy with zero restarts, and disposable keys were deleted and returned 401.

The shared OAuth account did not naturally reproduce the upstream 403 during this canary, so I did not claim a live fallback invocation or mutate account entitlement. The exact 403 branch is covered by focused tests; live proof covers the normal OAuth path and integration safety.

PR #4497, opened 2026-07-17
fix(grok): fail closed for ineligible OAuth media

heathermhuang:
## Summary

This is the follow-up to #4474 and #4484 for the OAuth media reports in #4420 and #4471.

- Require positive paid-entitlement evidence before a Grok OAuth account can receive new image or video generation traffic.
- Probe unobserved OAuth accounts on the request path before forwarding; newly imported accounts run the billing-first quota probe proactively.
- Keep API-key media, Grok chat, and existing video-status lookups unchanged.
- Reject HTTP 200 image responses that contain no actual image outputs and fail over instead of counting the requested `n` as generated output.

## Root cause

The previous quarantine only blocked explicit overrides and billing `403` observations. Missing, malformed, Free-tier, and inconclusive OAuth billing states still reached media forwarding. Separately, an empty `{"data":[]}` image response was treated as success and billed as one image by falling back to the request count.

## Behavior and compatibility

- OAuth media now fails closed for Free, forbidden, missing, malformed, or inconclusive billing state.
- Positive paid evidence or `extra.grok_media_eligible=true` enables OAuth media.
- `extra.grok_media_eligible=false` continues to quarantine an account.
- If no eligible account remains, generation returns `503 grok_media_no_eligible_account`.
- This does **not** claim to score semantic prompt adherence from image pixels. It removes cohorts that lack positive media entitlement and rejects structurally empty success responses.

## Production E2E (2026-07-17)

The exact PR head `e8606315` was integrated into a private production build and exercised through separate disposable Grok OAuth and API-key groups:

- OAuth text: HTTP 200, attributed to the OAuth account.
- API-key text: HTTP 200, attributed to the API-key account.
- OAuth image: HTTP 200 with one actual image, but only after the new request-path billing probe persisted complete weekly/monthly HTTP 200 observations plus authoritative paid-quota evidence.
- API-key image: HTTP 200 with one actual image, preserving the non-OAuth path.
- All four usage rows had the expected group, account type, model, request type, and billing mode.
- The app remained healthy with zero restarts and no fatal/panic log matches.
- Both disposable keys were audit-soft-deleted, cache-invalidated, and verified to return HTTP 401.

Interpretation: the OAuth image success above is expected because that production account has positive paid-entitlement evidence. Free, forbidden, missing, malformed, and inconclusive OAuth states are the rejection cases covered by this change; image-content quality remains upstream behavior.

## Verification

- `go test -tags=unit ./...`
[...]

PR #4543, opened 2026-07-18
fix(grok): securely proxy protected video content

heathermhuang:
## Summary

- adopt #4498's protected Grok video-content proxy on current `main`, preserving the original author's commit and attribution
- return only same-origin relative download URLs; proxy headers and request host cannot choose the returned origin
- bind each video request to the originating user and API key, require that owner binding before lookup, and fail closed unless the scheduler returns the exact originally bound upstream account
- preserve Range requests, `206` streaming, safe response headers, and `416 Content-Range` semantics

## Why this supersedes #4498

#4498 correctly identifies that protected video content must be fetched with the task's upstream credential, but its branch is not writable here and the original absolute URL builder trusted forwarded host/protocol headers. That could make a status response advertise an external origin.

This branch replays #4498's single commit with attribution, then applies the smallest security corrections on current `main`:

- relative `/v1/videos/.../content` or `/videos/.../content` URLs only
- no forwarded-host/protocol trust
- owner-scoped binding stored through a strict primary cache key, without sticky-session legacy dual-write or fallback
- sanitized `404` for missing, foreign, stale, or mismatched bindings
- no account failover for video lookup requests

## Coordination with #4497

This branch intentionally does not bundle #4497. The branches both edit `backend/internal/service/grok_media.go`; whichever lands second must rebase and retain both the protected-content status rewrite and #4497's empty-image-response failover check.

## Validation

- protected content uses the bound upstream account credential
- same-group alternate user/API-key ownership checks fail closed, including with a legacy session hash attached
- hostile forwarded headers cannot influence returned download URLs
- Range/`206`, full `200`, and `416` content responses preserve required semantics
- only allowlisted content headers are forwarded; upstream cookies/private headers are not
- matching nested/escaped request IDs rewrite while other request IDs and direct media URLs remain unchanged
- `go test -tags=unit ./internal/service ./internal/handler ./internal/server/routes -count=1`
- `git diff --check`
- independent security review found no remaining actionable issues; coverage audit reached 81% of material paths

Live protected-content verification remains unavailable without an affected upstream account.

Supersedes #4498


heathermhuang:
Live pre-review verification completed on the exact PR head `4c845ca6` as part of the combined current-main canary.

- Focused tests prove signed video URLs are restricted to the validated HTTPS provider origin, redirects are disabled, and neither bearer credentials nor account header overrides are sent to the media host. Deceptive origins and link-local targets are rejected before fetch.
- Both OAuth and API-key videos completed live. Status responses returned the relative protected URL and did not reflect hostile forwarded host/protocol headers.
- Both protected downloads returned 206 with exactly 1024 bytes and an MP4 `ftyp` marker.
- Invalid ranges returned 416. The other disposable key received 404 for both status and content, proving ownership isolation.
- Combined backend service/repository/handler/routes tests and vet passed. Public browser smoke returned 200 with no console errors.
- The canary stayed healthy with zero restarts and clean fatal/panic/migration logs; disposable keys were deleted and returned 401 afterward.

PR #4545, opened 2026-07-18
feat(openai): add durable quota reset tasks

GhostFlying:
## Summary

- add a durable, generic `background_task_runs` engine with PostgreSQL claiming, leases, claim-version fencing, retries, cancellation, audit state, and runner health reporting
- register an `openai_quota_reset` handler that schedules a one-shot reset for the exact nearest-expiring credit
- add account-row scheduling controls and an Ops task-instance view with cancel and indeterminate retry actions

## Why

OpenAI quota resets were previously manual and the reset request identity only lived for the duration of one call. A timer or ordinary retry loop would therefore be unsafe across timeouts, worker lease loss, or service restarts: the HTTP request may have reached the upstream even when the caller did not receive a response.

This change uses at-least-once HTTP delivery while preserving exactly-once logical consumption. Every task persists one immutable `redeem_request_id + credit_id` pair before the first dispatch. All automatic retries, lease reclaims, restarts, and manual retries reuse that pair and never fall back to an unscoped reset request.

## Backend

- add migration `183_background_task_runs.sql` with durable task state, creation-request aliases, active dedupe locking, dispatch metadata, leases, and fenced claims
- claim due work with `FOR UPDATE SKIP LOCKED`; use a two-minute lease, periodic renewal, and `claim_version` checks on every state transition
- persist dispatch intent before sending the upstream POST; cancellation is only allowed while `first_dispatch_at` is still null
- retry timeout, disconnect, 408/429/5xx, truncated, and unknown responses with the same request body and bounded backoff
- treat `reset` and `already_redeemed` as success; preserve ambiguous exhausted tasks as `indeterminate` and only allow same-key manual retry
- retain private credit IDs internally while keeping existing quota responses and task DTOs redacted
- expose admin create/list/cancel/retry APIs and report the runner through the existing Ops heartbeat surface

## Frontend

- place a schedule action between quota count and manual reset for supported OpenAI OAuth accounts
- require quota data to be loaded before scheduling and confirm the exact expiry, lead time, and resulting execution time
- support 10, 30, and 60 minute lead times, with 60 minutes as the default and an explicit immediate-execution state
- show existing task state from the account row, including pre-dispatch cancellation and same-task indeterminate retry
- add Runner Status and Task Instances tabs to the Ops background-task detail view

## Scope

- one-shot OpenAI quota reset tasks only; no cron rules, periodic reconciliation, or migration of existing background jobs
- OpenAI OAuth non-shadow accounts only; task creation fails closed when an exact credit ID cannot be established
- terminal task history is retained for audit; no retention cleanup is added in this change

## Validation

- `make test-unit`
- `make test-integration`
- `make test-frontend`
- `golangci-lint run ./... --timeout=30m`
- `govulncheck ./...`
- frontend production build
- frontend production audit exception validation

Remote validation for this exact PR head:

- [Fork CI](https://github.com/GhostFlying/sub2api/actions/runs/29635446004)
- [Fork Security Scan](https://github.com/GhostFlying/sub2api/actions/runs/29635446020)

The fault-injection coverage includes an upstream that consumes the first POST but delays its response past the client timeout. The second POST uses the identical request body and receives `already_redeemed`; the test asserts two HTTP calls, one unique request ID, one unique credit ID, and one actual credit consumption.


GhostFlying:
验证了下看起来没啥问题

telegram-cloud-photo-size-1-5109598102884125777-y

PR #4548, opened 2026-07-18
feat(auth): show "Last time used" badge on the last OAuth provider

broven:
## What

Adds a **"Last time used"** badge to the login page, marking the OAuth provider the visitor most recently completed a login with — much like Google's account chooser. When several OAuth providers are configured, only the one the user last used gets the badge, so returning users can pick the right button at a glance.

Password / 2FA logins intentionally get **no** badge.

## Why it's browser-local (not server-side)

The login page is unauthenticated, so the server can't tell the page which provider *this* visitor used last. The hint is therefore stored **per browser/device** in local storage (same model as Google's "Last used"). New device / cleared storage → no badge.

## How

- **`utils/lastUsedOAuth.ts`** — the whole mechanism in one place:
  - `setPendingOAuthProvider(id)` — stashed in `sessionStorage` at click, before the redirect.
  - `promotePendingOAuthProvider()` — on a successful OAuth login, promote the pending provider into a persisted "last used" record (`{ provider, ts }`). No-op when nothing is pending, so it never clobbers an existing value (e.g. email-verification links).
  - `clearLastUsedOAuthProvider()` — on password/2FA/register login, so the badge never shows stale info.
  - `getLastUsedOAuthProvider()` — read for display; **expires after 90 days** and prunes malformed/expired records, so a shared/public browser doesn't leak a previous visitor's provider indefinitely.
- **`stores/auth.ts`** — two one-line hooks at the existing choke points: promote in `setToken` (all OAuth/SSO callbacks funnel here), clear in `setAuthFromResponse` (password/2FA/register funnel here).
- **OAuth sections** (GitHub/Google, Linux.do, DingTalk, WeChat, OIDC) — each records its provider on click and renders a small `LastUsedBadge` in the button's top-right corner when it matches. `LoginView` reads the hint once and passes it down.
- **i18n** — `auth.lastTimeUsed` added to `en` and `zh`.

## Tests

- `utils/__tests__/lastUsedOAuth.spec.ts` — promote / no-op-without-pending / clear / 90-day expiry / invalid & malformed handling.
- `components/auth/__tests__/lastUsedBadge.render.spec.ts` — badge renders only on the matching provider button, and not at all when the last-used provider isn't shown.

Full frontend suite green (1209 tests), `typecheck`, `lint`, and `build` all pass.

## Notes

The change surface is intentionally small and localized (one util + two auth-store lines + a badge per provider button) to stay easy to review and to keep future merges clean.

---

## Screenshots

Real login page in local dev. OAuth providers are off in a fresh dev seed, so they were toggled on just for the demo — the **badge itself and the `localStorage` last-used logic are the real code path** (the hint is seeded exactly as a successful OAuth login writes it).

**Last login = GitHub** — badge on the GitHub button:

Last time used — GitHub

**Last login = Linux.do** — the badge moves to the Linux.do button:

Last time used — Linux.do

**Password login / no history** — no badge shown:

No badge

### DIFF
--- a/frontend/src/components/auth/DingTalkOAuthSection.vue
+++ b/frontend/src/components/auth/DingTalkOAuthSection.vue
@@ -1,6 +1,7 @@
 <template>
   <div class="space-y-4">
-    <button type="button" :disabled="disabled" class="btn btn-secondary w-full" @click="startLogin">
+    <button type="button" :disabled="disabled" class="btn btn-secondary relative w-full" @click="startLogin">
+      <LastUsedBadge v-if="lastUsedProvider === 'dingtalk'" />
       <svg
         class="icon mr-2"
         viewBox="0 0 24 24"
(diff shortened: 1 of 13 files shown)


jadelike-wine:

broven:

PR #4589, opened 2026-07-19
fix(apicompat): emit null stop_reason on Anthropic message_start

huangjunsen0406:
## Summary

- Change `AnthropicResponse.StopReason` from `string` to `*string` so streaming `message_start` serializes as JSON `null` instead of `""`.
- Match official Anthropic Messages SSE wire format and avoid strict clients (e.g. Claude Code) failing to finalize turns / persist session usage.
- Keep final non-stream responses and `message_delta` stop reasons as explicit non-null strings via `AnthropicStopReasonPtr`.
- Add regression test `TestMessageStartSSE_StopReasonIsJSONNull`.

## Problem

Streaming `message_start` currently embeds:

```json
"stop_reason": ""
```

Official Anthropic uses:

```json
"stop_reason": null
```

This repository already notes that empty `stop_reason` is invalid for strict Anthropic clients in the Chat Completions bridge path, but the Responses→Anthropic stream path still emitted `""` from a zero-value `string` field.

In practice this can prevent Claude Code from writing non-zero usage / stop reasons into session logs (CC Switch `session_log` then under-records), even when Sub2API usage logs and the wire `message_delta.usage` are correct.

## Test plan

- [x] `go test ./internal/pkg/apicompat/ -count=1`
- [x] `go build ./internal/service/ ./internal/handler/`
- [x] New test asserts SSE contains `"stop_reason":null` and not `""`
- [ ] Live: Claude Code multi-turn + tools through `/v1/messages` → session rows keep non-zero input/output tokens
- [ ] Live: confirm Sub2API usage logs unchanged

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_responses_test.go
+++ b/backend/internal/pkg/apicompat/anthropic_responses_test.go
@@ -208,7 +208,7 @@ func TestResponsesToAnthropic_TextOnly(t *testing.T) {
 	anth := ResponsesToAnthropic(resp, "claude-opus-4-6")
 	assert.Equal(t, "resp_123", anth.ID)
 	assert.Equal(t, "claude-opus-4-6", anth.Model)
-	assert.Equal(t, "end_turn", anth.StopReason)
+	assert.Equal(t, "end_turn", AnthropicStopReasonString(anth.StopReason))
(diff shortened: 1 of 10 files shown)


huangjunsen0406:

huangjunsen0406:
recheck

PR #4594, opened 2026-07-19
fix(update): support zip archive extraction on Windows

QinIndexCode:
GoReleaser packages Windows releases as .zip files, but the updater only handled .tar.gz / .tar archives. On Windows, this caused the downloaded .zip file to be copied directly over the binary path, resulting in a broken executable.

Changes:
- Add extractBinaryFromZip to extract sub2api(.exe) from .zip archives
- Reuse existing security checks: path traversal detection, size limit
- Add unit tests including real GitHub release download validation

Fixes Windows one-click update failures.

### DIFF
--- a/backend/internal/service/update_service.go
+++ b/backend/internal/service/update_service.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"archive/tar"
+	"archive/zip"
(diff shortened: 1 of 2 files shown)


QinIndexCode:

PR #4595, opened 2026-07-19
fix(openai): make HTTP bridge failover safe

kinsolee:
## Summary

Fix OpenAI Responses WebSocket `http_bridge` error handling so first-turn upstream failures can use the gateway's existing account failover path without exposing a transient error to Codex clients.

This covers:

- transport failures from `httpUpstream.Do` (EOF, proxy, DNS, TCP, TLS)
- failover-eligible HTTP statuses such as 401, 429 and 5xx
- first-event SSE `type:error` responses
- SSE read failures before any downstream event
- streams that send `[DONE]` without a Responses terminal event

## Root cause

`proxyOpenAIWSHTTPBridgeTurn` wrote an error event to the client and returned a plain `error`. The outer WebSocket handler only switches accounts for `*UpstreamFailoverError`, so the connection closed before `response.completed` even when another account was available.

The bridge also treated `[DONE]` plus any event as success even when neither `response.completed` nor `response.done` was observed.

## Safety invariant

Cross-account replay is allowed only when:

1. this is turn 1; and
2. no downstream semantic event has been written.

Turn 2+ errors deliberately do not return `UpstreamFailoverError`, because the outer handler stores the connection's first message. Replaying it would duplicate a completed first turn and could repeat tool calls.

Grok 429/5xx handling still persists account rate-limit state before first-turn failover.

## Tests

Added table-driven coverage for:

- turn-1 transport failover with zero client writes
- turn-2 transport error without replay
- turn-1 HTTP 401 / 429 / 500 failover
- turn-2 HTTP 500 without replay
- turn-1 SSE rate-limit error failover
- turn-2 SSE error without replay
- `[DONE]` without terminal event
- `response.created` followed by `[DONE]` being treated as truncated
- normal SSE-to-WebSocket relay regression
- Grok 429 rate-limit persistence regression

Validation performed:

- `go test ./internal/service -count=1`
- `go test -tags=unit ./internal/service -count=1`
- `go test -tags=integration ./internal/service -count=1`
- `golangci-lint v2.9.0 run --timeout=30m` (`0 issues`)
- `make build`
- full Docker image build
- production-compatible WebSocket smoke tests: 20/20 sessions, 40/40 turns; two real first-turn 502s transparently failed over

## Notes

The repository's full `golangci-lint run ./...` currently reports pre-existing findings outside the changed files (SSRF/path traversal taint warnings, one govet warning and one deprecated API warning). No new lint finding was observed in these two files.

PR #4622, opened 2026-07-19
fix(openai): resume later WebSocket turns after 429

kinsolee:
## Dependency

This is a stacked follow-up to #4595. Please review/merge #4595 first. Once #4595 lands, this PR will contain only the later-turn 429 resume commit.

## Summary

Allow an OpenAI Responses client WebSocket to stay open when an `http_bridge` account reaches its usage limit on a later turn. The gateway excludes the limited account, selects a replacement account, and retries only the current turn with a self-contained payload instead of replaying the connection's original first message.

## Safety boundary

Later-turn account migration is deliberately limited to:

- upstream status `429`;
- no downstream semantic event has been written;
- the retry payload can reconstruct the complete current-turn context;
- every `function_call_output` has matching tool-call context.

Transport errors, 5xx responses, truncated streams, and any failure after downstream output remain fail-closed and are not replayed.

## Context reconstruction

The retry payload:

- accumulates user input and prior assistant `response.output` items;
- preserves function/tool calls and outputs without duplication;
- removes the old account's `previous_response_id`;
- restores the client-facing original model;
- is carried by a private wrapper error and exposed through a cloning accessor.

## Verification

- `make test`
- `make test-unit`
- `make test-integration`
- `make build`
- golangci-lint v2.9.0: `0 issues`
- two-account WebSocket regression: account A succeeds on turn 1, returns 429 on turn 2, account B completes turn 2 on the same client WebSocket
- partial downstream output regression: no failover/replay
- orphan tool-output regression: fail-close
- independent final review: no P0/P1/P2 findings


kinsolee:

kinsolee:
Follow-up fixes for the latest CI feedback:

- Upgraded frontend `postcss` from 8.5.6 to 8.5.12, which resolves GHSA-6g55-p6wh-862q. No audit exception was added.
- Updated rollback API tests to assert the existing 15-minute timeout option.
- Verified frozen-lockfile install, frontend audit exception validation, 187 frontend test files / 1290 tests, and frontend build.

All PR checks are now green; mergeable state is clean.

Issue #1348, opened 2026-03-27
账号状态过载中


zhou8704:
账号管理，状态显示过载中，剩余10分钟重置，恢复后发了1条消息又显示负重过载中了，查看日志 显示status_code": 503。完整日志：
Mar 27 20:05:46 iZt4n0px97w3nlzsbw7znsZ sub2api[298875]: 2026-03-27T20:05:46.582+0800        INFO        middleware/logger.go:60        http request completed        {"service": "sub2api", "env": "production", "component": "http", "request_id": "565b3e30-3e46-4ea8-8a82-3b746309d177", "client_request_id": "", "path": "/v1/messages", "method": "POST", "client_request_id": "12b7dc2e-9802-407a-8592-4ac6d7a34c54", "component": "http.access", "status_code": 503, "latency_ms": 9004, "client_ip": "223.73.115.31", "protocol": "HTTP/1.1", "method": "POST", "path": "/v1/messages", "model": "claude-opus-4-6", "completed_at": "2026-03-27T20:05:46.582+0800"}

账号类型：Anthropic   claude max 订阅类型。
在服务器上的claude测试是正常的。

PR #4634, opened 2026-07-20
feat(gateway): add bounded pre-stream Anthropic OAuth 529 retries

joeeeeey:
## Problem

When Anthropic OAuth/setup-token forwarding receives a pre-stream HTTP 529 (`overloaded_error`), Sub2API immediately applies the configured cooldown and moves to account failover. That conservative behavior is appropriate as the default, but it provides no way to absorb a brief upstream overload:

- a single-account deployment returns a downstream 503 immediately;
- a pooled deployment rotates away from an otherwise healthy account for a transient response.

Issue #1348 includes several overload reports, including an explicit HTTP 529 report. This change addresses only that transient 529 recovery gap. It does not claim to address the account-restriction or ban reports discussed in the same issue.

## Why bounded and opt-in

Adding 529 to the existing generic retry predicate would allow up to five attempts and could amplify load while Anthropic is already overloaded. It would also bypass the existing 529 cooldown unless retry exhaustion were handled explicitly.

This PR therefore adds a separate operator-controlled budget:

- `0` retries remains the default, preserving current behavior;
- operators may enable at most `1` or `2` same-account retries;
- retries use the existing exponential backoff and 10-second elapsed-time budget;
- only responses received before streaming begins are replayed;
- exhaustion still applies the existing cooldown and failover behavior.

Keeping this policy upstream makes the behavior reusable and reviewable without requiring deployments to carry a long-lived gateway patch.

## Changes

- Add `oauth_retry_count` to the existing 529 overload settings, constrained to `0-2` and defaulting to `0`.
- Apply the setting only to pre-stream HTTP 529 responses from Anthropic OAuth/setup-token accounts.
- Load the setting lazily on the first matching 529, so successful requests do not add a settings lookup.
- Rebuild and replay the request through the existing retry loop.
- Apply the existing 529 account-state handling after the configured retries are exhausted, then return the normal failover error.
- Expose the setting in the admin UI with explicit scope and compatibility wording in English and Chinese.

| `oauth_retry_count` | Total attempts after consecutive 529 responses |
| --- | --- |
| `0` (default) | 1 |
| `1` | 2 |
| `2` | 3 |

## Compatibility and scope

- Existing stored JSON and older API clients omit the new field and retain the current behavior (`0` retries).
- API-key accounts, OAuth 403 handling, non-529 status handling, and post-stream/SSE errors are unchanged.
- No database migration is required.

## Tests

- [x] Default `0` performs no same-account retry.
- [x] `529 -> 529 -> 200` succeeds with `oauth_retry_count=2` and replays the same request body.
- [x] A third consecutive 529 applies the configured cooldown and returns the existing failover error.
- [x] API-key and non-Anthropic OAuth accounts remain outside the new policy.
- [x] Settings defaults, range validation, JSON compatibility, admin UI load/save, and i18n are covered.

Verification run:

```text
env -u OPENAI_API_KEY go test -tags unit ./... -count=1
go vet ./internal/service ./internal/handler
pnpm exec vitest run src/views/admin/__tests__/SettingsView.spec.ts
pnpm typecheck
pnpm exec eslint src/views/admin/SettingsView.vue src/views/admin/__tests__/SettingsView.spec.ts src/api/admin/settings.ts src/i18n/locales/en/admin/settings.ts src/i18n/locales/zh/admin/settings.ts
```

The integration suite was not run locally because its PostgreSQL/Redis test environment was not provisioned.

## Risk

Retrying an overloaded account can increase upstream pressure. The feature is therefore disabled by default, capped at two retries, restricted to pre-stream responses, and followed by the existing cooldown/failover path when retries are exhausted.

### DIFF
--- a/backend/internal/handler/admin/setting_handler_runtime.go
+++ b/backend/internal/handler/admin/setting_handler_runtime.go
@@ -50,7 +50,7 @@ func (h *SettingHandler) DeleteAdminAPIKey(c *gin.Context) {
 	response.Success(c, gin.H{"message": "Admin API key deleted"})
 }
 
-// GetOverloadCooldownSettings 获取529过载冷却配置
+// GetOverloadCooldownSettings 获取529过载处理配置
 // GET /api/v1/admin/settings/overload-cooldown
 func (h *SettingHandler) GetOverloadCooldownSettings(c *gin.Context) {
 	settings, err := h.settingService.GetOverloadCooldownSettings(c.Request.Context())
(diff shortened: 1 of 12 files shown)


joeeeeey:

PR #4666, opened 2026-07-20
fix(config): support Redis ACL username

TTopoo:
## Summary

- Add optional Redis ACL username configuration through `redis.username` / `REDIS_USERNAME`.
- Pass the username into go-redis options for the normal runtime Redis client and setup-time Redis connection tests.
- Include the field in auto setup, the web setup wizard, i18n labels, and deployment examples.
- Keep the username empty by default so existing Redis deployments using the default user continue to behave the same way.

## Why

Some managed Redis services and ACL-enabled Redis deployments require authentication with both a username and a password. Sub2API already exposed Redis host, port, password, database, TLS, and pool settings, but there was no way to provide the ACL username. In those environments, the app could not authenticate even when the password was correct.

This change only forwards the username when it is configured. Leaving it empty preserves the current password-only/default-user behavior supported by go-redis.

## Test plan

- [x] `git diff --check`
- [x] `go test ./internal/config -run TestLoadRedisUsernameFromEnvironment -count=1`
- [x] `go test ./internal/repository -run TestBuildRedisOptions -count=1`
- [x] `go test ./internal/setup -run 'TestWriteConfigFileIncludesRedisUsername|TestWriteConfigFileKeepsDefaultUserConcurrency' -count=1`
- [x] `docker build -t sub2api-acl:0.1.162-redis-username-r1 .`

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,7 @@ database:
 redis:
   host: "localhost"
   port: 6379
+  username: ""
   password: ""
 
 jwt:
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -1403,6 +1403,7 @@ func (d *DatabaseConfig) DSNWithTimezone(tz string) string {
 type RedisConfig struct {
 	Host     string `mapstructure:"host"`
 	Port     int    `mapstructure:"port"`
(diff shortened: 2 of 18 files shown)


TTopoo:

PR #4743, opened 2026-07-22
feat: add Alibaba Bailian (DashScope) platform — wan/happyhorse video generation + qwen text relay

SipengXie2024:
## Summary

Adds **bailian** (Alibaba Bailian / DashScope) as a sixth platform: async video generation for the **wan (万相)** and **happyhorse (欢乐马)** model families, plus qwen text relay through DashScope's OpenAI-compatible endpoint. Pure API-key platform — no OAuth machinery.

## What's included

**Video generation (async DashScope tasks)**
- `POST /v1/videos/generations` accepts the same flat OpenAI-style body as the existing Grok video endpoints (`model/prompt/negative_prompt/image/ratio/resolution/duration/seed/watermark`) and converts it into DashScope's nested `{model, input, parameters}` shape, creating the task with `X-DashScope-Async: enable`.
- `GET /v1/videos/{task_id}` polls the task and passes the DashScope response through verbatim. `output.video_url` is a signed public OSS link (24h) that clients download directly — no content proxy, no gateway bandwidth cost. `videos/edits|extensions|content` stay Grok-only (404 for bailian groups).
- Task ownership is pinned in Redis to the creating account with a **24h TTL** matching the DashScope task validity window (the 1h sticky-session TTL used by Grok would 404 overnight polls). The binding key includes user + API key IDs to prevent cross-tenant task_id probing.

**Billing**
- Charged at task-creation time: normalized `duration × per-second resolution-tier price`. The normalized duration/resolution are **always written into the upstream request body**, so the billed amount and the generated amount cannot diverge (no default-duration arbitrage).
- Generalizes the video usage gate (`isVideoUsageResult`) so `wan*`/`happyhorse*` reach the per-second billing path — previously only `grok-imagine-video*` matched and anything else silently fell back to token billing (zero cost).
- Reuses the group video price columns from migration 170; adds Bailian per-second defaults (flagged in-code as needing verification against the Bailian console; group prices override).

**Text (qwen)**
- Bailian accounts are OpenAI-compatible API-key accounts. `chat/completions`, `/v1/messages` and `/v1/responses` all dispatch onto the raw chat-completions bridge (DashScope has no `/v1/responses`) targeting `{root}/compatible-mode/v1/chat/completions`.
- Account connectivity test probes the same compatible-mode endpoint (default model `qwen-flash`).
- A single stored `base_url` serves both APIs: pasted `/compatible-mode(/v1)` or `/api/v1` suffixes are normalized away, so workspace-scoped domains (`https://{ws}.cn-beijing.maas.aliyuncs.com`) work as-is.

**Platform registration + admin UI**
- Platform constant, quota allowlist + ent schema validation + migration 186 (CHECK superset, re-entrant), simple-mode default group (with `allow_image_generation` enabled — the video route reuses that gate), scheduler snapshot buckets (`[5]string → [6]string`), error-passthrough platforms, LiteLLM provider mapping (`dashscope`), group handler `oneof` validators.
- Frontend: platform button in account create/edit (API-key only), model whitelist catalog, platform quota views, group badges/colors, zh («阿里百炼») / en i18n.

## Not changed

- Zero Grok behavior changes: bailian gets its own handler/service (`bailian_video.go`, ~400 lines service + handler) modeled on the grok_media structure instead of widening `handleGrokMedia`. Shared primitives (failover error, passthrough rules, HTTP upstream, usage pipeline) are reused as-is.

## Tests

- Table-driven unit tests: flat→DashScope conversion (defaults/clamping/optional fields), API-root normalization (pasted suffixes, workspace domains, path-escaped task IDs), queue-stub forwarding (async header, task_id extraction, verbatim polling passthrough, 400-passthrough vs 401/403/429/5xx failover, lookups never fail over), billing gates and default prices, session-hash ownership.
- Existing platform-count assertions updated for the sixth platform (scheduler snapshot buckets 12→14 per group, quota contract, platform quota admin).
- `make test-unit` green except two failures that also fail on a pristine upstream main checkout (`TestEstimateOpenAIInputTokens_CompareWithOpenAIAPI`, `admin.system.rollback.spec.ts`); `vue-tsc`, `vitest`, `vite build` green; new files lint-clean under golangci-lint.

## Known limitations / follow-ups

- Task binding lives in Redis only: a flush loses it and later polls return 404 (documented behavior; a `video_tasks` table could harden this later).
- Creation-time billing: an upstream FAILED task is not auto-refunded.
- Channel monitor support for bailian (adapter + provider CHECK migration) left as a follow-up.
- Default model IDs and per-second prices should be double-checked against the Bailian console before deployment; both are override-able per account/group.


SipengXie2024:
维护者好，这个 PR 已就绪，方便时望抽空 review，多谢 🙏

一句话简述：新增阿里百炼（DashScope）平台——wan/万相与 happyhorse/欢乐马视频生成（异步任务中转 + 按秒计费）以及 qwen 文本中转（OpenAI 兼容端点）。纯 API Key 平台，无 OAuth 依赖，对现有 Grok 等平台零行为变更。

分支已持续 rebase 到最新 main（含 composite 平台合并后的适配），后端单测全绿、CLA 已通过。有任何意见我会尽快跟进修改。

zhaoweiguo:
这个feature有合并计划吗？

PR #4832, opened 2026-07-24
fix(openai): sanitize API-key responses item IDs

ListenCodes:
## Summary

- sanitize replayed OpenAI Responses input item IDs for API-key accounts before passthrough or WS dispatch
- remove invalid `message.id` values unless they start with `msg`, and invalid call-input IDs unless they start with `fc`
- preserve valid IDs, output-item IDs, `call_id`, and all unrelated fields
- share the narrow prefix predicate with the existing OAuth filter from #4003 without applying the full OAuth transform to API-key requests

## Root cause

#3981 and #4003 covered the Codex OAuth transform, but OpenAI API-key passthrough returns before that transform runs. Codex clients can replay generated `item_*` IDs in full-history requests, so those IDs reached `/v1/responses` unchanged and OpenAI rejected them.

Invalid IDs are deleted rather than rewritten because fabricated `msg_*` or `fc_*` values may refer to a different upstream object.

## Tests

- regression test verified red on `upstream/main` and green with this patch
- focused API-key passthrough and existing `TestFilterCodexInput*` guards pass
- `go test -tags=unit ./...` passes
- `go build ./...` passes
- `golangci-lint run ./...`: 0 issues
- integration suite passes when excluding the unrelated flaky `TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig`; that test also fails on an unmodified `upstream/main` checkout at the same rate (2/3 local repetitions)

Refs #3981. Follows up #4003.

### DIFF
--- a/backend/internal/service/openai_codex_transform.go
+++ b/backend/internal/service/openai_codex_transform.go
@@ -1494,25 +1494,9 @@ func filterCodexInputWithOptions(input []any, opts codexInputFilterOptions) []an
 		if !opts.PreserveReferences {
 			ensureCopy()
 			delete(newItem, "id")
-		} else if isCodexToolCallInputType(typ) {
-			// 续链模式下保留 id 以维持上下文引用，但 function_call 等
-			// call-input 类 item 的 id 必须以 "fc" 开头（上游校验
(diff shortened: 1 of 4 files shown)


ListenCodes:

PR #4839, opened 2026-07-24
fix(composite): pass the requested model through when a prefix route leaves upstream_model empty

visa2:
## Problem

Composite model routes support a `match_type: prefix` mode with an optional
`upstream_model`. Leaving `upstream_model` empty is meant to pass the concrete
requested model through to the upstream, but every request matching the prefix
collapses to the route's `public_model` instead.

Example: a route with `public_model=deepseek-v4`, `match_type=prefix` and an
empty `upstream_model`. Requests for `deepseek-v4-flash` and `deepseek-v4-pro`
are both forwarded upstream as `deepseek-v4`, so the concrete model suffix is
lost and both requests hit the same upstream model.

## Root cause

`normalizeCompositeRouteInput` backfills an empty `upstream_model` with
`public_model` at save time:

```go
if input.UpstreamModel == "" {
    input.UpstreamModel = input.PublicModel
}
```

`CompositeRouteResolver.Resolve` already carries the intended passthrough
fallback:

```go
upstreamModel := strings.TrimSpace(route.UpstreamModel)
if upstreamModel == "" {
    upstreamModel = model // concrete requested model
}
```

Because the value is never stored empty, that fallback is dead code for explicit
routes. For exact matches the backfill is harmless (`model == public_model`),
but for prefix matches it turns every matched model into the fixed
`public_model`.

## Fix

- Skip the `upstream_model → public_model` backfill for `prefix` routes only.
  An empty `upstream_model` on a prefix route now stays empty and is resolved to
  the concrete requested model by the existing `Resolve` fallback.
- `exact` routes keep the backfill: an exact match always forwards a single
  concrete model (`model == public_model`), so preserving a non-empty stored
  value keeps the persisted/admin-display contract unchanged.
- Prefix route + empty upstream → each matched model forwards verbatim
  (`deepseek-v4-flash` → `deepseek-v4-flash`, `deepseek-v4-pro` → `deepseek-v4-pro`).
- Prefix/exact route + explicit upstream → forwards that fixed model (unchanged).
- No forward or billing path reads the raw `route.UpstreamModel`; the upstream
  model is produced solely by `Resolve` (`decision.UpstreamModel`), which never
  yields an empty value. Existing stored routes are unaffected (their
  `upstream_model` was already backfilled in the DB); re-saving a prefix route
  with an empty field opts into passthrough.
- Frontend: add an `upstream_model` hint (zh/en) documenting the
  empty = passthrough semantics.

## Tests

- `TestCompositeRouteResolverPrefixEmptyUpstreamPassesThroughRequestedModel` —
  prefix route with empty upstream; `deepseek-v4-flash` / `deepseek-v4-pro` /
  `deepseek-v4` each pass through unchanged.
- `TestCompositeRouteResolverPrefixExplicitUpstreamStillFixed` — prefix route
  with an explicit upstream still forwards the fixed model.
- `TestAdminService_CreateCompositeRoute_ExactEmptyUpstreamBackfillsPublicModel` —
  exact route with empty upstream still backfills `public_model`.
- Updated `TestAdminService_CreateCompositeRoute_NormalizesAndPersists` to assert
  a prefix route preserves an empty upstream (no backfill).
- Full `go test -tags unit ./internal/service/ ./internal/handler/...` is green.


visa2:

visa2:
recheck

PR #4848, opened 2026-07-25
fix(apicompat): emit rs-prefixed Responses reasoning item IDs

r266-tech:
Responses compatibility bridges generate generic `item_*` IDs for reasoning output, which strict Responses upstreams reject when that history is replayed. The Chat Completions stream also generates a different reasoning ID in `response.completed` from the one announced earlier.

Generate `rs_*` reasoning IDs in both Anthropic and Chat Completions conversions, and reuse the announced ID through the Chat Completions stream lifecycle. Non-compact OAuth passthrough removes reasoning input IDs under its existing `store=false` contract, preserving encrypted content, summaries, other fields, and non-reasoning IDs. Compact requests retain their IDs.

Rebased onto current main, preserving its Anthropic interleaved-thinking closure and request normalization. The newer API-key-side invalid-ID repair does not fix bridge output IDs or this OAuth normalization seam. Existing regression coverage now uses the current converter signature.

Validation on Go 1.27.1: `go test -tags=unit ./internal/pkg/apicompat ./internal/service -count=1` passed both complete packages; `go vet -tags=unit ./internal/pkg/apicompat ./internal/service`, gofmt, and `git diff --check` passed. `golangci-lint` was unavailable locally and its release download was refused by the local publication guard, so no fresh golangci-lint result is claimed.

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
+++ b/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
@@ -36,7 +36,7 @@ func AnthropicToResponsesResponse(resp *AnthropicResponse) *ResponsesResponse {
 			if block.Thinking != "" {
 				outputs = append(outputs, ResponsesOutput{
 					Type: "reasoning",
-					ID:   generateItemID(),
+					ID:   generateReasoningItemID(),
 					Summary: []ResponsesSummary{{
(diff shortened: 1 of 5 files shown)


r266-tech:

r266-tech:
recheck

PR #4855, opened 2026-07-25
feat(openai): support Anthropic /v1/messages over Responses WebSocket v2

SolitaryJune:
## Summary
- Add an account-level opt-in for OpenAI OAuth accounts to forward Claude Code `/v1/messages` through a single-turn Responses WebSocket v2 path, while keeping the downstream Anthropic SSE/JSON contract.
- Prefer `openai_ws_mode` when rendering usage request types so admin/user usage logs show WS instead of stream/sync for WS-backed traffic.
- Fix `scanUsageLog` so mixed `request_type=stream/sync` + `openai_ws_mode=true` rows no longer lose the WS flag via `ApplyLegacyRequestFields`.

## Scope notes
- Feature is **default off** and only applies to OpenAI OAuth accounts.
- First phase is single-turn direct-dial WS (`store=false`, no `previous_response_id` continuation / connection pooling).
- Includes unit coverage for the single-turn forwarder, DTO request-type preference, and repository scan legacy-field sync.

## Test plan
- [ ] Unit: `go test` for messages WS single-turn, DTO usage mapping, and `scanUsageLog` request-type cases
- [ ] Enable the account extra flag on one OpenAI OAuth account only
- [ ] Send Claude Code `/v1/messages` stream/sync requests and confirm upstream `/v1/responses` WS path + usage settlement
- [ ] Confirm usage API/admin UI shows `ws_v2` / WS for those rows
- [ ] Confirm non-WS accounts still show stream/sync

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -1115,12 +1115,13 @@ type GatewayOpenAIWSConfig struct {
 	// OAuthMaxConnsFactor: OAuth 账号连接池系数（effective=ceil(concurrency*factor)）
 	OAuthMaxConnsFactor float64 `mapstructure:"oauth_max_conns_factor"`
 	// APIKeyMaxConnsFactor: API Key 账号连接池系数（effective=ceil(concurrency*factor)）
-	APIKeyMaxConnsFactor  float64 `mapstructure:"apikey_max_conns_factor"`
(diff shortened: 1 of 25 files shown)


SolitaryJune:

PR #4856, opened 2026-07-25
fix(openai): estimate CC stream usage when upstream omits include_usage

SolitaryJune:
## Summary
Some OpenAI-compatible vendors (e.g. fengwind) ignore `stream_options.include_usage` on `/v1/chat/completions` streams, so Sub2API settled stream requests as **0 tokens / 0 cost** even though content was returned.

This PR adds a local fallback for the three CC stream paths:
- `/v1/messages` → CC fallback
- `/v1/responses` → CC fallback
- raw `/v1/chat/completions`

When the upstream stream ends without any usage counters, the gateway estimates:
- **input_tokens** from the request body text (messages / system / tools / instructions)
- **output_tokens** from accumulated `content` + `reasoning_content` + tool call args

Upstream-provided usage is always preferred; estimation only runs when every usage field is zero.

## Test plan
- [x] unit: keeps real upstream usage unchanged
- [x] unit: estimates when usage chunk is missing (messages / responses / raw CC)
- [x] unit: existing include_usage happy-path tests still pass
- [ ] production: accounts that must stay on force_chat_completions no longer log 0 for successful streams

### DIFF
--- a/backend/internal/service/openai_gateway_cc_pipeline.go
+++ b/backend/internal/service/openai_gateway_cc_pipeline.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Wei-Shaw/sub2api/internal/pkg/logger"
 	"github.com/Wei-Shaw/sub2api/internal/util/responseheaders"
 	"github.com/gin-gonic/gin"
+	"github.com/tidwall/gjson"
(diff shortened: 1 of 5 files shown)


SolitaryJune:

PR #4861, opened 2026-07-25
test: stop four concurrency tests from failing on a busy machine

coo1white:
Four tests in `internal/service` and `internal/securityaudit` pass on an idle machine and fail on a loaded one. Running them on a self-hosted runner, each of four consecutive CI runs failed with a *different* subset, while `make test-unit` was green on the same box when nothing else competed for CPU.

None of these tests is testing the clock, so none of them should be failing on it. Each has its own cause:

**`ollama_cloud_usage_test.go`** — `TestOllamaCloudUsageRefreshSingleflightAndRunnerDeduplicateSharedGroup`

`close(release)` fired right after the second goroutine was *started*, not after it had reached the singleflight group. When the first refresh won that race, the second became a fresh singleflight execution, re-read the account, saw the `LastAttemptAt` the first one had just written, and returned the 30-second manual-refresh 429 at line 685:

```
Error: Received unexpected error:
       code=429 reason="OLLAMA_CLOUD_USAGE_REFRESH_RATE_LIMITED"
```

`refreshAccount` loads the account once to build the group key, immediately before `refreshGroup.Do`. The test now counts loads and waits for the second caller's own load before releasing the first, so the join is no longer a guess. Adds a counting `GetByID` to the test repo.

**`gateway_hotpath_optimization_test.go`** — `TestGetUserGroupRateMultiplier_UsesCacheAndSingleflight`

A 20ms sleep was meant to let all 12 callers reach the cache before the loader was released. A caller arriving after the load had already finished recorded a *hit*, so `require.Equal(t, int64(12), miss)` saw 11. It now waits on the miss counter itself — the value the test asserts on.

**`token_refresh_pool_health_test.go`** — `TestTokenRefreshService_SaturatedProviderPreservesConcurrencyAndActualQPSStartSpacing`

The start timestamp is taken *after* the rate gate releases the goroutine, so scheduler delay compresses the observed gap even when pacing was correct (observed 32.4ms and 37.8ms against the 40ms floor). Slack is now half the configured spacing, at least 25ms. Starts bunched with no pacing at all — the regression this actually guards — are still caught.

**`prompt_guard_test.go`** — `TestGuardEvaluatorFlagSharedDeadlineFailClosedAndContextCancel`

The bound only has to show the failover shared the first endpoint's 70ms deadline rather than taking the second endpoint's own 500ms one. An unshared deadline lands near 535ms, so `< 350ms` still fails loudly, while `< 180ms` was flaky (observed 224ms).

---

Tests only — no product code is touched. Verified with `-count=10` on each, and again under `GOMAXPROCS=1` to starve the scheduler.

### DIFF
--- a/backend/internal/securityaudit/prompt_guard_test.go
+++ b/backend/internal/securityaudit/prompt_guard_test.go
@@ -222,7 +222,12 @@ func TestGuardEvaluatorFlagSharedDeadlineFailClosedAndContextCancel(t *testing.T
 		elapsed := time.Since(started)
 		require.Error(t, err)
 		require.Equal(t, 2, calls)
-		require.Less(t, elapsed, 180*time.Millisecond)
+		// The bound only has to prove the failover shared the first endpoint's
+		// 70ms deadline instead of taking the second endpoint's own 500ms one.
+		// An unshared deadline lands at ~535ms, so 350ms still fails loudly
+		// while leaving room for scheduler delay on a busy CI machine. A
+		// tighter bound made this test flaky, not stricter.
+		require.Less(t, elapsed, 350*time.Millisecond)
 		require.GreaterOrEqual(t, elapsed, 50*time.Millisecond)
 		require.Equal(t, int64(1), metrics.Snapshot().Failovers)
 		require.Equal(t, int64(1), metrics.Snapshot().Timeouts)
--- a/backend/internal/service/gateway_hotpath_optimization_test.go
+++ b/backend/internal/service/gateway_hotpath_optimization_test.go
@@ -219,7 +219,15 @@ func TestGetUserGroupRateMultiplier_UsesCacheAndSingleflight(t *testing.T) {
 	}
 
 	close(start)
-	time.Sleep(20 * time.Millisecond)
+	// Wait for every caller to have recorded its cache miss before releasing the
+	// loader. A fixed sleep raced here: a goroutine that reached the cache after
+	// the singleflight load had already finished got a hit instead of a miss, and
+	// the miss assertion below saw 11 of 12. The miss counter is the observable
+	// that says "all callers are now inside the singleflight group".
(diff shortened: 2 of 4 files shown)


coo1white:
Updated the `token_refresh_pool_health_test.go` part — the first version of it was not good enough, and the reasoning behind it was wrong.

**What was wrong.** I had widened the floor to `configuredSpacing - max(configuredSpacing/2, 25ms)` = 25ms and treated the problem as a tolerance that needed loosening. It failed again on a loaded machine at **13.3ms**. Loosening further would have been guessing.

**What I should have measured first:** what the assertion looks like when the gate is genuinely broken. Swapping the shared rate gate for `newTokenRefreshRateGateWithInterval(0)` (providerQPS=20, 8 attempts, concurrency 2):

| | minimum observed gap | span of all starts |
|---|---|---|
| gate at 50ms | **49.97ms** | 471ms |
| gate at 0 | **22µs** | 241ms |

Two things fall out:

1. **The original approach was right.** Minimum adjacent spacing separates paced from unpaced by three orders of magnitude. Only the number was wrong. The floor is now `configuredSpacing / 10` — 5ms here, which an unpaced gate misses by a factor of ~200 while leaving jitter plenty of room.

2. **A span assertion would have been vacuous.** I had also considered asserting on the total span of the starts, on the theory that the gate bounds it from below. It does not: the span is set by how long each attempt takes under the concurrency limit, not by the gate — 471ms paced against 241ms unpaced. I actually wrote that version, ran it against a zero-interval gate, and watched it pass. There is now a comment in the test warning against it, since it is the natural next idea for anyone who finds this flaky again.

Every fix in this PR is now checked in both directions: passes with the behaviour intact, still fails when the behaviour is broken on purpose. For the QPS one that means the zero-interval gate above; the failure reads `"56.625µs" is not greater than or equal to "5ms"`.

Also rebased onto current `main`.

PR #4869, opened 2026-07-25
feat(pool): support unlimited same-account retry via -1

SolitaryJune:
## Summary
- `pool_mode_retry_count = -1` now means unlimited same-account retries in pool mode (until the client disconnects or the error is no longer same-account-retryable).
- `0` still disables in-place retry; `1..10` remain finite limits (values `>10` still clamp to 10; other negatives clamp to 0).
- Admin create/edit UI accepts `-1` and updates zh/en hints accordingly.

## Why
Some pooled upstreams need continuous same-account retry instead of failing over after a small fixed budget. The previous max was 10, and `0` meant "no retry" rather than "infinite".

## Test plan
- [x] `go test -tags=unit ./internal/service/ -run TestGetPoolModeRetryCount`
- [x] `go test -tags=unit ./internal/handler/ -run TestHandleFailoverError`
- [ ] Set an account `pool_mode=true`, `pool_mode_retry_count=-1`, trigger retryable upstream status; confirm continuous same-account retries without switch/temp-unschedule until client cancel
- [ ] Set `0` and confirm immediate switch; set `1` and confirm single retry then switch

### DIFF
--- a/backend/internal/handler/failover_loop.go
+++ b/backend/internal/handler/failover_loop.go
@@ -84,14 +84,15 @@ func (s *FailoverState) HandleFailoverError(
 	}
 
 	// 同账号重试不算切换账号，粘性会话仅在实际切换时强制缓存计费。
-	sameAccountRetry := failoverErr.RetryableOnSameAccount && s.SameAccountRetryCount[accountID] < retryLimit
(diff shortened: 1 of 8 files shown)


SolitaryJune:

PR #4804, opened 2026-07-23
fix(openai): strip input item namespaces before HTTP forwarding / HTTP 转发前移除 input 项 namespace

scp-planet:
### English

- Proactively remove `namespace` from direct items in the OpenAI Responses API `input` array before the first upstream HTTP request for both OAuth and API Key accounts.
- Apply the behavior to Responses and Compact HTTP forwarding while preserving native Responses WebSocket v2 payloads.
- Keep namespace declarations outside direct input items, including tool namespaces and nested content namespaces.
- Preserve the existing rejected-field retry flow for unrelated unsupported parameters.
- Rebuild only the `input` array with `gjson`/`sjson` so untouched JSON values retain their original representation.

### 中文

- 对 OpenAI OAuth 和 API Key 账号，在第一次 HTTP 上游请求前主动移除 Responses API `input` 数组直接子项中的 `namespace`。
- 覆盖 Responses 和 Compact 的 HTTP 转发路径，同时保留原生 Responses WebSocket v2 请求中的字段。
- 不修改直接 input 项以外的命名空间声明，包括工具 namespace 和嵌套 content namespace。
- 保留现有的拒绝字段重试机制，用于处理其他不受支持的参数。
- 使用 `gjson`/`sjson` 仅重建 `input` 数组，确保未修改的 JSON 值保持原始表示。

### English

- Added unit coverage for OAuth and API Key account gating, HTTP and WebSocket v2 transport behavior, and Responses/Compact forwarding.
- Verified that eight direct input-item namespace fields are removed before the first upstream request.
- Verified preservation of nested namespaces, tool namespaces, large integers, scientific notation, escaped strings, and unrelated input shapes.
- Verified composition with the existing rejected-field retry flow.
- Relevant `internal/service` unit tests pass.

### 中文

- 新增单元测试，覆盖 OAuth/API Key 账号判断、HTTP/WebSocket v2 传输行为，以及 Responses/Compact 转发路径。
- 验证 8 个直接 input 项的 namespace 均在第一次上游请求前被移除。
- 验证嵌套 namespace、工具 namespace、大整数、科学计数法、转义字符串及其他 input 结构保持不变。
- 验证该逻辑可与现有拒绝字段重试机制正确组合。
- 相关 `internal/service` 单元测试已通过。

### English

[...]

PR #4888, opened 2026-07-26
fix(openai): preserve namespaced Codex calls

### LINKED ISSUE #4761: 上游报错 400 Unknown parameter: input[894].namespace 导致用户端显示 502 报错
最近出现了好几个 Unknown parameter: 'input[894].namespace 的 400 报错，体现在使用上就是 502 ，这个怎么处理呢

```
{
  "error": {
    "code": "unknown_parameter",
    "message": "Unknown parameter: 'input[894].namespace'.",
    "param": "input[894].namespace",
  }
}
```

lyy0709:
## Summary

- Fix the P0 regression introduced by #4804 while retaining its linear cleanup for unsupported direct `input[]` namespace fields.
- Preserve `namespace` on `function_call` and `custom_tool_call` items so Codex can round-trip the tool identity.
- Do not let the rejected-field retry remove those protocol fields after an upstream 400.
- Keep existing namespace declaration flattening, response restoration, nested-field handling, HTTP account gating, and native WSv2 behavior unchanged.

Fixes #4761.

## Why #4804 is incorrect

#4804 deletes `namespace` from every direct Responses `input[]` object. That includes historical namespaced tool calls. The upstream then cannot resolve calls such as `collaboration/list_agents` and returns `Missing namespace ... Round-trip the model function_call item with its namespace field included`.

This is not only a historical Codex behavior. The latest `openai/codex` main checked for this PR is commit `61a44880a85d2fd0d8770908dea5733495e571c8`:

- [`ResponseItem::FunctionCall` and `ResponseItem::CustomToolCall` both serialize an optional namespace](https://github.com/openai/codex/blob/61a44880a85d2fd0d8770908dea5733495e571c8/codex-rs/protocol/src/models.rs#L861-L923).
- [The Responses request builder clones and forwards the prompt input without stripping call namespaces](https://github.com/openai/codex/blob/61a44880a85d2fd0d8770908dea5733495e571c8/codex-rs/core/src/client_common.rs#L20-L60).
- [The tool router combines namespace and name to identify both function and custom tool calls](https://github.com/openai/codex/blob/61a44880a85d2fd0d8770908dea5733495e571c8/codex-rs/core/src/tools/router.rs#L130-L174).

Therefore the cleanup must be type-aware: remove schema-invalid residual namespaces from messages, outputs, and unknown direct items, but preserve them on the two call item types defined by current Codex.

## Regression coverage

- Removes eight unsupported direct item namespaces in one pass, beyond the old six-retry limit.
- Preserves function and custom tool call namespaces byte-for-byte when no cleanup is needed.
- Covers OAuth and API Key forwarding for both `/v1/responses` and `/v1/responses/compact` without requiring current tool declarations.
- Verifies an unrelated `max_output_tokens` retry keeps the call namespaces on every attempt.
- Verifies an explicit namespace rejection does not trigger a destructive retry.

## Validation

- `go test -tags=unit ./internal/service -run <namespace-and-rejected-field-tests> -count=1`
- `go test -tags=unit ./internal/pkg/apicompat -count=1`
- `golangci-lint run ./internal/service ./internal/pkg/apicompat` (`0 issues`)
- `git diff --check`

### DIFF
--- a/backend/internal/service/openai_responses_namespace.go
+++ b/backend/internal/service/openai_responses_namespace.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/Wei-Shaw/sub2api/internal/pkg/apicompat"
 	"github.com/gin-gonic/gin"
@@ -65,9 +66,11 @@ func flattenOpenAIResponsesNamespaces(c *gin.Context, body []byte) ([]byte, erro
 }
 
 // stripOpenAIResponsesInputNamespaces removes namespace only from direct input
-// array items. Namespace declarations and nested namespace fields are left
-// untouched. Rebuilding the input array once keeps this linear for long
-// histories and avoids decoding JSON numbers through float64.
+// array items whose Responses schema does not carry it. Function and custom
+// tool calls must retain namespace so Codex can round-trip namespaced tools.
+// Namespace declarations and nested namespace fields are left untouched.
+// Rebuilding the input array once keeps this linear for long histories and
+// avoids decoding JSON numbers through float64.
 func stripOpenAIResponsesInputNamespaces(body []byte) ([]byte, error) {
 	if !bytes.Contains(body, []byte(`"namespace"`)) {
 		return body, nil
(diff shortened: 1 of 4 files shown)


lyy0709:

PR #4909, opened 2026-07-26
fix(openai): support previous_response_id on REST responses

### LINKED ISSUE #2133: Follow-up to #1957: HTTP /v1/responses still rejects previous_response_id on 0.1.120
Follow-up to #1957.

Hi, thanks again for the earlier fix.

From my local testing on sub2api 0.1.120, it looks like the reasoning / `rs_*` replay side of #1957 may be improved, but HTTP `POST /v1/responses` still rejects `previous_response_id`.

Observed error:

`400 previous_response_id is only supported on Responses WebSocket v2`

The server log also reports:

`reason: previous_response_id_requires_wsv2`

From testing, enabling the account-level Responses WebSocket v2 flags for an OpenAI OAuth account did not make HTTP `/v1/responses` accept `previous_response_id`.

So my current understanding is:
- the reasoning replay part may be fixed
- but HTTP `/v1/responses` still treats `previous_response_id` as WSv2-only

Could you please confirm whether this is intentional?
If yes, what is the intended public/client-facing WSv2 entry path for this feature?

I’m keeping this report sanitized, but I can provide a minimal reproduction example without sensitive details if needed.

ZacharyHu0:
## Summary

- accept string or `null` `previous_response_id` values on REST `POST /v1/responses` while continuing to reject message/item IDs and invalid JSON types
- preserve non-empty IDs for native Responses API-key accounts over HTTP JSON/SSE
- expose eligible standard OAuth WSv2 accounts through the existing REST JSON/SSE facade from the first turn, keeping each continuation on the account and connection that produced the ID
- return `400 previous_response_not_found` for unknown, expired, cross-group, or connection-lost IDs instead of dropping the ID or failing over to another account
- retain OAuth response connections for the response-affinity TTL, bounded by `max_conns_per_account`, including independent REST sessions when the concurrency-derived pool limit is lower than the hard cap
- keep `function_call_output.call_id` validation strict when a continuation ID is present

This fixes the downstream regression reported in Lingtai-AI/lingtai-kernel#859, where LingTai had to replay the complete canonical history because sub2api rejected REST continuation IDs as WSv2-only.

OpenAI documents `previous_response_id` as the HTTP conversation-state mechanism, with WebSocket mode using the same continuation semantics: https://developers.openai.com/api/docs/guides/conversation-state#passing-context-from-the-previous-response

## Compatibility

| Account path | REST behavior |
| --- | --- |
| Native Responses API key | Keep HTTP JSON/SSE and forward the non-empty ID |
| Standard OAuth with Responses WSv2 enabled | Bridge REST JSON/SSE to the existing WSv2 forwarder and enforce account/connection affinity |
| OAuth PAT or Agent Identity | Do not opt into the REST-to-WSv2 bridge |

OAuth connection bindings remain process-local because the upstream response state is owned by a live WebSocket connection. A missing local connection therefore fails explicitly with `previous_response_not_found`; it never drifts to a new connection or account. Deployments with multiple sub2api instances should keep an active response chain on the same instance.

## Validation

- `go test ./internal/handler ./internal/service -count=1`
- `go test -tags=unit ./internal/handler ./internal/service -count=1`
- `go vet ./internal/handler ./internal/service`
- focused `go test -race` coverage for response connection retention, TTL refresh/cleanup, dynamic and Mode Router pool limits, and REST OAuth `A -> B -> A` continuation
- `git diff --check`

Fixes #2133

### DIFF
--- a/backend/internal/handler/openai_gateway_handler.go
+++ b/backend/internal/handler/openai_gateway_handler.go
@@ -280,7 +280,18 @@ func (h *OpenAIGatewayHandler) Responses(c *gin.Context) {
 		return
 	}
 	reqLog = reqLog.With(zap.String("model", reqModel), zap.Bool("stream", reqStream))
-	previousResponseID := strings.TrimSpace(gjson.GetBytes(body, "previous_response_id").String())
+	previousResponseResult := gjson.GetBytes(body, "previous_response_id")
+	previousResponseID := ""
+	if previousResponseResult.Exists() {
+		switch previousResponseResult.Type {
+		case gjson.Null:
+		case gjson.String:
+			previousResponseID = strings.TrimSpace(previousResponseResult.String())
+		default:
+			h.errorResponse(c, http.StatusBadRequest, "invalid_request_error", "previous_response_id must be a string or null")
+			return
+		}
+	}
 	if previousResponseID != "" {
 		previousResponseIDKind := service.ClassifyOpenAIPreviousResponseIDKind(previousResponseID)
 		reqLog = reqLog.With(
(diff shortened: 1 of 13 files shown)


ZacharyHu0:

PR #4920, opened 2026-07-26
feat: add passkey authentication

alexj11324:
## Summary

- add WebAuthn passkey registration, discoverable sign-in, listing, rename, and revocation
- validate RP ID and HTTPS origins at startup, require user verification and resident credentials, and keep ceremonies single-use with a five-minute Redis TTL
- add the existing System Settings → Security & Authentication owner as the persistent passkey sign-in switch, with visible RP configuration status and no parallel state
- keep credential management in the user profile; when sign-in is disabled, existing credentials remain visible and revocable
- add audit coverage, bounded WebAuthn request bodies, credential backup-state handling, and the public passkey availability field

## Configuration

Passkeys remain unavailable unless the server has a valid WebAuthn configuration:

```yaml
webauthn:
  enabled: true
  rp_display_name: Sub2API
  rp_id: example.com
  rp_origins:
    - https://example.com
```

`rp_id` must be a hostname without a scheme, port, or path. Every origin must use HTTPS, contain no path/query/fragment, and have a host equal to or below the RP ID. The admin switch cannot weaken these checks.

## Migration

Migration `191_passkey_credentials.sql` creates the passkey user-handle and credential tables plus indexes. No existing user data is rewritten; users enroll credentials from their profile after deployment.

## Verification

- focused Go service/handler/admin tests pass, including persistence, public visibility, disabled behavior, store-error handling, list/revoke availability, and request-size limits
- all affected Go packages compile
- frontend Settings and passkey API tests: 28 passed
- frontend typecheck and production build pass
- browser-tested on an isolated HTTPS ARM64 deployment: password login, passkey registration/rename, sign-out and passkey sign-in, revocation, rejected revoked credential, and password-login regression
- live admin UI shows the enabled switch plus valid RP ID and allowed HTTPS origin
- reviewed for secrets and deployment-only artifacts; none are included

## Security note

Enrollment and revocation currently rely on the existing authenticated session, consistent with other profile credential operations. This PR does not add a recent-password or step-up requirement for those actions; maintainers may want to require recent authentication before merge if that is the project policy.


alexj11324:

Wei-Shaw:
Pushed `97f44b21b` (maintainer edit) to get CI green and fix one functional issue found in review:

1. **Stale-switch deadlock (functional)**: `parseSettings` reported `passkey_enabled=true` from a leftover DB row even after the WebAuthn deployment config was removed. The new update-gate in `UpdateSettings` then rejected *every* admin settings save, while the UI toggle was disabled (`:disabled="!passkey_configured"`) so there was no recovery path from the panel. `passkeySettingEnabled` is now ANDed with `passkeyConfigured()` (matching the runtime `PasskeyEnabled()` semantics), plus a regression test.
2. **Contract test**: added the four new `passkey_*` fields to the `GET /api/v1/admin/settings` goldens in `api_contract_test.go`.
3. **Lint**: errcheck on `rows.Close()` in `passkey_repo.go` (repo convention `defer func() { _ = rows.Close() }()`), and staticcheck QF1001 (De Morgan) in the origin scheme validation in `config.go`.

No behavioral changes beyond item 1. The step-up/recent-auth question from your security note is intentionally left untouched and will be decided separately.

Wei-Shaw:
Pushed `38ef8dc06` (maintainer edit), resolving the open question from the PR's security note: **passkey enrollment and revocation now require the account password**.

- `BeginRegistration` and `Delete` verify the current user's password server-side (`verifyPasskeyPassword` in `service/passkey.go`), reusing the existing `PASSWORD_REQUIRED` / `PASSWORD_INCORRECT` errors. A hijacked session can no longer silently enroll an attacker passkey or revoke the victim's credentials.
- Password was chosen over TOTP step-up deliberately, so the guard also protects deployments without a `TOTP_ENCRYPTION_KEY` configured.
- The `password` key in both request bodies is covered by the audit middleware's key-substring redaction (`audit_log.go` sensitive-key list), so nothing credential-shaped reaches `audit_logs`.
- Frontend: the add form gains a current-password field; deletion now uses a password-confirm dialog (replacing `window.confirm`), modeled on `TotpDisableDialog`. Backend error messages (wrong password) are surfaced to the user.
- Rename stays password-free (cosmetic). Finish endpoints stay password-free: they are already bound to the one-time Redis ceremony session issued by the password-gated begin step.
- Note for docs: OAuth-provisioned accounts receive a random password at signup, so those users must set a password (password reset flow) before managing passkeys — same trade-off as TOTP setup with email verification disabled.

Coverage: service-level test `TestPasskeyEnrollmentAndRevocationRequireAccountPassword` plus frontend API contract specs for the password payloads.

PR #4927, opened 2026-07-26
feat(groups): add prompt processing policies

706412584:
## Summary
- add a persisted, validated prompt policy to every account group
- apply replace, block, prepend, and append rules before auditing and routing for Chat Completions, Messages, and Responses
- expose policy configuration in the admin group form and preserve it when duplicating groups

## Test plan
- [x] `go test ./internal/service ./internal/repository ./internal/handler ./internal/handler/admin ./internal/handler/dto`
- [x] `pnpm typecheck`
- [x] `pnpm lint && pnpm build`
- [ ] Manual API smoke test against a current local build (native service was still running an older binary)


706412584:

PR #4929, opened 2026-07-26
feat(openai): support targeted reset credit redemption

HeST-lab:
## Summary

- add an admin-only endpoint that returns available Codex reset credits with their opaque IDs and expiration metadata
- add targeted reset-credit redemption with a caller-provided `Idempotency-Key` forwarded as the upstream `redeem_request_id`
- reuse the existing admin idempotency coordinator while failing open to the upstream idempotency contract if the local store is unavailable
- preserve the existing quota response sanitization and legacy `/reset-quota` behavior

## Notes

- no database migration or frontend change is required
- `details_complete` is false when upstream caps the returned rows or an available credit lacks the ID required for targeted redemption
- reset-credit IDs are exposed only through the authenticated admin route, responses use `Cache-Control: no-store`, and IDs/idempotency keys are not logged
- the request and response fields follow the official Codex reset-credit client contract: https://github.com/openai/codex/blob/main/codex-rs/backend-client/src/client/rate_limit_resets.rs

## Tests

- `go test -tags=unit ./internal/service -run 'Test(QueryResetCredits|ResetCreditByID|ResetCreditLegacyRequest|TargetedResetOperations|ParseOpenAIRateLimitResetCreditDetails|QueryUsageResetCredit)' -count=1`
- `go test -tags=unit ./internal/handler/admin ./internal/server/routes ./internal/server -count=1`
- `go vet -tags=unit ./internal/service ./internal/handler/admin ./internal/server/routes`
- `git diff --check`

Full `go test -tags=unit ./... -count=1` was also attempted on Windows. The changed packages pass, but the repository baseline still has a Windows path failure in `TestResolvePageImagePath`; the unrelated timing-sensitive `TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig` can fail under package-wide load and passes when rerun alone.

### DIFF
--- a/backend/internal/handler/admin/openai_oauth_handler.go
+++ b/backend/internal/handler/admin/openai_oauth_handler.go
@@ -1,6 +1,7 @@
 package admin
 
 import (
+	"context"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


HeST-lab:

PR #4932, opened 2026-07-26
Fix Codex (Responses API) <-> Anthropic tool compatibility

Ricardo-binZzz:
# Fix Codex (Responses API) <-> Anthropic tool compatibility

## Summary

When proxying an OpenAI Responses-style client (Codex / Codex++) to an Anthropic
(Claude) upstream via `ForwardAsResponses`, several tool-related payloads fail to
convert cleanly, causing production `400` and `502` errors. This PR makes the
Responses -> Anthropic conversion tolerant of the shapes Codex actually sends, and
restores the Codex-expected shapes on the way back.

All changes are scoped to the Responses -> Anthropic path. OpenAI and Grok
forwarding, Chat Completions, account scheduling, billing, DB schema, quotas and
the admin UI are untouched.

## Problems fixed

1. **`namespace` tools not converted / not restored.**
   Codex groups tools under a `namespace` parent. These were not flattened to
   Anthropic function tools on the way out, and child/namespace names were not
   restored in buffered and streaming Responses output. The streaming
   `function_call_arguments.delta` and `.done` events also leaked the internal
   flattened name (e.g. `team__ping`) instead of restoring `name=ping` with
   `namespace=team`.

2. **`function_call_output.output` as a non-string caused a 502.**
   Production error:
   `json: cannot unmarshal array into Go struct field ResponsesInputItem.output of type string`.
   The field now accepts string, object, or array output; text/image arrays are
   converted into valid Anthropic tool-result content blocks.

3. **Native Anthropic server tools rejected with 400.**
   Production error:
   `tools.N.web_search_20250305.input_schema: Extra inputs are not permitted`.
   `AnthropicTool.InputSchema` now uses `omitempty`, so native server tools
   (e.g. `web_search`) no longer serialize `input_schema: null`. Regular function
   tools still always include `input_schema`.

4. **Codex dynamic `additional_tools` ignored.**
   `input[].type == "additional_tools"` is now lifted into top-level `tools`
   before custom / tool-search / namespace lowering is applied.

## Scope / safety

- Only affects `ForwardAsResponses` request conversion and response restoration.
- Function tools always keep `input_schema`; only Anthropic native server tools
  omit an empty one.
- Namespace restore logic is shared with the Grok compat layer but only rewrites
  names that match the mapping, so it is a pure compatibility fix.

## Tests

Added / extended unit tests under `internal/pkg/apicompat` and `internal/service`:

- namespace flatten + streaming/buffered restore (incl. delta/done event names)
- `function_call_output.output` string/object/array handling
- `omitempty` on native tool `input_schema`
- `additional_tools` lifting + lowering

```
go test -tags=unit ./internal/pkg/apicompat -count=1
go test -tags=unit ./internal/service -count=1
```

Also verified against a live Claude upstream: streaming namespace, array
`function_call_output` (text+image), native `web_search`, and `additional_tools`
all return HTTP 200 with the expected restored structure.

### DIFF
--- a/backend/internal/pkg/apicompat/responses_client_tools.go
+++ b/backend/internal/pkg/apicompat/responses_client_tools.go
@@ -586,7 +586,7 @@ func (r *ResponsesClientToolStreamRestorer) restoreNamespaceEvent(event Response
 			event.Item.Name, event.Item.Namespace = name.Name, name.Namespace
 		}
 	}
-	if event.Type == "response.function_call_arguments.done" {
+	if event.Type == "response.function_call_arguments.delta" || event.Type == "response.function_call_arguments.done" {
 		if name, ok := r.adapter.NamespaceTools[event.Name]; ok {
 			event.Name = name.Name
 		}
--- a/backend/internal/pkg/apicompat/responses_client_tools_test.go
+++ b/backend/internal/pkg/apicompat/responses_client_tools_test.go
@@ -146,7 +146,13 @@ func TestResponsesClientToolStreamRestorer_RestoresNamespaceLifecycle(t *testing
 	require.Equal(t, "open", gjson.GetBytes(added[0], "item.name").String())
 	require.Equal(t, "browser", gjson.GetBytes(added[0], "item.namespace").String())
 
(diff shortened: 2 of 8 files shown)


Ricardo-binZzz:

PR #4942, opened 2026-07-27
fix(openai): honor passthrough over non-empty model_mapping in account selection (#4936)

### LINKED ISSUE #4936: OpenAI passthrough account wrongly excluded by `model_mapping` whitelist during account selection (passthrough bypass missing in scheduler eligibility)
## Summary

For an OpenAI OAuth account with **auto-passthrough enabled** (`extra.openai_passthrough = true`, "replace auth only, allow all models"), the gateway still refuses to route a model that is **not listed** in the account's `credentials.model_mapping`. The request fails account selection and the client receives:

```
404 Model "<model>" is not supported by any configured account in this group
```

even though the account can actually serve that model (a direct "test account connection" with the same model succeeds).

Root cause: the OpenAI account-selection eligibility check calls `Account.IsModelSupported()` **directly**, which enforces the `model_mapping` allowlist and is **unaware of passthrough**. A parallel model-support check does honor passthrough, so the two checks are inconsistent.

## Environment

- Version: observed on `v0.1.133` (+46 commits); code paths unchanged through at least `0.1.164`.
- Account type: OpenAI OAuth (Codex/ChatGPT), `openai_passthrough = true`.

## Root cause (code references)

Scheduler eligibility — **passthrough-blind** path used for OpenAI account selection:

- `internal/service/openai_gateway_service.go:1329` `isOpenAIAccountEligibleForRequest(...)`
  - line `1344`: `if requestedModel != "" && !account.IsModelSupported(requestedModel) { return false }`
- `internal/service/account.go:622` `func (a *Account) IsModelSupported(requestedModel string) bool`
  - Returns `true` only if the account has **no** `model_mapping` (`len==0` → allow all) **or** the model is present in the mapping. It has **no passthrough awareness**.

Parallel model-support check that **does** honor passthrough (inconsistent with the above):

- `internal/service/gateway_service.go:3801` `func (s *GatewayService) isModelSupportedByAccount(...)`
  - line `3813`: `if account.Platform == PlatformOpenAI && account.IsOpenAIPassthroughEnabled() { return true }` (bypasses the whitelist for passthrough accounts, then falls back to `IsModelSupported`)
- `internal/service/account.go:1287` `func (a *Account) IsOpenAIPassthroughEnabled() bool`

Because `isOpenAIAccountEligibleForRequest` uses the passthrough-blind check, a passthrough account with a non-empty `model_mapping` is filtered out of the candidate set for any model not in the mapping → zero candidates → `ErrNoAvailableAccounts`.

## Steps to reproduce

1. Configure an OpenAI OAuth account and enable auto-passthrough (`extra.openai_passthrough = true).
2. Ensure the account has a **non-empty** `credentials.model_mapping` that does **not** include some model `M` that the account can actually serve (e.g. a newly-released model). Note: such a mapping is commonly left over from before passthrough was enabled.
3. Put the account in a group and, as a client of that group, request model `M`, e.g. `POST /responses` with `{"model":"M", ...}`.

## Expected

The request is routed to the account (passthrough means "only replace auth, allow all models"), and `M` is forwarded upstream as-is.

## Actual

- Account selection excludes the account: `openai.account_select_failed`, `error: "no available accounts"`, `excluded_account_count: 0`.
- Client receives `404 Model "M" is not supported by any configured account in this group`.
- A direct admin "test account connection" with model `M` on the same account **succeeds** (it bypasses `IsModelSupported`), which is what makes the failure look contradictory.

## Suggested fix

In `isOpenAIAccountEligibleForRequest`, honor passthrough before the whitelist check — either:

- short-circuit when `account.Platform == PlatformOpenAI && account.IsOpenAIPassthroughEnabled()` (mirroring `isModelSupportedByAccount:3813`), or
- route the model-support check through the passthrough-aware `isModelSupportedByAccount` instead of calling `account.IsModelSupported` directly.

## Workaround (for operators)

Clear `credentials.model_mapping` on passthrough OpenAI accounts. Note that clearing it via the admin UI may not persist (the edit form can resubmit the existing mapping); clearing it directly in the DB (`credentials = credentials - 'model_mapping'`) and restarting the backend (to drop the cached account) is reliable

HuntercodeT:
Fixes #4936.

## Problem

An OpenAI account with auto-passthrough enabled (`extra.openai_passthrough = true`, "replace auth only, allow all models") is still filtered out during account selection when it carries a **non-empty** `credentials.model_mapping` that does not list the requested model. The client then gets:

```
404 Model "<model>" is not supported by any configured account in this group
```

…even though a direct admin "test account connection" with the same model **succeeds** (that path goes through `isModelSupportedByAccount`, which already honors passthrough).

## Root cause

Account selection (`isOpenAICompatibleAccountEligibleForRequest`, `openai_gateway_scheduling.go:290`) calls `Account.IsModelSupported(requestedModel)` directly. In `account.go`, `IsModelSupported` only bypasses the `model_mapping` whitelist for passthrough accounts **when the mapping is empty**:

```go
func (a *Account) IsModelSupported(requestedModel string) bool {
    mapping := a.GetModelMapping()
    if len(mapping) == 0 {
        if a.IsOpenAIOAuth() && !a.IsOpenAIPassthroughEnabled() {
            return isOpenAIOAuthServableModel(requestedModel)
        }
        return true
    }
    // mapping is non-empty -> whitelist enforced, passthrough ignored  <-- bug
    ...
}
```

A leftover non-empty mapping (common after switching an account from whitelist mode to passthrough) therefore excludes the account → zero candidates → `ErrNoAvailableAccounts` → 404. This is inconsistent with `isModelSupportedByAccount`, which short-circuits passthrough before the whitelist.

## Fix

Short-circuit passthrough at the top of `Account.IsModelSupported`, before the `model_mapping` check, so account selection is consistent with `isModelSupportedByAccount`. `IsOpenAIPassthroughEnabled()` is platform-guarded (returns false unless `IsOpenAI()`), so non-OpenAI accounts are unaffected. The now-redundant `&& !a.IsOpenAIPassthroughEnabled()` in the empty-mapping branch is removed.

## Test

Adds `TestIsModelSupported_OpenAIOAuthPassthroughIgnoresLeftoverMapping` covering passthrough + non-empty leftover mapping (the previously untested case). Existing `Passthrough|Wildcard|ModelSupport|Eligible|ModelMapping|Servable` tests still pass.

## Workaround (for operators on affected versions)

Clear `credentials.model_mapping` on passthrough OpenAI accounts. Note the admin UI edit form may resubmit the existing mapping; clearing it directly in the DB (`credentials = credentials - 'model_mapping'`) and restarting the backend (to drop the cached account) is reliable.

### DIFF
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -779,9 +779,16 @@ func resolveRequestedModelInMapping(mapping map[string]string, requestedModel st
 // 请求卡死在该账号上、无法 failover 到真正支持该模型的 API Key 账号（#3662）。
 // 未知/自定义别名仍保持允许（兼容渠道级映射），见 isOpenAIOAuthServableModel。
 func (a *Account) IsModelSupported(requestedModel string) bool {
+	// 透传模式仅替换认证、模型语义完全交由上游决定，因此放行所有模型。
+	// 该短路必须在 model_mapping 判定之前：账号从"白名单模式"切换到透传后，
+	// credentials 里常残留旧的非空 model_mapping，若不在此放行，透传账号会被
+	// model_mapping 白名单错误排除出候选集，导致 no available accounts / 404（issue #4936）。
+	if a.IsOpenAIPassthroughEnabled() {
+		return true
+	}
 	mapping := a.GetModelMapping()
 	if len(mapping) == 0 {
-		if a.IsOpenAIOAuth() && !a.IsOpenAIPassthroughEnabled() {
+		if a.IsOpenAIOAuth() {
 			return isOpenAIOAuthServableModel(requestedModel)
 		}
 		return true // 无映射 = 允许所有
--- a/backend/internal/service/openai_oauth_model_support_test.go
+++ b/backend/internal/service/openai_oauth_model_support_test.go
@@ -81,6 +81,20 @@ func TestIsModelSupported_OpenAIOAuthPassthroughAllowsAll(t *testing.T) {
 	require.True(t, account.IsModelSupported("deepseek-v4"))
 }
 
+func TestIsModelSupported_OpenAIOAuthPassthroughIgnoresLeftoverMapping(t *testing.T) {
+	account := newOpenAIOAuthAccountForModelTest()
+	account.Extra = map[string]any{"openai_passthrough": true}
+	// 账号从"白名单模式"切到透传后，credentials 里常残留旧的非空 model_mapping。
+	// 透传应无视该白名单，放行不在其中的模型（issue #4936）；否则透传账号会被
+	// 调度期的 IsModelSupported 排除，客户端收到 404 "not supported by any account"。
+	account.Credentials = map[string]any{
+		"model_mapping": map[string]any{"gpt-5.4": "gpt-5.4"},
+	}
+
+	require.True(t, account.IsModelSupported("gpt-5.6-sol"), "透传应放行不在残留白名单中的新模型")
+	require.True(t, account.IsModelSupported("deepseek-v4"), "透传应放行任意模型")
+}
+
 func TestIsModelSupported_OpenAIAPIKeyEmptyMappingAllowsAll(t *testing.T) {
 	account := &Account{
 		ID:       2,


HuntercodeT:

PR #4992, opened 2026-07-28
fix(cache): synchronize runtime invalidation across replicas

heyrry:
## Problem

Several hot-path caches are process-local while their source of truth is shared in PostgreSQL. In a multi-replica deployment, an admin write handled by replica A only invalidates replica A, so requests routed to replica B can continue using stale state.

Affected state includes:

- channel routing, model mappings, and channel pricing snapshots (10-minute TTL)
- gateway forwarding/runtime settings such as metadata passthrough and version bounds (60-second TTL)
- embedded frontend settings/ETag and CSP frame origins

This makes the same operation appear to work or fail depending on which replica receives the next request.

## Solution

- add a Redis Pub/Sub invalidation bus on `settings:cache:invalidate`
- subscribe every replica during router startup
- after a settings or channel write, invalidate the local process first and broadcast to peers
- clear runtime setting snapshots, eagerly rebuild the channel snapshot, invalidate embedded HTML/ETag state, and refresh CSP frame origins
- bound channel, runtime-setting, and embedded HTML caches to a 30-second TTL as a fallback if a Pub/Sub notification is missed
- keep local invalidation working and report an error if Redis publish/subscribe fails

The existing subscription-cache invalidation mechanism is unchanged.

## Verification

Added tests cover:

- invalidation broadcast reaching peer replicas
- local invalidation when publish fails
- subscribe failure reporting
- channel CRUD using the cross-replica callback
- expiration of all process-local settings snapshots
- HTML TTL fallback

```bash
go test ./internal/server ./internal/service ./internal/web
```

All listed packages pass against the current upstream `main`.

### DIFF
--- a/backend/internal/handler/admin/channel_handler.go
+++ b/backend/internal/handler/admin/channel_handler.go
@@ -25,6 +25,24 @@ func NewChannelHandler(channelService *service.ChannelService, billingService *s
 	return &ChannelHandler{channelService: channelService, billingService: billingService, pricingService: pricingService}
 }
 
+// InvalidateRuntimeCache clears the process-local channel routing snapshot.
+// The server calls this when another replica broadcasts a configuration write.
+func (h *ChannelHandler) InvalidateRuntimeCache() {
+	if h == nil || h.channelService == nil {
+		return
+	}
+	h.channelService.InvalidateRuntimeCache()
+}
+
+// SetOnUpdateCallback connects channel CRUD writes to the server's
+// cross-replica configuration invalidation broadcaster.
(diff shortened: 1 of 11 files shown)


heyrry:

heyrry:
recheck

heyrry:
Rebased onto the latest upstream `main` (`0b3fe95af`) and resolved the two conflicts while retaining the newer channel-monitor runtime listener logic.

Verification rerun:
```bash
go test ./internal/server ./internal/service ./internal/web
```
All listed packages pass. The PR is now mergeable. @Wei-Shaw, review would be appreciated when convenient.

PR #4999, opened 2026-07-28
fix(openai-ws): retire pooled connections before 60-minute limit

i0v0i-code:
## Summary

- retire idle `ctx_pool` connections before OpenAI's 60-minute WebSocket lifetime limit
- reserve the configured upstream read timeout plus a 30-second scheduling buffer before assigning a connection to a new turn
- check long-lived session leases before every turn
- transparently rotate non-strict/store-backed continuations; fail closed with WebSocket 1013 for strict `store=false` continuations before sending the turn upstream
- add pool and ingress regression coverage for near-expiry, pinned, stored, and strict-affinity sessions

## Problem

OpenAI Responses WebSocket connections have a hard 60-minute lifetime. The pool used the same 60-minute maximum age and only removed a connection after it was already over that age and idle. A connection acquired or still leased near minute 59 could therefore start another turn and be closed by the upstream in the middle of the response.

This patch distinguishes the hard connection age from the minimum lifetime needed to safely start a turn. With the default 15-minute upstream read timeout, pooled connections stop accepting new turns at approximately 44m30s instead of running directly into the upstream limit.

For strict `store=false` continuation affinity, the connection is not silently migrated. The client receives a retryable 1013 close before the next turn is sent upstream, avoiding partial output or duplicated side effects.

## Test plan

```bash
go test ./internal/service -count=1
```

Result on current `main`: pass (`96.243s`).

### DIFF
--- a/backend/internal/service/openai_ws_forwarder_ingress.go
+++ b/backend/internal/service/openai_ws_forwarder_ingress.go
@@ -1385,6 +1385,40 @@ func (s *OpenAIGatewayService) ProxyResponsesWebSocketFromClient(
 			}
 		}
 		forcePreferredConn := isStrictAffinityTurn(currentPayload)
+		if sessionLease != nil && !sessionLease.HasTurnLifetimeBudget(time.Now()) {
+			now := time.Now()
+			remainingLifetime := sessionLease.RemainingLifetime(now)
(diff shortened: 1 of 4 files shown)


i0v0i-code:
recheck

PR #4998, opened 2026-07-28
feat(deploy): make Redis persistence settings configurable

Ricardo-binZzz:
## Summary

- make the Redis RDB snapshot interval and change threshold configurable in all Compose deployment variants
- expose the Redis latency monitor threshold while preserving the current disabled default
- pass Compose Redis arguments directly instead of through `sh -c`, allowing the official entrypoint to drop privileges to the `redis` user
- propagate and validate the same persistence settings in the Apple Container workflow
- add CI fixtures for default and custom Redis Compose argv across production, local, and development files

## Compatibility

Existing deployments keep the current defaults: `save 60 1` and latency monitoring disabled (`0`). Empty `REDIS_PASSWORD` remains supported; Redis receives an empty `--requirepass` value and keeps authentication disabled. No application, PostgreSQL, Redis data format, or persistence format changes are introduced.

## Verification

- `git diff --check`
- Compose config validation for production, local, and development files
- default argv: `--save 60 1`, latency threshold `0`, empty password
- custom argv: `--save 900 2`, latency threshold `100`, configured password
- isolated `redis:8-alpine` runtime check: Redis PID 1 runs as UID 999 and empty-password `PING` returns `PONG`
- `bash -n deploy/apple-container.sh`

The Apple Container lifecycle and Compose argv fixtures are covered by repository CI.
- refreshed `v0.1.168` rebase: shell, backend unit/integration tests, golangci-lint, frontend, backend/frontend security, and CLA all pass

PR #5001, opened 2026-07-28
fix(ci): suppress embed build staticcheck false positive

Ricardo-binZzz:
## Summary

- suppress the two `SA4023` report locations with line-scoped golangci directives
- document why the result is not constant across build tags
- leave runtime behavior unchanged

## Context

`web.NewFrontendServer` has build-tag-specific implementations. The default `!embed` stub always returns an error, while the `embed` implementation can return either `nil` or an initialization error. Staticcheck sometimes analyzes the guarded block using only the `!embed` implementation and reports both the call site and `err != nil` comparison.

This surfaced in PR #4998 even though its GitHub merge ref has no differences from `main` under `backend/`; the same `main` SHA passed the push CI. Line-local suppressions are preferable to disabling `SA4023` globally.

## Verification

- `git diff --check`
- no production code or build-tag behavior changes
- repository CI verifies lint and all test suites

### DIFF
--- a/backend/internal/server/router.go
+++ b/backend/internal/server/router.go
@@ -71,7 +71,9 @@ func SetupRouter(
 
 	// Serve embedded frontend with settings injection if available
 	if web.HasEmbeddedFrontend() {
+		//nolint:staticcheck // Build-tag-specific implementations make this result non-constant.
 		frontendServer, err := web.NewFrontendServer(settingService)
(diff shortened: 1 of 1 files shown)


Ricardo-binZzz:
Superseded by the equivalent SA4023 suppression merged in #5003. Closing this duplicate; #4998 has been rebased onto the updated main branch.

PR #5006, opened 2026-07-28
fix: preserve admin weekly quota reset boundaries

zhsh2980:
## What changed

- preserve a manually reset weekly quota boundary for the full seven-day window
- advance expired weekly boundaries in seven-day periods instead of snapping them back to Monday
- use the same effective boundary for repository accounting, Redis quota checks, quota responses, and `window_resets_at`

## Why

`POST /api/v1/admin/users/:id/platform-quotas/reset` stores the reset time in `weekly_window_start`. The next billed request previously compared that value with `timezone.StartOfWeek(now)` and then always wrote the calendar-week start back to the row.

As a result, an admin reset outside Monday 00:00 could be shortened at the next calendar boundary. Weekly usage could also be cleared again on the next billed request, while cache eligibility checks and retry metadata continued to use the calendar week.

This change treats the persisted weekly boundary as the source of truth. Missing or future boundaries still fall back to the current calendar-week start, preserving the initialization behavior.

## Impact

After an admin weekly reset, subsequent billing and quota reads keep the reset timestamp and accumulated usage until seven days have elapsed. Weekly limit values are not changed. Daily and monthly quota behavior is unchanged.

## Validation

- `make test-unit`
- `go vet ./internal/pkg/timezone ./internal/handler/quotaview ./internal/repository`
- `go vet -tags unit ./internal/service`
- focused weekly-boundary service tests, repeated 10 times
- `git diff --check`

The PostgreSQL integration test was compiled with `-tags integration`, but the local integration harness skipped execution because Docker was unavailable. GitHub CI is expected to execute that test with its PostgreSQL test environment.

### DIFF
--- a/backend/internal/handler/quotaview/helpers.go
+++ b/backend/internal/handler/quotaview/helpers.go
@@ -13,7 +13,7 @@ import (
 // includeWindowStart=true 时输出 *_window_start 字段（admin 视角调试用）
 func LazyZeroQuotaForResponse(r service.UserPlatformQuotaRecord, now time.Time, includeWindowStart bool) map[string]any {
 	daily := buildWindowSlice(r.DailyUsageUSD, r.DailyLimitUSD, r.DailyWindowStart, NeedsDailyReset(r.DailyWindowStart, now), nextDailyResetTime(now), includeWindowStart)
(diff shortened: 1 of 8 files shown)


zhsh2980:

PR #5011, opened 2026-07-28
fix(vertex): avoid daily lockout for transient service-account 429s

CoffeeHi:
## Summary

- use the existing configurable seconds-level 429 fallback for resetless Vertex service-account `RESOURCE_EXHAUSTED` responses
- keep explicit upstream reset hints and daily-quota handling unchanged
- keep the existing Code Assist, Google One, API key, and AI Studio policies unchanged
- add regression coverage for configured cooldown, the default 5-second cooldown, disabled fallback, and explicit reset hints

## Why

`handleGeminiUpstreamError` currently has no `service_account` branch. When Vertex returns a transient 429 without `quotaResetDelay`, `Please retry in`, or daily-quota text, the account falls through to the API key / AI Studio fallback and is marked rate-limited until the next PST midnight.

That turns a temporary Vertex capacity or rate-limit response into an hours-long local lockout. The generic rate-limit service already has `rate_limit_429_cooldown_settings`, but the Gemini compatibility handler bypasses that fallback for 429 responses.

## Behavior after this change

- Vertex service account with an explicit reset hint: continue using the upstream reset time
- Vertex service account without a reset hint: use `rate_limit_429_cooldown_settings` (enabled with a 5-second default)
- disabled fallback setting: do not add a local rate-limit mark
- other Gemini account types: unchanged

## Relation to existing PRs

- #1254 changes the resetless fallback for Google One OAuth accounts.
- #2408 detects specific model-capacity error text.
- This PR is scoped to Vertex service accounts and covers any resetless transient 429 while reusing the existing administrator-configurable fallback.

## Testing

- `go test -tags=unit ./...`
- `go vet ./...`

All commands passed locally with Go 1.26.5.

### DIFF
--- a/backend/internal/service/gemini_error_policy_test.go
+++ b/backend/internal/service/gemini_error_policy_test.go
@@ -489,6 +489,97 @@ func TestHandleGeminiUpstreamError_GoogleOneCapacityExhaustedUsesTierCooldown(t 
 	require.True(t, repo.lastRateLimitReset.Before(after.Add(5*time.Minute).Add(2*time.Second)))
 }
 
+func TestHandleGeminiUpstreamError_VertexServiceAccountTransient429UsesConfiguredFallback(t *testing.T) {
+	repo := &rateLimit429AccountRepoStub{}
+	settingRepo := newMockSettingRepo()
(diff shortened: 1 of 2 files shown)


CoffeeHi:

Issue #4799, opened 2026-07-23
composite 分组不支持 /v1/messages Dispatch


wanqiu410:
如题：
claude code通过composite 分组接官方的gpt模型，报错

<img width="2354" height="836" alt="Image" src="https://github.com/user-attachments/assets/d5ac03b8-ebba-4414-8d94-583ae4c58ac5" />

<img width="1888" height="1258" alt="Image" src="https://github.com/user-attachments/assets/90a15344-5c1b-44fd-affe-c203bff5d8ba" />

缺少了OpenAI Messages 调度配置功能

<img width="1964" height="1550" alt="Image" src="https://github.com/user-attachments/assets/171146b3-5981-40b4-999b-04f759a48d46" />
尝试过修改数据库的配置，依然未能解决，最后放弃，转而通过CC使用OpenAI平台来接gpt模型

希望能在composite 分组中支持 /v1/messages Dispatch

MrHear:
同求

doubleDrake:
Confirmed on **v0.1.165** (and the same code is on current `main`). Sharing the root cause and a reproduction, since the OP found that editing the database directly did not help either — here is why.

## Root cause

The dispatch fields are not merely missing from the composite UI; they are **cleared server-side for every non-OpenAI platform**, which includes `composite`:

```go
// backend/internal/service/openai_messages_dispatch.go
func sanitizeGroupMessagesDispatchFields(g *Group) {
	if g == nil || g.Platform == PlatformOpenAI {
		return
	}
	g.AllowMessagesDispatch = false
	g.DefaultMappedModel = ""
	g.MessagesDispatchModelConfig = OpenAIMessagesDispatchModelConfig{}
}
```

It is called on both create and update (`backend/internal/service/admin_group.go:483` and `:819`). That is why a direct DB edit does not survive: any later write to the group resets the flag.

## Reproduction (API only, no UI)

1. `POST /api/v1/admin/groups` with `platform=composite` (note: `status` and `rate_multiplier` are required, otherwise it returns 500)
2. Add composite routes, e.g. `claude-opus-5 → anthropic`, `gpt-5.6-sol → openai`
3. `PUT /api/v1/admin/groups/{id}` with `{"allow_messages_dispatch": true}` → responds `code:0 success`
4. `GET /api/v1/admin/groups/{id}` → `allow_messages_dispatch` is `false` again (and `messages_dispatch_model_config` is back to `{}`)

[...]

PR #5013, opened 2026-07-28
feat(composite): add endpoint-aware default routing

### LINKED ISSUE #4813: Composite 平台类型分组 没有达到想要的效果
Composite 平台类型分组  类型还是没有达到想要的效果

想要的效果是  OpenAI分组，但是要 支持 允许 /v1/messages 调度

然后它再和 已有 Authropic 分组合并， 这样接到 claude code就能统一用 /v1/messages 调度

目前这个 Composite 是否可以增加 允许 OpenAI  /v1/messages 调度 能力呢

visa2:
## Summary

- Add an opt-in endpoint-default routing mode for Composite groups so protocol-specific requests can select a compatible upstream platform without one rule per model.
- Authorize the Composite → OpenAI Messages compatibility bridge from the resolved route instead of an OpenAI-only group switch, so `gpt-*` models work through an Anthropic-protocol client with no route configuration.
- Preserve explicit rule precedence, the built-in model detector, and all existing behavior when the flag is off.

Rebuilt on the latest `main` and squashed into a single commit. The group profit-control work landed after this branch was opened and moved the auth snapshot, the auth-cache invalidation trigger and the admin group DTO underneath it, so the original commit no longer applied cleanly. Generated `ent` code is regenerated from the schema rather than merged by hand.

## Routing precedence

1. Explicit exact/prefix rule (ranking unchanged: exact > prefix, then concrete endpoint > `any`, then longer prefix, then priority, then id)
2. Built-in model detector
3. Built-in endpoint default, when the group opted in

The endpoint default runs last on purpose. It infers a platform from the caller's protocol rather than from the model, so ordering it ahead of the detector would send a `gpt-*` request arriving on `/v1/messages` to Anthropic. Running it last also keeps the flag purely additive: it only produces a decision where resolution previously failed outright, so enabling it never retargets traffic that already worked.

The explicit-rule comparator is not modified by this change.

## Endpoint defaults

| Endpoint | Target platform |
| --- | --- |
| `messages`, `count_tokens` | Anthropic |
| `responses`, `chat_completions`, `embeddings` | OpenAI |
| `gemini` | Gemini |

Images are excluded because image models are cross-provider; that endpoint keeps its existing detector behavior. Video routing is unchanged.

Endpoint-default routing applies to the HTTP request paths. The Responses WebSocket receives its model only in the first frame and keeps its existing detector-only resolution — a pre-existing limitation this branch deliberately leaves untouched rather than wiring partially.

## OpenAI Messages authorization

`allowOpenAICompatibleMessagesDispatch` previously read `Group.AllowMessagesDispatch` for every platform except Grok. Composite groups fell into that branch and could never pass: `sanitizeGroupMessagesDispatchFields` clears that field for all non-OpenAI platforms on create and update, and the admin UI only renders the switch for OpenAI groups. The switch is therefore structurally always false for a Composite group, which is the permanent 403 reported in #4799 — including for users who set the column directly in the database.

Composite groups are now authorized by the resolved route instead:

| Group | Rule |
| --- | --- |
| Direct OpenAI | `allow_messages_dispatch` (unchanged) |
| Direct Grok | allowed (unchanged) |
| Composite → OpenAI | allowed when the route source is an explicit rule or the model detector |
| Composite → Grok | allowed |
| Composite → any other target | not eligible for the bridge |
| Missing group / other platforms | unchanged |

The endpoint default deliberately does not authorize the bridge: it selects a platform from the caller's protocol rather than from the model. `/v1/messages` and `/v1/messages/count_tokens` share one helper so the two endpoints cannot drift.

## Persistence and compatibility

- Migration `194` adds the column with `IF NOT EXISTS` and re-issues `enqueue_group_auth_cache_invalidation` from `193`'s body, preserving every profit-control and peak-window guard while adding the new field comparison. A regression test asserts that 194 keeps all of 193's predicates and adds exactly one.
- The API-key auth snapshot moves to `v19`, retaining the `v18` profit-control meaning; stale snapshots are refused and rebuilt.
- The flag is persisted through group create, update and duplicate, the repository `Select` list, the auth snapshot, the admin DTO and the admin UI. It is exposed on the admin group payload only, and forced to `false` for non-Composite platforms.
- Default is `false` for existing and newly created groups.

## Tests

- [x] `go test ./internal/service ./internal/handler ./internal/server/routes ./migrations -count=1`
- [x] `go test ./... -run '^$'`
- [x] `go vet ./...`
- [x] `vitest run src/views/admin/__tests__/groupsEndpointDefaultRouting.spec.ts`
- [x] `npm run build`

Behavioral coverage includes explicit-comparator preservation, the `any`-rule vs endpoint-default collision, detector precedence over the endpoint default, the flag being purely additive, images exclusion, flag-off equivalence with the previous resolver, every endpoint default, the full Messages authorization matrix, `/v1/messages` and `count_tokens` agreeing case for case, the snapshot version bump, and group create/update/duplicate propagation.

Fixes #4813
Refs #4799


heathermhuang (commented):
Reviewed against current main 5a6143097. The PR cherry-picks cleanly, and I found no blocking Composite routing issue in the diff. I verified the exact PR head and a rebased head with golangci-lint v2.9; both pass locally. The rebased head also passes the full Go unit and integration-tag suites plus frontend Vitest, typecheck, and build. The failed CI job only reported SA5011 in three unrelated pre-existing test files and does not reproduce. I cannot rerun the job without repository admin rights; please rebase or push a no-op to trigger a fresh run. This is strong source/test evidence for #4813, but #4799 should remain open until a real Claude Code /v1/messages and /v1/messages/count_tokens canary succeeds through an explicit Composite-to-OpenAI route.

doubleDrake:
这个 PR 正好能解决 #4799（composite 分组不支持 `/v1/messages` Dispatch）。补充一份实测数据，希望对推进合并有帮助。

## 实测（v0.1.165；v0.1.169 与 `main` 的相关代码相同）

用同一个 composite key（路由配置：`claude-opus-5 → anthropic`、`gpt-5.6-sol → openai`）：

| 客户端 | 协议格式 | Claude 模型 | GPT 模型 |
|---|---|---|---|
| Codex CLI | OpenAI（`/v1/responses`、`/v1/chat/completions`） | ✅ 正常 | ✅ 正常 |
| Claude Code | Anthropic（`/v1/messages`） | ✅ 正常 | ❌ `This group does not allow /v1/messages dispatch` |

结论：composite 在 **OpenAI 协议侧已经完全可用**，唯独 Anthropic 协议侧的 GPT 方向被挡住。

## 被挡的位置

composite 路由本身是对的 —— `POST /api/v1/admin/groups/{id}/composite-routes/preview` 带 `endpoint=messages` 能正确匹配到 `gpt-5.6-sol → openai`，只是随后被 dispatch 开关拒绝：

- `service/openai_messages_dispatch.go:104` `sanitizeGroupMessagesDispatchFields`：对非 openai 平台在**写入时**强制清空 `AllowMessagesDispatch` / `MessagesDispatchModelConfig`（这也是 #4799 里 OP 说「直接改数据库也没用」的原因）
- `handler/openai_gateway_handler.go:184` `allowOpenAICompatibleMessagesDispatch`：**运行时**只看 `Group.Platform`，composite 一律 false

## 为什么这个 PR 的做法合适

本 PR 描述里的

> Allow explicitly configured Composite OpenAI Messages routes to use the compatibility bridge while keeping implicit dispatch gated.

正好是「只放行显式配置的路由、隐式 dispatch 继续挡住」——既解决了这个场景，又不放松默认行为。按解析后的目标平台（composite 路由已经解析出来了）判断，而不是按分组平台判断，方向上也更一致。

## 相关 issue

同一个诉求（一个 key 在 Claude Code 里同时用 Claude 和 GPT）在这些 issue 里都出现过：#4799、#4813、#4422、#2039、#1942。

如果需要，我可以在合并后用上面的 composite 配置再跑一遍验证并回帖。

visa2:
Rebased onto v0.1.172 and renumbered the migration `194` → `196`: #5396 landed `194_add_usage_log_upstream_response_model.sql` and `195_add_usage_log_upstream_model_mismatch_index_notx.sql` in the meantime.

No functional change — only the migration filename and the corresponding test constant / function names. The trigger body still baselines on `193_group_profit_control_auth_cache_invalidation.sql`, which is still the latest migration touching `enqueue_group_auth_cache_invalidation()` (194/195 don't).

visa2:
Rebased onto v0.1.173 to resolve the conflicts with the Grok billing work.

Two things changed beyond the rebase itself:

1. **`apiKeyAuthSnapshotVersion` 19 → 20.** Both sides claimed 19 — upstream for the group `search`/`audio`/`video_model_prices` billing fields, this PR for the composite endpoint-default routing gate. Since v19 already shipped in v0.1.173, this PR now takes v20 so upgrading still forces a refresh of pre-v20 snapshots.

2. **Migration `196` → `221`.** `196_channel_monitor_v2_ignored_error_categories.sql` landed in the meantime; 221 is the next free number. The trigger body still baselines on `193_group_profit_control_auth_cache_invalidation.sql`, which remains the latest migration touching `enqueue_group_auth_cache_invalidation()` — nothing between 194 and 220 replaces it.

The conflicting `ent/mutation.go` and `ent/runtime/runtime.go` were not hand-merged: the whole `backend/ent/` tree was reset to the target branch and regenerated from the merged schema, so the only Group-related delta is the new `endpoint_default_routing_enabled` field plus the field-index shifts that come with it.

No behavioural change to the routing logic itself.

visa2:
@heathermhuang 你之前提到需要 rebase 或推一个空提交来触发新的 CI 运行。已经 rebase 到 v0.1.175,当前所有检查通过,包括 golangci-lint:

| Check | Result |
|---|---|
| test | pass |
| golangci-lint | pass |
| frontend | pass |
| frontend-security | pass |
| backend-security | pass |
| shell | pass |

PR 目前是 MERGEABLE / CLEAN。你此前观察到的 SA5011 未再出现。

PR #5019, opened 2026-07-28
fix(openai): preserve model-not-found semantics after failover

404QAQ:
## Summary

- preserve OpenAI-compatible `400/404 model_not_found` responses after failover exhaustion instead of rewriting them as a generic gateway error
- diagnose the no-account path as `404 model_not_found` when every model-supporting account is cooling down for the same deterministic upstream rejection
- scope the new HTTP 400 handling to OpenAI-compatible accounts so other platforms keep their existing behavior

This is the Luna/model-error portion split out of #3951.

## Why

GPT-5.6 model availability errors can arrive as HTTP 400. After all candidate accounts reject the model, the current handler loses the deterministic client error and returns a retryable 502/503, which causes unnecessary retries and account churn.

## Test plan

- [x] `go test -tags=unit ./internal/handler ./internal/service -count=1`
- [x] `go test -tags=integration ./internal/handler ./internal/service -run '^$' -count=1`
- [x] Verified non-OpenAI HTTP 400 model errors remain unaffected

### DIFF
--- a/backend/internal/handler/no_account_error.go
+++ b/backend/internal/handler/no_account_error.go
@@ -17,8 +17,10 @@ import (
 //
 //   - 404 model_not_found — the group has accounts, but none of them are
 //     configured to serve the requested model (config / typo / unsupported
-//     model). Returning 503 here misleads operators and trips reverse-proxy
-//     health checks; 404 lets the client surface the real problem.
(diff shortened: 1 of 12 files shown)


404QAQ:

PR #5020, opened 2026-07-28
fix(openai): classify non-streaming context overflow as bad request

404QAQ:
## Summary

- keep the terminal `response.failed` payload when converting non-streaming SSE responses
- classify deterministic context-window failures as HTTP 400 `invalid_request_error`
- preserve the same semantics for managed and passthrough responses, including compact keepalive responses whose HTTP status is already committed

This is the context-overflow portion split out of #3951.

## Why

A non-streaming upstream may return an SSE `response.failed` event for context overflow. The current SSE-to-JSON bridge rewrites it as a generic HTTP 502, hiding a client-correctable request error and encouraging pointless account failover.

## Test plan

- [x] Covered managed, passthrough, and already-committed compact keepalive paths

### DIFF
--- a/backend/internal/service/openai_gateway_context_window_nonstreaming_test.go
+++ b/backend/internal/service/openai_gateway_context_window_nonstreaming_test.go
@@ -1,0 +1,87 @@
+package service
+
+import (
+	"errors"
+	"io"
+	"net/http"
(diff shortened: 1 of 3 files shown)


404QAQ:

PR #5021, opened 2026-07-28
fix(openai): fail over pre-write response read errors

404QAQ:
## Summary

- treat non-streaming upstream body read failures as transport failures while no downstream bytes have been committed
- trigger account failover for managed and passthrough OpenAI responses
- avoid replay after downstream output, on client cancellation, or when the response exceeds the configured size limit

This is the response-read portion split out of #3951.

## Why

HTTP/2 stream resets can happen after upstream headers are received but before the non-streaming response body is read. The current path returns a hard 502 without trying another account even though the downstream response is still untouched.

## Test plan

- [x] Covered HTTP/2 read failure, passthrough, post-write, cancellation, and body-size-limit cases


404QAQ:

PR #5022, opened 2026-07-28
fix(openai): detach response affinity persistence

404QAQ:
## Summary

- persist response-to-account affinity with a context detached from request cancellation
- retain request-scoped values while replacing expired deadlines with a bounded 3-second timeout
- apply the helper consistently across HTTP responses, WS turns, prewarm responses, and sticky-account refreshes

This is the response-affinity portion split out of #3951.

## Why

Response affinity is written after the upstream turn completes. At that point the request context may already be canceled, causing the state-store write to fail and later `previous_response_id` requests to lose account affinity.

## Test plan

- [x] Verified canceled and already-expired parent contexts are replaced by a live bounded context


404QAQ:

PR #5023, opened 2026-07-28
fix(openai): preserve invalid JSON schema errors

404QAQ:
## Summary

- recognize the exact structured OpenAI `invalid_json_schema` error code
- return HTTP 400 `invalid_request_error` with the sanitized upstream message instead of a generic 502
- preserve the behavior across managed HTTP, passthrough HTTP, non-streaming SSE conversion, and compact keepalive output
- keep generic upstream 4xx responses on the existing sanitized error path

## Why

Schema validation failures are deterministic client errors. Rewriting them as HTTP 502 makes clients retry or rotate accounts even though the request must be corrected.

## Test plan

- [x] Covered top-level and nested `response.failed` envelopes plus managed and passthrough HTTP errors

### DIFF
--- a/backend/internal/service/openai_gateway_invalid_json_schema_test.go
+++ b/backend/internal/service/openai_gateway_invalid_json_schema_test.go
@@ -1,0 +1,139 @@
+package service
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net/http"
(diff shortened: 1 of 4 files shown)


404QAQ:

PR #5025, opened 2026-07-28
test: make nil guards explicit for staticcheck

404QAQ:
## Summary
- add explicit returns after six fatal nil guards across five test files
- satisfy staticcheck SA5011 without changing test behavior
- unblock lint checks for unrelated pull requests based on current main

## Validation
- targeted tests for all affected cases
- full tests for the affected packages
- unit-tag tests and integration-tag compile checks
- golangci-lint v2.9 with Go 1.26.5, cache cleared and same-issue limit disabled: 0 issues


404QAQ:
Closing this diagnostic PR after reproducing the CI behavior. The exact merge tree passes a clean Linux/amd64 golangci-lint v2.9 scan with Go 1.26.5, including a local main-cache-then-PR-cache reproduction. The SA5011 findings only appear with the existing GitHub Actions cache and are emitted in batches because max-same-issues defaults to 3. Adding returns to unrelated tests is therefore not the root fix; the upstream lint cache needs to be invalidated or disabled.

PR #5033, opened 2026-07-28
fix(deploy): prevent application privilege gains

Ricardo-binZzz:
## Summary

- enable `no-new-privileges` for the Sub2API application in all four Compose variants
- add a shell contract test that keeps the setting scoped to the application service
- run the contract in the existing deployment CI job

## Rationale

The application entrypoint starts as root only to repair `/app/data` ownership and then drops to the `sub2api` user. Without `no-new-privileges`, a compromised application process retains a path to gain privileges through future setuid/file-capability additions in the runtime image.

This change does not alter PostgreSQL or Redis, resource limits, networking, volumes, or application configuration.

## Verification

- current release image `0.1.168` completed an isolated `su-exec sub2api` transition with `no-new-privileges:true`
- production application container was recreated with this setting and verified healthy with PID 1 at UID 1000 and `NoNewPrivs: 1`
- all four Compose files pass `docker compose config --quiet`
- positive shell contract and missing-setting negative fixture pass
- `git diff --check`
- refreshed `v0.1.168` rebase: shell, backend unit/integration tests, golangci-lint, frontend, backend/frontend security, and CLA all pass

### DIFF
--- a/.github/workflows/backend-ci.yml
+++ b/.github/workflows/backend-ci.yml
@@ -16,6 +16,7 @@ jobs:
         run: |
           /bin/bash -n deploy/apple-container.sh
           /bin/bash deploy/tests/apple-container-test.sh
+          /bin/sh deploy/tests/docker-compose-security-test.sh
           /bin/sh deploy/test-caddyfile-cache.sh
 
   test:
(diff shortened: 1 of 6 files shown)


Ricardo-binZzz:
All checks except golangci-lint pass. The lint job reports only the same three pre-existing main-branch SA5011 diagnostics in payment_handler_test.go, proxyurl/parse_test.go, and openai_images_incomplete_test.go; this PR changes only Compose files, deployment CI wiring, and a shell contract test. The isolated baseline fix is #5027, after which this deployment hardening PR can be rebased for a fully green run.

PR #5039, opened 2026-07-29
feat(keys): add cURL/Python examples and connection testing

AgentsDancePro:
## What changed

- add dedicated cURL and Python tabs to the API key usage dialog
- prefill the group default mapped model and allow loading models visible to the current API key
- let users send a real smoke-test request and inspect HTTP status, latency, and the formatted response
- support OpenAI-compatible, Anthropic, Gemini, and Antigravity gateway shapes
- include copyable input examples and commented sample outputs
- add English and Chinese UI copy plus component coverage

## Why

Newly created API keys currently require users to configure a separate client before they can confirm that the key, group routing, model permissions, and quota are working. This makes that verification available directly in the existing "Use key" dialog.

The connection test intentionally goes through the normal gateway with the selected API key, so routing, billing, usage recording, and model authorization behave exactly like a regular request.

## Validation

- `pnpm typecheck`
- `pnpm exec eslint src/components/keys/UseKeyModal.vue src/components/keys/__tests__/UseKeyModal.spec.ts src/views/user/KeysView.vue src/i18n/locales/en/dashboard.ts src/i18n/locales/zh/dashboard.ts`
- `pnpm test --run src/components/keys/__tests__/UseKeyModal.spec.ts src/views/user/__tests__/KeysView.spec.ts`

Result: 2 test files and 22 tests passed.


AgentsDancePro:

PR #5068, opened 2026-07-29
fix(openai): preserve Lite tool call namespaces

### LINKED ISSUE #5067: [Bug] Responses Lite HTTP forwarding strips namespace from continuation history
## Summary

On the OpenAI OAuth HTTP `/v1/responses` path, a Responses Lite continuation can
reach the ChatGPT Codex upstream with two different identities for the same tool:

- the active declaration remains a native namespace under
  `input[].additional_tools.tools[]`; but
- a prior direct `function_call` for that tool loses its `namespace` before the
  first upstream request.

This is turn-dependent: the first namespaced collaboration call can succeed, while
the next request replays it as an unqualified history item.

This report is intentionally narrower than #4978. That issue covers namespace
flattening of the model-visible declaration. This issue covers the Responses Lite
continuation state where the declaration remains native but the matching history is
stripped.

## Environment

| Item | Value |
|---|---|
| sub2api | `main@5a6143097db142b72a6fc848c214e97214470bdd` |
| Selected upstream account | OpenAI OAuth subscription |
| Client endpoint | HTTP `POST /v1/responses` |
| Model | `gpt-5.6-sol` |
| Request mode | Responses Lite |
| Header | `X-OpenAI-Internal-Codex-Responses-Lite: true` |

The API key used by the client authenticates to the gateway; the selected upstream
account is still an official OpenAI OAuth subscription targeting the ChatGPT Codex
Responses backend. This is not an OpenAI Platform API-key upstream.

## Minimal continuation shape

```json
{
  "input": [
    {
      "type": "additional_tools",
      "role": "developer",
      "tools": [
        {
          "type": "namespace",
          "name": "collaboration",
          "tools": [
            {"type": "function", "name": "spawn_agent"},
            {"type": "function", "name": "wait_agent"}
          ]
        }
      ]
    },
    {
      "type": "function_call",
      "call_id": "call_1",
      "namespace": "collaboration",
      "name": "spawn_agent",
      "arguments": "{}"
    },
    {
      "type": "function_call_output",
      "call_id": "call_1",
      "output": "created"
    }
  ],
  "tool_choice": "auto"
}
```

## Observed provider-facing shape

The declaration remains namespaced:

```json
{"type":"namespace","name":"collaboration"}
```

but the historical call becomes unqualified:

```json
{"type":"function_call","name":"spawn_agent"}
```

A full `OpenAIGatewayService.Forward` regression using the HTTP upstream recorder
fails on the baseline with:

```text
expected: "collaboration"
actual:   ""
message:  the provider-facing declaration and prior function-call history must keep the same identity
```

## Static cause

The request passes through these transformations:

1. `normalizeOpenAIResponsesLiteToolsPayload`
2. `flattenOpenAIResponsesNamespaces`
3. `stripOpenAIResponsesInputNamespaces`

Lite normalization places namespace declarations under
`input[].additional_tools`. The namespace flattener examines top-level `tools`, so
it does not build a mapping for the nested declaration. The subsequent strip
deletes `namespace` from every direct input item while leaving nested declarations
unchanged. Later OAuth transforms do not restore the field.

## User-visible impact

The malformed continuation does not necessarily produce a simple local
unsupported-tool error. It changes the model-visible tool contract after the first
collaboration call. The observed symptom family includes later collaboration intent
being emitted as unrelated `exec` or exec-cell `wait` calls, including dummy cell
IDs.

OpenAI Codex has independent reports where the wrong tool identity is already
present in the raw model response after an earlier successful `spawn_agent`:

- [openai/codex#35620](https://github.com/openai/codex/issues/35620)
- [openai/codex#35541](https://github.com/openai/codex/issues/35541)

Those reports do **not** establish sub2api as their cause. They are included as
matching external symptoms and motivation for preserving a consistent tool
contract, not as proof of shared causality.

## Why the current behavior is understandable

This appears to be the composition of several individually reasonable compatibility
changes:

- #4154 added namespace flatten/restore for strict native Responses upstreams.
- #4324 moved Lite namespace declarations into `input.additional_tools`, because
  Responses Lite rejects namespace declarations at top-level `tools`.
- #4804 proactively removed direct `input[].namespace` fields after compact/public
  endpoints rejected long histories.

Together they leave the Lite declaration nested and namespaced while stripping its
matching direct history item. #4761 also documents the underlying contract split:
normal ChatGPT Codex Responses can require namespace round-tripping, while compact
rejects the same field. A global preserve or global strip rule is therefore unsafe.

## Expected behavior

For OpenAI OAuth HTTP non-compact Responses Lite requests, namespace identity should
remain consistent across:

- the nested declaration;
- `tool_choice`;
- historical `function_call` / `custom_tool_call` items; and
- returned calls.

Compact and standard/custom API-key endpoints should retain their stricter cleanup
behavior.

## Regression coverage

A fix should cover:

- managed and passthrough OAuth HTTP Lite;
- initial request and continuation history;
- `function_call` and `custom_tool_call` identity preservation;
- residual namespace cleanup on message/reasoning/output items;
- OAuth compact and API-key negative controls;
- OAuth non-Lite flatten/restore control;
- native WSv2 no-strip control; and
- provider-facing plus returned-call assertions through the complete `Forward`
  path.

A narrowly scoped PR with this coverage will follow.

X-T-E-R:
## Summary

- preserve `namespace` on direct historical `function_call` and
  `custom_tool_call` items for OpenAI OAuth HTTP, non-compact Responses Lite
  requests;
- keep the existing namespace cleanup for message, reasoning, output, unknown,
  compact, API-key/custom-provider, and non-Lite paths; and
- add end-to-end `OpenAIGatewayService.Forward` coverage for both the exact
  provider-facing request and the returned namespaced function call.

Closes #5067.

## Root cause

Responses Lite normalization moves native namespace declarations into
`input[].additional_tools.tools[]`. The existing flattener only builds its mapping
from top-level `tools`, while the following cleanup removes `namespace` from every
direct input item. The result is a split identity on continuation requests: the
active declaration stays namespaced, but its historical call becomes unqualified.

This change makes the cleanup context-aware at the proven boundary instead of
globally preserving or globally stripping namespaces.

## Behavior matrix

| Path | Result |
|---|---|
| OpenAI OAuth HTTP, non-compact, Responses Lite | Preserve namespace on direct function/custom call history |
| OpenAI OAuth HTTP, non-Lite | Keep current flatten/restore behavior |
| OpenAI OAuth compact | Keep stripping unsupported direct namespaces |
| OpenAI API-key/custom provider | Keep stripping unsupported direct namespaces |
| Native WSv2 transport | Unchanged; no HTTP cleanup |

## Why this is narrower than #4981

#4981 proposes preserving namespaces broadly for non-compact OAuth requests and
adds account-level configuration/UI. This PR does not change the default behavior
of non-Lite requests and adds no configuration surface. It only fixes the
Responses Lite continuation mismatch described in #5067, while retaining strict
cleanup everywhere else.

## Tests

Passed locally with Go 1.25-compatible unit tags:

```text
go test -tags unit ./internal/service -run '^(TestOpenAIGatewayServiceForward_NormalizesResponsesLiteToolsForOAuth|TestShouldFlattenOpenAIResponsesNamespaces|TestShouldStripOpenAIResponsesInputNamespaces|TestShouldPreserveOpenAIResponsesLiteToolCallNamespaces|TestStripOpenAIResponsesInputNamespaces.*|TestNormalizeOpenAIResponsesLiteToolsPayload_PreservesResponseCreateShape|TestApplyCodexOAuthTransform_PreservesLiteNamespaceToolChoice)$' -count=1

go test -json -tags unit ./internal/service -count=1

git diff --check
```

The regression suite covers managed and passthrough OAuth Lite requests,
continuation history, `function_call` and `custom_tool_call`, residual cleanup,
compact and API-key negative controls, the non-Lite control, native WSv2, and the
complete HTTP forwarding path.

## Scope and residual risk

- No frontend, database, or account configuration changes.
- The full service unit suite passes.
- The exact upstream HTTP body and downstream SSE call are asserted with a local
  recorder; a live ChatGPT Codex provider request was not used for this PR.

### DIFF
--- a/backend/internal/service/openai_gateway_forward.go
+++ b/backend/internal/service/openai_gateway_forward.go
@@ -73,7 +73,8 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 		}
 	}
 	if shouldStripOpenAIResponsesInputNamespaces(account, wsDecision.Transport, passthroughEnabled) {
-		body, err = stripOpenAIResponsesInputNamespaces(body)
+		preserveLiteToolCallNamespaces := shouldPreserveOpenAIResponsesLiteToolCallNamespaces(c, account, wsDecision.Transport, passthroughEnabled)
+		body, err = stripOpenAIResponsesInputNamespaces(body, preserveLiteToolCallNamespaces)
 		if err != nil {
 			setOpsUpstreamError(c, http.StatusBadRequest, err.Error(), "")
 			c.JSON(http.StatusBadRequest, gin.H{"error": gin.H{
--- a/backend/internal/service/openai_responses_lite_tools_test.go
+++ b/backend/internal/service/openai_responses_lite_tools_test.go
@@ -268,7 +268,7 @@ func TestOpenAIGatewayServiceForward_NormalizesResponsesLiteToolsForOAuth(t *tes
 				StatusCode: http.StatusOK,
 				Header:     http.Header{"Content-Type": []string{"text/event-stream"}},
 				Body: io.NopCloser(strings.NewReader(
-					"data: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp_lite\",\"usage\":{\"input_tokens\":1,\"output_tokens\":1}}}\n\n" +
+					"data: {\"type\":\"response.completed\",\"response\":{\"id\":\"resp_lite\",\"output\":[{\"type\":\"function_call\",\"namespace\":\"collaboration\",\"name\":\"spawn_agent\",\"call_id\":\"call_2\",\"arguments\":\"{}\"}],\"usage\":{\"input_tokens\":1,\"output_tokens\":1}}}\n\n" +
 						"data: [DONE]\n\n",
 				)),
 			}}
@@ -288,7 +288,10 @@ func TestOpenAIGatewayServiceForward_NormalizesResponsesLiteToolsForOAuth(t *tes
 					{"type":"tool_search"},
 					{"type":"namespace","name":"collaboration","tools":[{"type":"function","name":"spawn_agent","parameters":{"type":"object"}}]}
 				],
-				"input":[{"type":"message","role":"user","content":"hello"}],
+				"input":[
+					{"type":"function_call","namespace":"collaboration","name":"spawn_agent","call_id":"call_1","arguments":"{}"},
+					{"type":"message","role":"user","namespace":"residual-message-namespace","content":"hello"}
+				],
 				"tool_choice":{"type":"namespace","name":"collaboration"}
 			}`)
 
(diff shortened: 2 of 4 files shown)


X-T-E-R:

PR #5134, opened 2026-07-31
fix(openai): failover on in-stream error frames, add delayed status code switch

pcmid:
## Problem

When an OpenAI Responses upstream returns HTTP 200 and then reports the failure inside the SSE stream, the gateway cannot switch accounts. A real capture:

```
HTTP/2.0 200
content-type: text/event-stream

event: response.created
data: {"type":"response.created","response":{...,"status":"in_progress"}}

event: response.in_progress
data: {"type":"response.in_progress","response":{...}}

event: error
data: {"type":"error","error":{"type":"service_unavailable_error","code":"server_is_overloaded",
       "message":"Our servers are currently overloaded. Please try again later."}}

event: response.failed
data: {"type":"response.failed","response":{...,"status":"failed",
       "error":{"code":"server_is_overloaded",...}}}
```

All four frames arrive within the same second. The client gets a 200 plus an error stream instead of a successful retry on a healthy account.

## Root cause

| # | frame | `startsClientOutput` | effect |
|---|---|---|---|
| 1 | `response.created` | false (preamble) | buffered, not committed |
| 2 | `response.in_progress` | false (preamble) | buffered, not committed |
| 3 | `event: error` | **true** | written and flushed — HTTP 200 lands, `clientOutputStarted = true` |
| 4 | `response.failed` | — | `shouldFailover` is true, but the guard `!openAIStreamClientOutputStarted(...)` is already false → switch skipped |

Frame 3 is the culprit: `eventType` is read from the data JSON `type` field, and `"error"` is not in the preamble allow-list, so `openAIStreamDataStartsClientOutput` returns true and the frame is treated as normal client output.

The account-switch machinery itself is fine — `NextAccountAction`'s zero value is `NextAccountLegacyRetry`, so once the guard passes, the existing loop excludes the failed account and reselects. Nothing new is needed there.

## Changes

**1. Bare `type=error` frames are treated as failed terminal events** (unconditional)

They are no longer written out as client output, so the failover guard stays open. This mirrors the handling already present in `openai_gateway_messages.go`. The check is added at the call sites rather than in the shared terminal-event helpers (`isOpenAICompatResponsesTerminalEvent`, `openAIStreamEventTypeIsTerminal`), to avoid changing the passthrough `sawTerminalEvent` and stream-truncation semantics.

For chat/completions this also decouples the pre-output buffer from the silent-refusal 64KB threshold. Requests below that threshold previously flushed the preamble and committed HTTP 200, which made the handler reject the switch because the written size had changed — recognising the error frame without this decoupling would leave failover broken on that path.

**2. New per-group switch `delayed_status_code`** (default off)

Suppresses the keepalive heartbeat until the first semantic output. The heartbeat otherwise bypasses the buffer, commits HTTP 200 and closes the same failover window whenever the upstream stalls after `response.created`.

There is deliberately no timeout fallback: before the first output no bytes are sent at all, so a slow first token means a longer silent wait for the client and intermediate proxies may drop the connection on idle. That is why the switch defaults to off and is rolled out per group.

## Paths covered

| path | change 1 | change 2 |
|---|---|---|
| native Responses streaming | yes | yes |
| Responses passthrough streaming | yes | no keepalive on this path |
| chat/completions compat | yes | yes |
| messages compat | already implemented | yes |
| raw chat/completions | n/a — talks to the upstream chat endpoint, never sees Responses events | n/a |

## Behaviour change

Change 1 applies to all groups: **while nothing has been written to the client yet**, an upstream `event: error` is no longer forwarded and triggers an account switch instead. Streams that have already started output are untouched. When every account fails, the client still receives a real HTTP status code and JSON error body, not a 200 plus error stream.

Change 2 is off by default and must be enabled per group.

## Verification

```
go build ./...                                     ok
go vet ./internal/{service,repository,handler}/...  no findings
go test -tags=unit ./internal/service/... ./internal/repository/... ./internal/handler/...
                                                    7 packages ok
npm run build                                       ok
```

New unit tests cover: the four-frame capture above switching accounts; an `error` frame with no following `response.failed`; an `error` frame arriving after output has started (must still pass through); `content_policy` not switching; the chat/completions path; and keepalive behaviour with the switch on and off.

One existing case in `openai_gateway_response_flush_test.go` asserted the old behaviour (error frame forwarded verbatim, no error returned). It is split into two: before output → failover, after output → pass through unchanged.

End-to-end verified against a mock upstream covering nine overload shapes (in-stream `error`, in-stream `response.failed`, `slow_down`, non-streaming `status:failed`, HTTP 429/503/529), with a second healthy account as the switch target. All shapes now behave consistently.


pcmid:

PR #5135, opened 2026-07-31
feat(passthrough): add skip_failover to error passthrough rules

pcmid:
## Problem

The `/v1/responses` streaming paths evaluate error passthrough rules **before** the failover decision, while every other path does the opposite. An identical upstream failure therefore behaves differently depending on how the upstream reports it.

Verified against a mock upstream with a rule matching the overload message enabled (`keywords=["Our servers are currently overloaded"]`, `response_code=529`):

| upstream overload shape | status | switched account? | upstreams hit |
|---|---|---|---|
| in-stream `error` (server_is_overloaded) | **529** | **no** | `error_mock` |
| in-stream `response.failed` (same) | **529** | **no** | `failed_mock` |
| in-stream `error` (slow_down — message does not match the rule) | 200 | yes | `slow_down_mock` → `good` |
| non-streaming 200 + `status:failed` | 200 | yes | `json_failed_mock` → `good` |
| HTTP 429 | 200 | yes | `429_mock` → `good` |
| HTTP 503 (same overload message) | 200 | yes | `503_mock` → `good` |
| HTTP 529 | 200 | yes | `529_mock` → `good` |

Row 3 is the control: the same frame type switches accounts as soon as its message stops matching the rule. The streaming path can fail over — the rule was simply consulted first.

## Change

Rather than silently reordering the checks, this adds an explicit `skip_failover` column on error passthrough rules, defaulting to `false` (retry allowed):

- **false** — run failover first. The rule still rewrites the response after every account has failed, through the existing match in the exhaustion handler, so a configured status code is preserved.
- **true** — return the rule's response immediately, which is today's behaviour.

The decision is made **at the failover branch itself** rather than where the rule is applied, so it works uniformly on every platform:

```go
if s.shouldFailoverX(...) && !ruleSkipsFailover(c, platform, status, body) {
    return failover
}
// falls through to the path's existing error writer, where the rule rewrites
// the status code and message
```

Sixteen failover branches are covered:

| platform | branches |
|---|---|
| OpenAI | HTTP forward, Responses passthrough, Responses streaming, chat/completions (streaming + non-streaming), messages (streaming + non-streaming), images |
| Anthropic | main forward branch |
| Antigravity | claude, gemini, compat |
| Gemini | messages (×2), chat/completions |
| Grok | media |

Stream paths go through a thin wrapper that applies the same semantic-status inference and body normalisation as the rule matcher, so the two always agree on whether a rule matches.

`applyErrorPassthroughRule` keeps its original four return values — the skip decision is resolved before it runs, so it has no reason to report it.

## Behaviour change

The default flips existing behaviour for rules matching **retryable** errors: they now retry first and apply only once every account has failed. Set `skip_failover` on such a rule to keep returning immediately. Rules matching non-retryable errors (content policy, invalid request, context length) are unaffected — those never reached the failover branch to begin with.

## Verification

```
go build ./...                                     ok
go vet -tags=unit ./internal/service/               no findings
go test -tags=unit ./internal/service/... ./internal/repository/... ./internal/handler/...
                                                    7 packages ok
npm run build                                       ok
```

New unit tests cover the helper returning true/false per rule config, no match, platform mismatch, no rule service bound, and the stream wrapper agreeing with the matcher on semantic status; plus the Responses streaming path under both flag values and for a non-retryable error.

End-to-end re-run of the matrix after the change:

| shape | `skip_failover=false` | `skip_failover=true` |
|---|---|---|
| in-stream `error` | 200, switched | 529, not switched |
| in-stream `response.failed` | 200, switched | 529, not switched |
| HTTP 429 / 503 / 529 | 200, switched | 529, not switched |


pcmid:
recheck

PR #5149, opened 2026-07-31
feat(channel-monitor): retry group accounts on probe failure

asas6661:
## Summary

- resolve a channel monitor API key to its configured group
- retry scheduler-selected accounts after retryable or account-specific probe failures
- prefer an operational result while retaining the fastest degraded result as fallback
- support both OpenAI and Anthropic group-backed monitors

## Validation

This contribution branch was not rebuilt or retested locally at submission time. The behavior is already used in a production-derived deployment, while repository-specific validation is intentionally left to the upstream CI. Regression tests are included but were not executed locally for this submission.

## Scope

This PR contains only provider-neutral group probing and fallback behavior. It does not include deployment configuration, account data, pricing rules, private endpoints, or product-specific media adapters.


asas6661:

Issue #4949, opened 2026-07-27
Codex API-key providers lose web search for Responses Lite GPT-5.6 models


Lang-code-ops:
## Environment

- sub2api: current `main` (`f18f3143e` during reproduction)
- Codex CLI: `0.145.0`
- Client transport: custom API-key Responses provider pointing at sub2api
- Codex config: `web_search = "live"`
- Affected models: `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`

## Problem

The Codex model catalog marks the three GPT-5.6 models above with
`use_responses_lite=true`. In Responses Lite mode, Codex removes hosted
Responses tools and expects a client-side `web.run` extension instead.

For a custom API-key provider, the `web.run` extension is not installed. The
resulting request has no top-level hosted tools, and its `additional_tools`
entry contains only the normal client tools, with no `web.run`. Native web
search is therefore unavailable even though sub2api supports both hosted
`web_search` and `POST /v1/alpha/search`.

This is related to #3908, but the standalone search endpoint fix does not cover
the case where the Codex client never advertises or calls `web.run`.

## Reproduction evidence

With the bundled GPT-5.6 catalog entry (`use_responses_lite=true`):

```text
top-level tools: []
additional_tools: exec, wait, request_user_input, collaboration
web.run: absent
result: WEB_SEARCH_UNAVAILABLE
```

A direct request to sub2api's alpha search endpoint succeeds:

```text
POST /v1/alpha/search -> 200
response keys: encrypted_output, output, results
```

Using an otherwise identical temporary catalog with only the following changes:

```json
{
  "gpt-5.6-sol": { "use_responses_lite": false },
  "gpt-5.6-terra": { "use_responses_lite": false },
  "gpt-5.6-luna": { "use_responses_lite": false }
}
```

the same Codex command produces:

```text
top-level tools: custom, function, function, namespace, web_search
web_search lifecycle: started -> completed
result: sourced web result returned successfully
```

## Suggested behavior

For model manifests served to custom API-key Codex clients, advertise a mode
that matches tools actually available to those clients. One narrow workaround
is to normalize `use_responses_lite=false` for the three GPT-5.6 models when
serving the Codex models manifest.

If preserving Lite mode is preferred, the gateway/client contract needs another
way to guarantee that `web.run` is installed before hosted tools are suppressed.

There is also a related resilience issue: when the live ChatGPT models manifest
[...]

PR #5160, opened 2026-08-01
fix(openai): add opt-in Responses web-search bridge

chenty2333:
## Summary

- Add a default-off API-key account capability that translates Sub2API's existing standalone-search request into a standard Responses API request using the `web_search` tool.
- Route the translated request through the account's configured base URL, bearer token, model mapping, proxy, and explicit header overrides, without adding ChatGPT-only identity, session, version, or beta headers.
- Record `/v1/responses` as the actual upstream endpoint when the bridge is used, while retaining `/v1/alpha/search` as the inbound endpoint.
- Expose the capability in create, edit, and bulk-edit account flows.
- Generate custom-provider configuration with the `sub2api` / `Sub2API` identity instead of labeling a third-party base URL as the built-in OpenAI provider.

## Contract and scope

The bridge is an explicit compatibility adapter:

- Input: Sub2API's existing `/v1/alpha/search` gateway request.
- Upstream: the API-key account's `/v1/responses` endpoint with `stream: true`, `store: false`, and `tools: [{"type":"web_search"}]`.
- Output: the existing standalone-search response adapter, with usage accounting tied to the actual upstream endpoint.

The account capability is disabled by default. When it is absent or false, API-key standalone-search forwarding is unchanged. OAuth and personal-access-token behavior is unchanged.

This PR does not depend on a particular client version or model. It does not change model manifests, namespace-tool transforms, Responses Lite selection, multi-agent routing, or compaction behavior. The generated config change only corrects custom-provider identity; it does not declare undocumented client capabilities.

Related to #4949; this is a server-side compatibility mechanism, not a claim that all client-side search discovery cases are fixed.

## Tests

- `make test-unit`
- `TESTCONTAINERS_RYUK_DISABLED=true make test-integration`
- `make test-frontend`
- 90 focused Vitest cases for account forms and generated provider config
- `pnpm build`

The local integration command disables the testcontainers reaper because the local Docker daemon had a stale reaper in `removing` state; the complete integration suite passed with normal per-test container cleanup.

### DIFF
--- a/backend/internal/handler/endpoint_test.go
+++ b/backend/internal/handler/endpoint_test.go
@@ -219,6 +219,12 @@ func TestResolveOpenAIUpstreamEndpointPrefersForwardResult(t *testing.T) {
 			result:  &service.OpenAIForwardResult{},
 			want:    EndpointResponses,
 		},
+		{
+			name:    "standalone search bridge reports responses",
+			account: &service.Account{Platform: service.PlatformOpenAI, Type: service.AccountTypeAPIKey},
+			result:  &service.OpenAIForwardResult{UpstreamEndpoint: EndpointResponses},
+			want:    EndpointResponses,
+		},
 	}
 
 	for _, tt := range tests {
--- a/backend/internal/handler/openai_alpha_search.go
+++ b/backend/internal/handler/openai_alpha_search.go
@@ -236,7 +236,7 @@ func (h *OpenAIGatewayHandler) recordAlphaSearchUsage(
 	sessionID := service.ExtractClientSessionID(c)
 	requestPayloadHash := service.HashUsageRequestPayload(body)
 	inboundEndpoint := GetInboundEndpoint(c)
-	upstreamEndpoint := GetUpstreamEndpoint(c, account.Platform)
+	upstreamEndpoint := resolveOpenAIUpstreamEndpoint(c, account, result)
 	quotaPlatform := service.QuotaPlatform(c.Request.Context(), apiKey)
 
 	h.submitMandatoryUsageRecordTask(c.Request.Context(), func(ctx context.Context) {
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -1860,6 +1860,20 @@ func (a *Account) IsOpenAIResponsesFlattenNamespacesEnabled() bool {
 	return ok && enabled
 }
 
+// IsOpenAIAlphaSearchResponsesFallbackEnabled 返回 API Key 账号是否将网关的
+// standalone search 请求转换为标准 Responses API web_search 请求。
+// 字段：accounts.extra.openai_alpha_search_responses_fallback，缺省 false。
+//
+// 该兼容开关只适用于支持 Responses API web_search、但没有兼容 standalone
+// search 端点的 OpenAI 兼容上游。
+func (a *Account) IsOpenAIAlphaSearchResponsesFallbackEnabled() bool {
+	if a == nil || !a.IsOpenAIApiKey() || a.Extra == nil {
+		return false
+	}
+	enabled, ok := a.Extra["openai_alpha_search_responses_fallback"].(bool)
+	return ok && enabled
+}
+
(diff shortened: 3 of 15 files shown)


Monika-shipship:
## Additional reproduction: Grok OAuth pool and native Responses web_search

This is related to this PR, but differs from #5343. The affected upstream is Sub2API's native Grok OAuth account pool, not an OpenAI-compatible API-key account.

### Environment

- Sub2API: `weishaw/sub2api:0.1.176`
- Deployment: Docker on Windows
- Client: Grok Build `1.0.3`
- Account platform/type: Grok OAuth
- Downstream endpoint: local OpenAI-compatible `/v1`
- Model: `grok-4.5` (also observed with the gateway's `grok-4.6`)
- Grok Build custom model configuration:
  - `api_backend = "responses"`
  - `supports_backend_search = true`
  - `[models].web_search = "grok-4.5"`

No credentials or account identifiers are included.

### Reproduction

Using the same local Sub2API API key:

1. Normal Chat Completions request succeeds:
   `POST /v1/chat/completions` -> HTTP 200.
2. Plain Responses request succeeds:
   `POST /v1/responses` -> HTTP 200.
3. The official xAI Responses request with the current search tool times out:

```json
{
  "model": "grok-4.5",
  "input": "Search the live web for the current Codex version.",
  "tools": [{"type": "web_search"}],
  "stream": false
}
```

Observed result: no response after 90 seconds. Sub2API logs show repeated Grok upstream failover attempts with HTTP 502, followed by a gateway timeout/503.

4. The legacy compatibility form succeeds:

```json
{
  "model": "grok-4.5",
  "input": "Search the live web for the current Codex version.",
  "tools": [{"type": "web_search_preview"}],
  "stream": false
}
```

Observed result: HTTP 200, with a completed response and a source URL.

5. Sub2API's standalone search route also works when `max_results` is supplied:

```json
{"query": "xAI Grok", "max_results": 5}
```

`POST /v1/web_search` returns HTTP 200 with `provider: "grok-native"` and a non-empty `results[]` array. The admin test page currently reports a false negative because it counts Responses `tool_calls`/`sources` rather than the standalone top-level `results[]`.

### Expected behavior

A Grok Build request using the official Responses tool:

```json
"tools": [{"type": "web_search"}]
```

should be forwarded to the Grok OAuth account pool and return a completed Responses search result, including the expected `web_search_call` / source data.

### Suggested direction

Please check the Grok OAuth Responses forwarding path for the newer `web_search` tool type. The legacy `web_search_preview` path works through the same gateway and account pool, so this appears to be a protocol-specific compatibility or request-normalization issue rather than an account-health issue.

Also consider updating the standalone Grok search probe to treat a successful non-empty top-level `results[]` response from `/v1/web_search` as success.

### Evidence summary

- Normal chat: HTTP 200
- Plain Responses: HTTP 200
- Responses + `web_search_preview`: HTTP 200
- Responses + official `web_search`: reproducible timeout
- Standalone `/v1/web_search` with `max_results`: HTTP 200 with results

Monika-shipship:
Correction and stronger reproduction from the same Grok OAuth deployment on v0.1.176:

- A local one-line experiment allowed `web_search_preview` through the Grok Responses sanitizer. The upstream `cli-chat-proxy.grok.com` then returned HTTP 422 and explicitly said `web_search_preview` is unknown; `web_search` is in its supported variant list.
- A direct Responses request using `tools:[{"type":"web_search"}]`, `tool_choice:"required"`, and `include:["web_search_call.action.sources"]` returned HTTP 200 after about 39s, but `output[]` contained only `message`: zero `web_search_call`, zero source annotations.
- A forced Grok Build 1.0.3 run with only built-in `web_search` visible also ended with `server_tool_use.web_search_requests=0`.

Therefore the earlier HTTP 200 preview result with URL-shaped text was not proof of a server-side search. The remaining compatibility gap is that the Grok OAuth CLI upstream does not return the standard Responses search event/source shape required by Grok Build. The local experiment has been rolled back to the official image; no database or account data was changed.

PR #5163, opened 2026-08-01
fix(openai): safely retry websocket capacity failures

margetrp-hub:
## Background

OpenAI Responses WebSocket can report a transient capacity failure through `response.failed` after the handshake succeeds. The previous flow forwarded that terminal event directly, bypassing the existing account-pool retry and failover path and exposing the upstream capacity message to clients.

## Changes

- Convert transient capacity `response.failed` events into the existing failover signal only on the first turn and before text or tool output starts.
- In pool mode, retry the same account up to `pool_mode_retry_count`, then exclude it and select another account.
- When all attempts fail, return only a generic `503 Upstream service temporarily unavailable` response.
- Once text/tool output or a later turn has started, do not replay the request; close with generic `1013 Try Again Later` to avoid duplicate output and side effects.
- Cover ingress, WS v2, and WS v2 passthrough paths.

## Scope

- This PR only changes Responses WebSocket behavior; regular HTTP `/v1/responses` behavior is unchanged.
- Passthrough still forwards `response.created` immediately to preserve prompt cancellation, so a safe retry can produce another `response.created` prelude.

## Tests

- Added focused regression coverage for retry/failover, message sanitization, tool-output replay prevention, and later-turn replay prevention.

### DIFF
--- a/backend/internal/handler/openai_gateway_handler.go
+++ b/backend/internal/handler/openai_gateway_handler.go
@@ -1668,6 +1668,7 @@ func (h *OpenAIGatewayHandler) ResponsesWebSocket(c *gin.Context) {
 	maxAccountSwitches := h.maxAccountSwitches
 	switchCount := 0
 	failedAccountIDs := make(map[int64]struct{})
+	sameAccountRetryCount := make(map[int64]int)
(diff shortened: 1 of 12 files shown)


margetrp-hub:

PR #5182, opened 2026-08-02
fix(scheduler): stop work after request cancellation

rick147:
## Summary

- stop scheduler snapshot and account lookup work when the request context is canceled
- avoid reporting expected caller cancellation as cache read or publish-token errors
- skip publish-token acquisition, database fallback, and cache publication after the caller is gone
- preserve the existing database fallback behavior for genuine cache failures

## Root cause

The scheduler continued using the same canceled request context after a cache read returned. That produced a second publish-token error and could continue into unnecessary database fallback work.

This path is used by the Codex `/models` account selection flow and is unrelated to the OpenAI quota-reset changes in #5045.

## Testing

- `go test -tags=unit ./internal/service -run 'TestSchedulerSnapshot(List|GetAccount)StopsAfterRequestCancellation|TestSchedulerFallbackReturnsDBAccountsWhenBucketRetired' -count=1`
- `git diff --check`

### DIFF
--- a/backend/internal/service/scheduler_snapshot_cancellation_test.go
+++ b/backend/internal/service/scheduler_snapshot_cancellation_test.go
@@ -1,0 +1,79 @@
+//go:build unit
+
+package service
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type schedulerCancellationCache struct {
+	SchedulerCache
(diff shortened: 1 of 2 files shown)


rick147:
Superseded by #5183, which carries this fix on the requested `codex/feat-openai-reset-credit-cache` branch.

PR #5200, opened 2026-08-02
fix(auth): prevent refresh token races across tabs

mrlitong:
## Summary

- route proactive refreshes and Axios 401 recovery through one same-document single-flight operation
- serialize one-time refresh-token rotation across tabs with Web Locks, with bounded peer reconciliation when Web Locks are unavailable
- adopt only a complete token pair from the same signed-in user
- prevent stale requests from clearing a session that logged out or switched accounts while refresh was in flight
- add the concurrency regressions to the critical frontend test target

## Root cause

The proactive refresh timer and the 401 response interceptor used separate refresh paths. They could submit the same one-time refresh token concurrently, particularly with multiple tabs or after a background tab resumed. The winning request rotated the token, while the losing request treated the old token as invalid and cleared the shared browser session, including the winner's replacement tokens.

## Validation

- frontend ESLint
- vue-tsc --noEmit
- 8 critical Vitest files, 129 tests passed
- production frontend build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 .PHONY: build build-backend build-frontend test test-backend test-frontend test-frontend-critical
 
 FRONTEND_CRITICAL_VITEST := \
+	src/api/__tests__/client.spec.ts \
+	src/api/__tests__/tokenRefresh.spec.ts \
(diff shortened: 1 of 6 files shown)


mrlitong:

PR #5270, opened 2026-08-04
fix(openai): normalize strict Responses stream fields

XiXi-desu:
Supersedes #5269. This PR is resubmitted with the commit identity linked to the GitHub account for CLA verification.

## Summary

- Normalize Responses API message and output_text fields required by strict clients.
- Keep message IDs stable across output_item.added, output_item.done, and response.completed.
- Apply the same compatibility normalization to converted and passthrough streaming/non-streaming responses while preserving unknown output item fields.

## Root Cause

Some response.completed payloads serialized output_text parts without annotations or logprobs, and message items without a stable id/status. The passthrough path also bypassed the compatibility serializer. Strict clients such as Grok deserialize these fields as required and failed with missing field annotations.

## Tests

- go test ./internal/pkg/apicompat -count=1
- go test ./internal/service -count=1
- git diff --check

### DIFF
--- a/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
+++ b/backend/internal/pkg/apicompat/anthropic_to_responses_response.go
@@ -632,3 +632,11 @@ func generateItemID() string {
 	_, _ = rand.Read(b)
 	return "item_" + hex.EncodeToString(b)
 }
+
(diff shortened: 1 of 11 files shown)


XiXi-desu:

PR #5287, opened 2026-08-05
[codex] suppress unchanged prompt config reload logs

zvensmoluya:
## Summary

- keep the five-second fallback configuration refresh unchanged
- emit `prompt_guard.config_loaded` only for the initial load, config-version changes, global risk-control changes, or recovery from a load error
- add regression coverage for unchanged refreshes and each observable transition

## Root cause

`ConfigManager.refreshLoop` calls `Reload` every five seconds as a bounded fallback when Redis invalidation is unavailable or missed. `Reload` emitted `prompt_guard.config_loaded` after every successful call, even when the active configuration had not changed.

A production instance emitted 131,208 identical INFO entries in about 7.6 days. They accounted for roughly 66% of the application log.

## Impact

Unchanged fallback refreshes now stay silent while meaningful configuration activation and recovery events remain observable. Runtime refresh, fail-closed behavior, Redis invalidation, and snapshot semantics are unchanged.

## Validation

- `gofmt` on changed Go files
- `git diff --check`
- targeted unit coverage added; CI is expected to run the repository test suite

### DIFF
--- a/backend/internal/securityaudit/prompt_config_store.go
+++ b/backend/internal/securityaudit/prompt_config_store.go
@@ -134,14 +134,27 @@ func (m *ConfigManager) Reload(ctx context.Context) error {
 	}
 	now := m.clock.Now()
 	previous := m.snapshot.Load()
+	recovered := m.hasLoadError()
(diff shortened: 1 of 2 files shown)


zvensmoluya:

PR #5288, opened 2026-08-05
fix(ccswitch): default OpenAI imports to GPT-5.6 Terra

### LINKED ISSUE #5238: CC switch 快捷导入 默认模型 应更新为 gpt-5.6
RT

yan9651688:
## Summary

- update OpenAI/Codex CC Switch imports from GPT-5.5 to GPT-5.6 Terra
- keep Anthropic, Gemini, Antigravity, and Grok import behavior unchanged
- keep the default pinned by the existing CC Switch import test

Terra is the balanced GPT-5.6 variant, so an imported provider is immediately usable with a general-purpose model without adding a model-selection step.

Fixes #5238

## Risk

Narrow frontend-only default-value change. No API, schema, backend scheduling, or UI changes.

## Validation

- `pnpm --dir frontend exec vitest run src/utils/__tests__/ccswitchImport.spec.ts` (10/10)
- `pnpm --dir frontend lint:check`
- `pnpm --dir frontend typecheck`
- `pnpm --dir frontend build`

The full frontend suite was also checked against the same base: 206/207 files and 1457/1459 tests passed. The two failures are existing rollback API assertions, and the run also reports existing GroupsView mock rejections; neither affected source or tests in this change.

### DIFF
--- a/frontend/src/utils/__tests__/ccswitchImport.spec.ts
+++ b/frontend/src/utils/__tests__/ccswitchImport.spec.ts
@@ -13,7 +13,7 @@ function paramsFromDeeplink(deeplink: string): URLSearchParams {
 
 describe('ccswitchImport utils', () => {
   it('defaults OpenAI CC Switch imports to the current Codex model', () => {
-    expect(OPENAI_CC_SWITCH_CODEX_MODEL).toBe('gpt-5.5')
+    expect(OPENAI_CC_SWITCH_CODEX_MODEL).toBe('gpt-5.6-terra')
   })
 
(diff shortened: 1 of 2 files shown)


yan9651688:
Hi @Wei-Shaw, this is a small frontend-only fix for #5238: it updates the CC Switch OpenAI/Codex default from GPT-5.5 to the balanced `gpt-5.6-terra` variant and adjusts the existing regression test. All CI and CLA checks are passing, and the PR is currently mergeable. When you have a moment, could you please review and merge it if the approach looks good? Thank you!

PR #5345, opened 2026-08-06
fix(security): block OAuth account takeover via pending exchange

puppywang:
## Summary

Fixes a **critical account-takeover vulnerability (0day)** in the pending OAuth flow: an attacker who only knows a victim's email address can bind their own LinuxDo / OIDC / WeChat / DingTalk identity to the victim's account and log in as them.

## Attack chain

1. Attacker logs in with their own OAuth account (`intent=login`) and lands on the email-completion step.
2. Attacker submits the **victim's email** via `POST /auth/oauth/pending/create-account` with an arbitrary password. Because the email already exists, the handler transitions the pending session into the choice state with `TargetUserID = victim`, `existing_account_bindable = true` — **no password, no email verification code, no captcha, no proof of ownership** is ever checked on this path (the captcha check is skipped when the email exists).
3. Attacker calls `POST /auth/oauth/pending/exchange` with an adoption decision (`adopt_display_name` / `adopt_avatar` — pure booleans unrelated to account ownership). `ExchangePendingOAuthCompletion` only guarded against `email_completion` / `bind_login_required` steps, so the request falls through to `applyPendingOAuthAdoption` → `applyPendingOAuthBinding`, and `shouldBindPendingOAuthIdentity` returns `true` unconditionally for `intent=login`.
4. The attacker's OAuth identity is now bound to the victim's account. On the next OAuth login, `findOAuthIdentityUser` resolves the attacker to the victim and the exchange endpoint issues the **victim's token pair**.

## Fix

`ExchangePendingOAuthCompletion` now only performs adoption/binding for:

1. **Terminal login sessions** (`canIssueTokenPair == true`) — the identity was already safely bound to that user during a completed login;
2. **`bind_current_user` sessions** — initiated by an authenticated user whose binding target comes from the login-session cookie.

Every other state (including `choose_account_action_required`) returns the payload **without binding the identity and without consuming the session**, so the attacker cannot proceed to the next step either.

## Testing

- New regression test `TestExchangePendingOAuthCompletionChoiceStateDoesNotBindIdentity` reproduces the exact attack (choice-state session + adoption decision → identity not bound, session not consumed, no token issued).
- All existing `TestExchangePendingOAuthCompletion*`, `TestCreateOIDCOAuthAccount*`, `TestSendPendingOAuthVerifyCode*`, and the full `internal/handler` package tests pass.

## Recommendation for follow-up (not included)

Defense in depth: when `create-account` / `send-verify-code` discovers an existing email, the session should transition to a **password/2FA-required** state (`bind_login_required` semantics) instead of a choice state that implies bindability. Please also audit existing `auth_identities` rows for bindings whose upstream email does not match the owning user.

### DIFF
--- a/backend/internal/handler/auth_oauth_pending_flow.go
+++ b/backend/internal/handler/auth_oauth_pending_flow.go
@@ -1998,6 +1998,21 @@ func (h *AuthHandler) ExchangePendingOAuthCompletion(c *gin.Context) {
 		response.Success(c, payload)
 		return
 	}
+	// ─── 安全修复（账号接管 0day）────────────────────────────────────────────
+	// 非终态 session（如 choose_account_action_required）的 TargetUserID 可能来自
+	// 攻击者提交的他人邮箱：createPendingOAuthAccount / SendPendingOAuthVerifyCode
+	// 发现邮箱已存在时会把本 session 指向该邮箱用户，全程无密码、无邮箱验证码、
+	// 无账号所有权证明。若此时带着 adoption decision 继续执行，下方的
+	// applyPendingOAuthAdoption 会把本 OAuth identity 直接绑定到 TargetUserID，
+	// 攻击者随后再次 OAuth 登录即被系统识别为受害者本人（完整账号接管）。
+	// 只有两类 session 允许在此处执行 adoption/binding：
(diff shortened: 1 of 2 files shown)


puppywang:

github-actions[bot] ***@***.***> 于2026年8月7日周五 03:41写道：

puppywang:
recheck

PR #5355, opened 2026-08-07
fix(openai): fail over streamed capacity error events

cyhhao:
## Summary

Codex Responses can emit an SSE `event: error` with `type: error` and `server_is_overloaded` before the terminal `response.failed` event. The messages bridge previously passed that event through, so the client saw `Selected model is at capacity` and the failover loop never got a chance to switch accounts.

This change recognizes that pre-output capacity event in both streaming paths and returns the existing failover error before any semantic SSE bytes are committed. Configured error passthrough rules are evaluated first, and deterministic context-window errors remain non-failover. Existing same-account retry and account-switch limits are unchanged.

WebSocket handling is intentionally unchanged.

## Tests

- `go test ./internal/service -run 'TestOpenAIStreaming(ResponseFailedBeforeOutputCapacityErrorReturnsFailover|CapacityErrorEventHonorsPassthroughRule|PassthroughCapacityErrorEventBeforeFailedReturnsFailover|PassthroughCapacityErrorEventHonorsPassthroughRule)$' -count=1`

The handler regression exercises three OAuth accounts, verifies the first two capacity-shed accounts are retried according to the existing policy, and confirms the healthy account completes without exposing the capacity error to the client.

### DIFF
--- a/backend/internal/handler/openai_gateway_handler_test.go
+++ b/backend/internal/handler/openai_gateway_handler_test.go
@@ -1846,6 +1846,61 @@ func (u *openAIHTTPPassthroughSSERateLimitUpstream) calls() []int64 {
 	return append([]int64(nil), u.accountIDs...)
 }
 
+type openAIHTTPOAuthCapacityFailoverUpstream struct {
+	service.HTTPUpstream
+	mu         sync.Mutex
+	accountIDs []int64
+	failedIDs  map[int64]struct{}
+}
+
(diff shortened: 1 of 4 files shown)


cyhhao:
@codex review

cyhhao:
Superseded by #5398, which fixes the same pre-output Codex capacity-shed failover issue and adds client-retryable error rewriting for post-output and WS-HTTP bridge paths. Closing this duplicate. Thank you.

PR #5357, opened 2026-08-07
fix: fail over Anthropic transport errors

yunhe-dev:
Routes Anthropic transport-level upstream failures through account failover. Persistent connection, proxy, and network failures are temporarily unscheduled; transient timeouts and resets fail over without eviction. Added focused tests and verified the Docker build.


yunhe-dev:

PR #5406, opened 2026-08-07
fix(openai): forward OAuth routing hints

bestony:
## Summary

Forward gateway-derived OpenAI OAuth routing hints across the Responses HTTP and WebSocket paths, stop negotiating the obsolete OAuth Responses beta token, and make WebSocket connection reuse honor routing hints as soft dial affinity.

This is the upstream-targeted counterpart of [bestony/sub2api#68](https://github.com/bestony/sub2api/pull/68). Related Codex behavior is tracked in [openai/codex#37345](https://github.com/openai/codex/issues/37345).

## Problem

OpenAI OAuth requests can have their effective model or service tier changed by model mapping, OAuth request transformation, account policy, or Fast-tier normalization. The gateway previously did not consistently send routing metadata derived from that final upstream payload. It could also forward caller/account-supplied routing hints and synthesize the legacy `OpenAI-Beta: responses=experimental` token on OAuth HTTP requests.

The WebSocket pool additionally treated all handshake metadata as hard compatibility, which is too strict for routing hints: a continuation must keep its selected connection even when the requested hint changes, while ordinary reuse should prefer a connection dialed with the same hint and still degrade to a hard-compatible connection when capacity is full.

## Wire contract

- `x-codex-routing-hint` is gateway-owned. Every case-insensitive caller or account-override spelling is removed before generation.
- Hints are generated only for OpenAI OAuth accounts. API-key and provider-credential paths sanitize the header but do not generate one.
- The hint uses the final mapped/transformed model and final normalized service tier.
- `fast` is emitted as `priority`; `priority` and `flex` are retained.
- `default`, omitted, and unsupported tiers emit a model-only hint.
- Empty models, invalid HTTP header values, and models containing routing delimiters are omitted.
- OAuth HTTP no longer synthesizes `OpenAI-Beta: responses=experimental`.
- OAuth passthrough removes only that exact comma-delimited beta token and preserves independent beta tokens. API-key beta headers remain caller-controlled.

### HTTP and compact

- Generate the routing hint after account header overrides for ordinary Responses HTTP, passthrough, and compact requests.
- Preserve `service_tier` during compact request normalization.
- Add final upstream request capture tests covering model mapping, `fast -> priority`, `priority`, `flex`, `default`, invalid values, OAuth-only behavior, spoofed header casing, and exact legacy beta-token removal.

### WebSocket transport

- Pass the final model and tier into OAuth WebSocket handshake construction for WSv2, ingress first and subsequent turns, passthrough, and `generate=false` prewarm dials.
- Keep beta features as hard handshake compatibility while treating the routing hint as soft affinity.
- Preserve continuation and explicitly selected connection reuse when the requested hint changes.
- Prefer same-hint connections for ordinary reuse, but queue on a hard-compatible different-hint connection when the pool is at capacity.
- Keep routing-affinity extraction deterministic when a raw header map contains multiple case variants.

### Pool lifecycle and diagnostics

- Wake topology waiters after successful dials and account clears.
- Prevent canceled waiters from consuming a lease.
- Update the prewarm target only after a successful acquire and discard stale targets.
- Isolate cleared account generations so stale dials cannot repopulate a new pool generation.
- Add Debug-level routing diagnostics containing only transport, account ID, normalized final model/tier, hint-generation state, and WS affinity decisions. Tokens and caller-provided header values are never logged.

No external API, database field, or migration is added.

## Validation

- `go test ./internal/service -count=1` (pass, 98.977s)
- `go test -race ./internal/service -count=1 -run 'Test(OpenAIWSConnPool|OpenAIWSConnPool_)'` (pass, 3.175s)
- `make test-unit` from `backend/` (pass, 145.535s)
- `make test-integration` from `backend/` (pass, 98.482s)
- `golangci-lint run --timeout=30m --new ./...` (pass, 0 new issues)
- `git diff --check 32e4de79420f747ddef741e15474ff5e6515000a..HEAD` (pass)
- Fork PR CI passed backend tests, frontend checks, security scans, shell checks, and `golangci-lint` on the same head commit.

Full repository lint still reports nine pre-existing findings outside the changed files; this PR introduces no new lint findings.

## Integration note

The three commits are based on the shared merge base `32e4de79420f747ddef741e15474ff5e6515000a`. At PR creation time, the head is three commits ahead of and 23 commits behind upstream `main`. The branch was intentionally not auto-rebased or merged with upstream, so this PR contains only the routing-hint change set and leaves integration strategy to the maintainers.

## Deployment acceptance

This PR does not perform a production deployment. After deployment, validation must use fresh connections for HTTP priority, HTTP fast, and WebSocket priority requests. The acceptance criterion is the final event field:

```text
response.completed.response.service_tier == "priority"
```

Also regress default, flex, API-key, continuation, capacity queue fallback, and prewarm scenarios.

## Risk and rollback

The behavioral surface is limited to OpenAI OAuth routing metadata, OAuth beta negotiation, WebSocket connection selection, and the associated pool lifecycle. API-key routing-hint generation remains disabled. If deployed wire behavior diverges, rollback is a revert of these three commits. Production E2E remains pending deployment.

### DIFF
--- a/backend/internal/service/openai_compact_service_tier_test.go
+++ b/backend/internal/service/openai_compact_service_tier_test.go
@@ -1,0 +1,100 @@
+package service
+
+import (
+	"bytes"
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+func TestNormalizeOpenAICompactRequestBodyPreservesServiceTier(t *testing.T) {
+	body := []byte(`{
+		"model":"gpt-5.6-sol",
+		"input":[{"type":"message","role":"user","content":"hello"}],
+		"service_tier":"priority",
+		"prompt_cache_key":"compact-cache-key",
+		"store":false,
+		"stream":true
+	}`)
+
+	normalized, changed, err := normalizeOpenAICompactRequestBody(body)
+	require.NoError(t, err)
+	require.True(t, changed)
+	require.Equal(t, "gpt-5.6-sol", gjson.GetBytes(normalized, "model").String())
+	require.Equal(t, "priority", gjson.GetBytes(normalized, "service_tier").String())
+	require.False(t, gjson.GetBytes(normalized, "prompt_cache_key").Exists())
+	require.False(t, gjson.GetBytes(normalized, "store").Exists())
+	require.False(t, gjson.GetBytes(normalized, "stream").Exists())
+}
+
+func TestOpenAIOAuthCompactHTTPBuildersUsePreservedServiceTierInRoutingHint(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	body := []byte(`{
+		"model":"gpt-5.6-sol",
+		"input":[{"type":"message","role":"user","content":"hello"}],
+		"service_tier":"priority",
+		"stream":true
+	}`)
+	normalized, changed, err := normalizeOpenAICompactRequestBody(body)
+	require.NoError(t, err)
+	require.True(t, changed)
+	require.Equal(t, "priority", gjson.GetBytes(normalized, "service_tier").String())
+
+	account := &Account{
+		Platform: PlatformOpenAI,
+		Type:     AccountTypeOAuth,
+		Credentials: map[string]any{
+			"chatgpt_account_id": "test-account",
+		},
+	}
+	svc := &OpenAIGatewayService{}
+
+	tests := []struct {
+		name  string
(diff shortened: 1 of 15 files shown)


bestony:
## Upstream CI note

The upstream `frontend-security` check currently fails in `Check audit exceptions` because the inherited `frontend/audit.json` reports `nanoid` / `GHSA-2v37-7h3g-55p8` without an entry in `.github/audit-exceptions.yml`.

This PR changes no frontend files, dependency manifests, audit output, or exception policy. The changed-file set is backend-only, and the same head passed the fork's backend tests, frontend checks, security scans, shell checks, and `golangci-lint`. I am leaving the unrelated upstream audit remediation out of this routing-hint PR so maintainers can handle it independently.

PR #5432, opened 2026-08-08
feat(admin): add per-account fallback-only OpenAI session stickiness

### LINKED ISSUE #5431: Add per-account fallback-only OpenAI session stickiness
## Problem

OpenAI `session_hash` stickiness is currently global and unconditional for every schedulable account. When a low-priority fallback account receives a session during a primary-pool outage, that session can remain pinned to the fallback after higher-priority accounts recover, unnecessarily consuming fallback quota.

Changing the global sticky weight is too broad because it also changes normal sticky behavior for primary accounts. Weakening `previous_response_id` affinity is also unsafe because Responses encrypted context may only be usable by the original account.

## Proposed behavior

Add a per-account OpenAI session sticky mode:

- `normal` (default): preserve current behavior.
- `fallback_only`: keep the account sticky only while no strictly higher-priority compatible account can immediately acquire a concurrency slot.

The policy should:

- affect only `session_hash` stickiness;
- preserve hard `previous_response_id` account affinity;
- apply to legacy, advanced, and weighted advanced scheduling;
- retain the fallback when higher-priority accounts are model/endpoint/image/transport/compact incompatible, rate limited, unschedulable, full, or lose the acquisition race;
- atomically try every eligible higher-priority account in priority order, then acquire or wait on the original fallback;
- expose the setting in the OpenAI account edit form without changing create or bulk-edit defaults.

## Compatibility

Existing accounts should migrate to `normal`, and update requests that omit the field must preserve the stored value. Unknown values should be rejected at the API boundary.

apple-ouyang:
## Summary

- add a per-account `openai_session_sticky_mode` enum with `normal` and `fallback_only`, defaulting existing accounts to `normal`;
- expose the setting through the admin account API and OpenAI account edit form with omission-preserving updates and strict value validation;
- make `fallback_only` session-hash selection atomically try all compatible higher-priority accounts before acquiring or waiting on the original fallback;
- apply the behavior to legacy, advanced, and weighted advanced scheduling while preserving hard `previous_response_id` affinity.

## Scheduling contract

A `fallback_only` sticky account yields only to a strictly higher-priority account that supports the requested model, endpoint, image capability, transport, and compact mode and can immediately acquire a concurrency slot. Candidate acquisition races and transient candidate acquisition errors continue through the remaining higher-priority candidates. If none can acquire, the original fallback is acquired or returned with its sticky wait plan.

## Validation

- `go test -tags=unit ./...` (105 packages, 8523 tests)
- `pnpm typecheck`
- `pnpm test:run src/components/account/__tests__/EditAccountModal.spec.ts src/components/account/__tests__/EditAccountModal.grokUpstream.spec.ts` (54 tests)
- `pnpm run build`
- local deployment smoke test with a synthetic session identifier confirmed selection returned to an available higher-priority account

Closes #5431

### DIFF
--- a/backend/ent/account.go
+++ b/backend/ent/account.go
@@ -81,6 +81,8 @@ type Account struct {
 	ParentAccountID *int64 `json:"parent_account_id,omitempty"`
 	// 'global' (default) or 'spark' (shadow reads codex_bengalfox).
 	QuotaDimension account.QuotaDimension `json:"quota_dimension,omitempty"`
+	// OpenaiSessionStickyMode holds the value of the "openai_session_sticky_mode" field.
+	OpenaiSessionStickyMode account.OpenaiSessionStickyMode `json:"openai_session_sticky_mode,omitempty"`
 	// Edges holds the relations/edges for other nodes in the graph.
 	// The values are being populated by the AccountQuery when eager-loading is set.
 	Edges        AccountEdges `json:"edges"`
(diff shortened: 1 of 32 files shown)


apple-ouyang:
Closing this stale PR and its linked proposal. It received no maintainer review, is 510 commits behind main, and now has merge conflicts. The branch and local worktree remain available if upstream requests the feature.

PR #5439, opened 2026-08-09
feat(billing): support safe billing by upstream response model

pigzwy:
## Problem

Some upstream providers automatically downgrade a requested model at runtime. For example, a request for `claude-fable-5` may be answered by `claude-opus-4-8`. The project already observes and records the upstream response model, but billing still uses the requested, channel-mapped, or sent upstream model.

The existing `upstream` billing source means the model sent to the provider; it is not the model declared by the completed upstream response.

## Change

Add an opt-in channel billing source, `response_model`, that bills by the observed upstream response model after a successful request. Existing channels remain unchanged by default.

The response model is accepted only when it is non-empty, unambiguous across a stream, priced, and no more expensive than the baseline billing result. Otherwise billing safely falls back to the existing model. This prevents an upstream declaration from increasing the user charge.

The request model, sent model, response model, and mismatch audit fields remain preserved. No database migration is introduced.

## Validation

Added coverage for downgrade billing, rejecting a more expensive response declaration, conflict/empty/unpriced fallbacks, default behavior, and channel restriction prechecks.

CI should run the repository backend unit/integration tests and frontend typecheck/build.

Related upstream work: #5396 (upstream response-model audit).

### DIFF
--- a/backend/internal/handler/admin/channel_handler.go
+++ b/backend/internal/handler/admin/channel_handler.go
@@ -33,7 +33,7 @@ type createChannelRequest struct {
 	GroupIDs                   []int64                          `json:"group_ids"`
 	ModelPricing               []channelModelPricingRequest     `json:"model_pricing"`
 	ModelMapping               map[string]map[string]string     `json:"model_mapping"`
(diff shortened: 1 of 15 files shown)


pigzwy:
recheck

PR #5441, opened 2026-08-09
feat(billing): support New API rates and recharge multipliers

shirosoralumie648:
## Summary

Adds New API upstream billing-rate probing and a configurable recharge multiplier for API-key upstream accounts, with explicit safeguards around New API catalogue sources.

## Changes
- Probe `GET /v1/sub2api/billing` first and fall back to New API catalogue routes only on `404` or `405`.
- Request `GET /api/pricing` anonymously. The relay API key and account header overrides are never sent to dashboard/catalogue routes.
- If anonymous `/api/pricing` returns `401`, `403`, or an HTTP-200 authentication error envelope, try anonymous `GET /api/user/groups` as a smaller catalogue fallback.
- Treat the instance as `newapi_pricing_auth_required` only when protected pricing has no usable anonymous `/api/user/groups` fallback, using unsupported-status backoff.
- Require an explicit `upstream_billing_newapi_group` for every New API fallback, including when the intended group is literally `default`. Neither catalogue endpoint is treated as API-key group discovery.
- Parse only the explicitly selected `group_ratio` or nested `data[group].ratio` entry, so malformed unrelated groups do not invalidate a valid selection.
- Permit New API automatic writeback only when the observed `upstream_group` matches the account's current explicit group; repository CAS still rejects concurrent configuration changes.
- Add `upstream_billing_recharge_multiplier` with default `1` and compute corrected rates as upstream rate multiplied by the recharge multiplier.
- Preserve raw and corrected values in the probe snapshot; corrected values drive sorting, scheduling, display, and automatic account-rate synchronization. Peak-hour coefficients remain excluded from static writeback.
- Keep settings consistent across create, edit, bulk edit, duplicate, CRS synchronization, repository updates, and snapshot invalidation.
- Update bilingual UI guidance and show actionable New API group/authentication probe errors.

## Verification
- `go test ./internal/repository ./internal/service -count=1`
- `go test -tags unit ./internal/server ./internal/handler/admin -count=1`
- `go test ./... -run '^$'`
- Frontend Vitest: 5 files, 102 tests passed.
- `pnpm run typecheck`
- `pnpm run build`
- Targeted ESLint and `git diff --check`

## Notes
The New API response and authentication semantics were checked against current upstream behavior and reviewer live-instance findings. This revision does not add or reuse dashboard credentials. A protected `/api/pricing` route is supported only when anonymous `/api/user/groups` supplies the explicitly configured group; the key-scoped `/v1/sub2api/billing` path is unchanged.


shirosoralumie648:

Brisbanehuang:
Thanks for working on this. I tested the fallback path against a live New API deployment using an ordinary relay API key. All instance, account, group, and rate details are intentionally redacted.

I found a blocker in the current authentication flow:

1. `GET /v1/sub2api/billing` returned `404`, so the compatibility fallback was entered as expected.
2. `GET /api/pricing` with `Authorization: Bearer <relay-api-key>` returned HTTP `200`, but the JSON payload had `success: false` with `Unauthorized, invalid access token`.
3. The same pricing request without `Authorization` returned `401` because that deployment requires dashboard authentication for the pricing module.

This matches current New API behavior: `/api/pricing` is behind `HeaderNavModuleAuth("pricing")`; `TryUserAuth` interprets any `Authorization` credential as a dashboard session token/PAT. A normal relay API key is not a valid dashboard access token. However, `requestUpstreamBillingProbe` currently adds the account relay key to both the Sub2API endpoint and the New API dashboard endpoint.

As a result, the fallback is not usable for a protected pricing page. It may also break an otherwise-public pricing page, because supplying an invalid relay key causes optional dashboard authentication to fail instead of continuing anonymously. In the observed case, `parseNewAPIUpstreamBillingResponse` would reject the HTTP-200 error body and persist `invalid_response` with status `200`.

There are two related correctness concerns:

- An empty configured group is converted to `default`. New API deployments are not required to expose a group literally named `default`; the tested deployment did not. Such accounts will still fail with `newapi_group_required` even after the authentication issue is resolved.
- `/api/pricing` describes groups available to the authenticated dashboard user and may apply user-group-to-target-group overrides. It does not prove which group the relay API key actually uses. Automatic rate synchronization can therefore write the wrong multiplier if the configured group is incorrect.

Suggested direction:

- Do not reuse the relay API key when requesting dashboard endpoints.
- Either introduce a separate optional New API dashboard credential, or use a relay-key-compatible/dedicated billing endpoint.
- A smaller group-catalog endpoint can be used only as a catalog source with an explicitly configured group; it still should not be treated as automatic API-key group discovery.
- Require explicit confirmation of the group before enabling automatic rate writeback for this fallback source.
- Consider parsing only the selected `group_ratio` entry. The current loop rejects the entire response if any unrelated group entry is malformed.

The existing `/v1/sub2api/billing` path is unaffected; the issue is specifically the New API `/api/pricing` compatibility fallback and its credential semantics.

shirosoralumie648:
@Brisbanehuang Thanks for the detailed live-instance report. I addressed the blocker and related correctness points in `b72de61eb`.

- `/api/pricing` is now requested anonymously. The relay API key and account header overrides are sent only to `/v1/sub2api/billing`, never to the New API dashboard route.
- The fallback now requires an explicitly configured group, including a literal `default`. A blank group stops before the pricing request; `/api/pricing` is not used to discover the relay API key group.
- The parser reads only the selected `group_ratio` entry and ignores malformed unrelated entries.
- Automatic writeback for New API data is accepted only when `data.upstream_group` matches the account's current explicit group. The repository CAS also rejects a concurrent group/configuration change.
- `401`, `403`, and HTTP-200 auth-error envelopes are recorded as `newapi_pricing_auth_required` with unsupported-status backoff. Deployments that protect the pricing module remain intentionally unsupported in this revision; no dashboard PAT/session is reused or added.
- The UI now explains the explicit-group requirement and protected-pricing limitation.

Verification passed: backend repository/service tests, unit-tagged server/admin-handler tests, all-package compilation, 4 frontend test files (100 tests), typecheck, production build, targeted ESLint, and `git diff --check`.

Brisbanehuang:
Thanks — I reviewed `b72de61eb`, and the changes address the issues I raised.

One compatibility gap remains in the same redacted live test: anonymous `/api/pricing` returns `401`, while anonymous `/api/user/groups` returns `200` and provides the explicitly selected group's ratio. Would you consider `/api/user/groups` as an additional catalogue fallback, still requiring an explicit group and never treating it as API-key group discovery?

Otherwise, reporting protected `/api/pricing` instances as unsupported looks correct.

shirosoralumie648:
@Brisbanehuang Thanks for pointing out the smaller public catalogue. I added the `/api/user/groups` fallback in `857a38c31`.

- When anonymous `/api/pricing` returns `401`, `403`, or an HTTP-200 authentication error envelope, the probe now requests `/api/user/groups` anonymously.
- Neither catalogue request carries the relay API key or account header overrides.
- The parser reads only `data[configured_group].ratio`; an explicit group is still required, unrelated malformed entries are ignored, and the endpoint is never used to discover or guess the relay API key's group.
- If `/api/user/groups` is absent or also requires authentication, the probe keeps `newapi_pricing_auth_required` and the unsupported-status backoff.
- Recharge multiplication and the explicit-group match required for automatic writeback are unchanged.

Verification passed: repository/service tests, unit-tagged server/admin-handler tests, all-package Go compilation, 5 frontend test files (102 tests), typecheck, production build, targeted ESLint, and `git diff --check`.

Brisbanehuang:
Thanks — I reviewed `857a38c31`, and the compatibility behavior looks good.

One UX improvement may be worthwhile: instead of requiring administrators to type the explicit New API group name manually, could the account form first let them choose the billing probe type (`Sub2API` or `New API`)? When `New API` is selected, it could anonymously fetch `/api/user/groups` from the configured base URL and present the returned group names in a dropdown.

The selected exact group name would still be stored explicitly, and the current group-match/writeback safeguards could remain unchanged. This would preserve the existing correctness model while avoiding typos and confusing assumptions around names such as `default`.

shirosoralumie648:
@Brisbanehuang Thanks for the suggestion. A group catalogue dropdown would improve the setup experience while preserving the explicit-group safeguards.

I would prefer to keep this PR focused on rate discovery and recharge synchronization. Supporting group discovery safely before account creation would require a dedicated admin endpoint, URL/SSRF validation, proxy and TLS handling, plus create/edit form states and tests. This seems better suited to a follow-up PR.

For this PR, the exact New API group will remain explicitly configured, and no API-key group inference will be performed.

PR #5443, opened 2026-08-09
fix(gemini): honor RetryInfo on 429 and stop cooling Vertex service accounts until PST midnight

llissery:
## Problem

Vertex AI (`platform=gemini`, `type=service_account`) reports transient 429s with a standard `google.rpc.RetryInfo` detail:

```json
{"error":{"code":429,"status":"RESOURCE_EXHAUSTED","details":[
  {"@type":"type.googleapis.com/google.rpc.RetryInfo","retryDelay":"39s"}]}}
```

`ParseGeminiRateLimitResetTime` only understands the AI Studio `metadata.quotaResetDelay` shape (plus the `Please retry in Xs` regex), so parsing always fails for Vertex bodies. The fallback in `handleGeminiUpstreamError` then has no `service_account` branch — `IsGeminiCodeAssist()` / `GeminiOAuthType()` only apply to `type=oauth` — so the account falls into the API-key / AI-Studio branch and is rate limited until the next PST midnight.

Net effect: a ~39s per-minute throttle on a pay-as-you-go account turns into an up-to-24h local lockout. We hit this three times within 48 hours in production before patching our deployment.

## Change

- `ParseGeminiRateLimitResetTime` now also honors `google.rpc.RetryInfo.retryDelay` from `error.details` (`"39s"`, `"1.5s"`, ...), rounded up to whole seconds like the existing `quotaResetDelay` path, and capped at 15 minutes to guard against absurd upstream values.
- When a 429 body carries no parseable reset time at all, `service_account` accounts now cool down for 60s instead of until PST midnight. Vertex pay-as-you-go quotas are short-window RPM/TPM; there is no daily free-tier reset to wait for.

## Scope / behavior notes

- Code Assist, Google One, and API-key fallback branches are unchanged; a new regression test pins the API-key PST-midnight behavior.
- The daily-quota check (`per day` wording → PST midnight) still runs first, so AI Studio free-tier daily limits are unaffected. Bodies that carry RetryInfo without daily-quota wording now honor the server-provided delay, which was already this function's intent for `quotaResetDelay` / `Please retry in Xs`.
- The 15-minute cap applies only to the new RetryInfo path; existing `quotaResetDelay` behavior is untouched.
- Since `ParseGeminiRateLimitResetTime` is shared, the antigravity retry path and the rate-limit service pick up RetryInfo parsing as well, with the same cap.

## Relation to existing PRs

#5011 and #5344 fix the same PST-midnight fallback (via `rate_limit_429_cooldown_settings` and `geminiCooldownForTier` respectively), but neither parses `RetryInfo`, so accounts would still ignore the delay the server actually asks for. With RetryInfo honored, the fallback rarely fires for Vertex — real-world Vertex 429s virtually always carry it — and the fixed 60s covers the rare body without it. Happy to switch the fallback to either existing mechanism if that is preferred. #2408 (MODEL_CAPACITY_EXHAUSTED) and #1254 (Google One) target different account paths.

## Tests

- New table cases in `TestParseGeminiRateLimitResetTime`: RetryInfo parse, fractional round-up, 15-minute cap, `@type` mismatch ignored, invalid duration ignored, `quotaResetDelay` precedence kept.
- New `handleGeminiUpstreamError` tests: RetryInfo path (39s), resetless Vertex fallback (asserts minutes-scale, not PST midnight), API-key PST-midnight regression guard.
- `make test-unit` and `make test-integration` pass locally.

### DIFF
--- a/backend/internal/service/gemini_error_policy_test.go
+++ b/backend/internal/service/gemini_error_policy_test.go
@@ -588,6 +588,77 @@ func TestHandleGeminiUpstreamError_PoolMode429(t *testing.T) {
 	}
 }
 
+// Vertex（service_account）429 响应带 google.rpc.RetryInfo 时，应按 retryDelay 冷却，
+// 而不是回退到 PST 午夜。
+func TestHandleGeminiUpstreamError_VertexRetryInfoUsesParsedDelay(t *testing.T) {
+	repo := &rateLimit429AccountRepoStub{}
+	svc := &GeminiMessagesCompatService{accountRepo: repo}
+
+	account := &Account{
+		ID:       601,
+		Platform: PlatformGemini,
+		Type:     AccountTypeServiceAccount,
+		Credentials: map[string]any{
+			"project_id": "my-vertex-project",
+			"location":   "global",
+		},
+	}
+	body := []byte(`{"error":{"code":429,"message":"Resource exhausted. Please try again later.","status":"RESOURCE_EXHAUSTED","details":[{"@type":"type.googleapis.com/google.rpc.RetryInfo","retryDelay":"39s"}]}}`)
+
+	before := time.Now()
+	svc.handleGeminiUpstreamError(context.Background(), account, http.StatusTooManyRequests, http.Header{}, body)
+
+	require.Equal(t, 1, repo.rateLimitCalls)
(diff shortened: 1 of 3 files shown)


llissery:

PR #5462, opened 2026-08-09
[codex] restore original CodexRip prompt and unify Skill source

HTExplicit:
## What changed

- Restored the full CodexRip system prompt from the fixed offline source, preserving the original body and limiting transformations to branding, the exact user address, and the two routed Skill blocks.
- Pinned the native reverse Skill seed to upstream commit a5d8c9233b98c52df387d5b1a0ef669fcaa51374, with one canonical bundle route and local bundle-relative loading.
- Replaced GitHub client installation metadata with content-addressed HTTPS PowerShell/Python bootstraps from codexrip.vip, including hash verification, atomic install, no redirects, and rollback-safe failure behavior.
- Hardened source/manifest/path checks and kept legacy moxinggang overlay compatibility scoped to the supported offline artifact.

## Validation

- python tools/verify_remote_skill_registry.py
- python tools/test_remote_skill_seed_contract.py
- python tools/test_remote_skill_bootstrap_contract.py
- go test ./internal/service -count=1
- go test ./internal/handler/admin ./internal/middleware -count=1
- go test ./migrations -count=1
- go vet ./...
- frontend Vitest SystemPromptsView, typecheck, lint:check, and build

This is a draft for CI and upstream review; no production deployment or prompt enablement was performed in this worktree.


HTExplicit:
Opened against the upstream repository by mistake. This is a downstream-only patch and is being moved to HTExplicit/sub2api.

PR #5471, opened 2026-08-09
fix(claude): stop Claude OAuth tokens from silently never refreshing

xiongnemo:
## Problem

Claude OAuth accounts can stop refreshing entirely. The stored `access_token` is then served past its expiry until the upstream rejects it:

```
401 {"type":"error","error":{"type":"authentication_error",
     "message":"OAuth access token has expired. Re-authenticate to continue."},"request_id":null}
```

The failure is silent: `ClaudeTokenProvider.GetAccessToken` logs `claude_token_refresh_failed` at Warn and then returns the stale token anyway (`claude_token_provider.go:121`), so the only visible symptom is an upstream 401 that looks like a credential problem rather than a refresh problem.

## Root cause

`ClaudeTokenRefresher.NeedsRefresh` (`token_refresher.go:52`) is the single gate used by **both** refresh paths:

- the background `TokenRefreshService` — `token_refresh_service.go:619`
- the on-demand second check inside `OAuthRefreshAPI.RefreshIfNeeded` — `oauth_refresh_api.go:250`

If it returns `false`, nothing refreshes, ever. Two ways that happens:

**1. `expires_at` recorded in Unix milliseconds.**

`GetCredentialAsTime` (`account.go:342`) parsed every numeric timestamp as seconds:

```go
if ts, err := strconv.ParseInt(s, 10, 64); err == nil {
    t := time.Unix(ts, 0)
    return &t
}
```

A millisecond value therefore lands far in the future and `time.Until()` saturates:

```
expires_at=1786102299052 -> time.Unix() = 58569-04-30 | time.Until = 2562047h47m | NeedsRefresh(3min) = false
```

This is reachable in normal use: Claude Code stores `expiresAt` in **milliseconds** in `~/.claude/.credentials.json`, and account credentials are persisted as supplied — `CreateAccountRequest.Credentials` is a raw `map[string]any` and only `NormalizeHeaderOverrideCredentials` runs over it (`admin_account.go:312`), so an imported credential document carries the millisecond value straight into the DB.

**2. `expires_at` missing or unparseable returned `false`, meaning "never refresh".**

This also contradicts the provider, which treats a nil expiry as *needing* a refresh:

```go
// claude_token_provider.go:82
needsRefresh := expiresAt == nil || time.Until(*expiresAt) <= claudeTokenRefreshSkew
```

So the provider requests a refresh, `RefreshIfNeeded` silently declines it at step 3, and the stale token is served and cached anyway.

`GrokTokenRefresher` (`grok_token_refresher.go:49`) and `OpenAITokenRefresher` (`token_refresher.go:103`) already treat an unknown expiry as "refresh" and skip accounts with no refresh token. Claude was the odd one out.

## Changes

- **`GetCredentialAsTime` detects millisecond timestamps.** Values above `1e11` (which as seconds would be year 5138) are parsed with `time.UnixMilli`. Shared by every platform's expiry handling, so Gemini/OpenAI/Grok/Antigravity benefit from the same normalization.
- **`ClaudeTokenRefresher.NeedsRefresh` treats an unknown expiry as "refresh now"** and skips accounts without a `refresh_token`, matching the existing Grok/OpenAI refreshers.
- **Parse `refresh_token_expires_in` and persist it as `refresh_token_expires_at`.** The Anthropic token endpoint returns this field; sub2api discarded it, so a refresh token that expires in ~28 days fails with no warning and no way to surface the deadline in the UI. It is omitted when upstream does not return it, so `MergeCredentials` preserves any previously known value.

## Ground truth used

Refresh behaviour was cross-checked against the official Claude Code client (v2.1.222). Its own OAuth refresh records:

```js
expiresAt: Date.now() + e.expires_in * 1000,           // milliseconds
refreshTokenExpiresAt: vFo(e.refresh_token_expires_in, true),
```

and refreshes when `Date.now() + 300000 >= expiresAt` (a 5-minute window). It warns the user once `refreshTokenExpiresAt` is within 3 days. That confirms both that `refresh_token_expires_in` is a real response field and that the millisecond `expiresAt` in `.credentials.json` is what users are copying in.

## Testing

```
go build ./...                 # OK
go test -tags=unit ./...       # OK
```

Two failures are **pre-existing** and unrelated — verified to reproduce identically on a clean checkout of `main` (`git stash` + rerun):

- `TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig` (`internal/service`)
- `TestAliyunCaptchaVerifier_TransportError` (`internal/repository`)

New/updated coverage:

- `TestGetCredentialAsTime_UnitDetection` — seconds vs milliseconds vs RFC3339, including the `9999999999` boundary that must stay seconds.
- `TestGetCredentialAsTime_MillisecondsDoNotLandInTheFarFuture` — regression test for the exact defect.
- `TestClaudeTokenRefresher_NeedsRefresh` — extended with millisecond cases, unknown-expiry cases, and a no-`refresh_token` case. Existing cases were given a `refresh_token` so they exercise the window logic rather than the new guard.
- `TestBuildClaudeAccountCredentials_RefreshTokenExpiresAt{,OmittedWhenUnknown}` — including that a previously known value survives `MergeCredentials`.

## Deliberately not included

Two further differences from the official client were found but left out, because they change the outbound request to Anthropic and I could not verify them against a live token endpoint without risking a regression for setups that work today:

1. Claude Code sends `scope` on refresh (the value matches this repo's `oauth.ScopeAPI` exactly); `claude_oauth_service.go:236` omits it. Note the Anthropic SDK's own `userOAuthProvider` also omits it, so this is likely tolerated.
2. Claude Code sends `anthropic-beta: oauth-2025-04-20` on `/v1/oauth/token`; `claude_oauth_service.go:246` does not.

Happy to add either if you would like them in scope.

---

<details>
<summary>中文摘要</summary>

Claude OAuth 账号可能完全停止刷新，过期的 `access_token` 会被继续发往上游，直到返回 401。

`ClaudeTokenRefresher.NeedsRefresh` 是后台定时刷新和请求路径二次确认**共用的唯一闸门**，两种情况下它会永远返回 `false`：

1. `expires_at` 存成了毫秒。`GetCredentialAsTime` 一律按秒解析，`1786102299052` 会变成公元 58569 年，`time.Until()` 饱和到最大值。Claude Code 的 `~/.claude/.credentials.json` 正是用毫秒记录 `expiresAt`，而账号凭据是原样入库的。
2. `expires_at` 缺失或无法解析时返回 `false`（永不刷新），这与 `ClaudeTokenProvider` 第 82 行的判断相互矛盾 —— provider 认为该刷新，`RefreshIfNeeded` 却静默否决。

修复：`GetCredentialAsTime` 自动识别毫秒时间戳；`NeedsRefresh` 在过期时间未知时刷新一次，并对齐 Grok/OpenAI 的 `refresh_token` 守卫；解析并保存 `refresh_token_expires_in`，以便在 refresh token（约 28 天）到期前有迹可循。

</details>

---

### DIFF
--- a/backend/internal/pkg/oauth/oauth.go
+++ b/backend/internal/pkg/oauth/oauth.go
@@ -184,13 +184,17 @@ func BuildAuthorizationURL(state, codeChallenge, scope string) string {
 
 // TokenResponse represents the token response from OAuth provider
 type TokenResponse struct {
-	AccessToken  string       `json:"access_token"`
-	TokenType    string       `json:"token_type"`
-	ExpiresIn    int64        `json:"expires_in"`
-	RefreshToken string       `json:"refresh_token,omitempty"`
-	Scope        string       `json:"scope,omitempty"`
-	Organization *OrgInfo     `json:"organization,omitempty"`
-	Account      *AccountInfo `json:"account,omitempty"`
+	AccessToken  string `json:"access_token"`
+	TokenType    string `json:"token_type"`
+	ExpiresIn    int64  `json:"expires_in"`
+	RefreshToken string `json:"refresh_token,omitempty"`
+	// RefreshTokenExpiresIn 是 refresh_token 的剩余有效期（秒）。
+	// Anthropic 的 token 端点会返回该字段，官方 Claude Code 据此记录
+	// refreshTokenExpiresAt 并在临近到期时提示重新授权。
+	RefreshTokenExpiresIn int64        `json:"refresh_token_expires_in,omitempty"`
+	Scope                 string       `json:"scope,omitempty"`
+	Organization          *OrgInfo     `json:"organization,omitempty"`
+	Account               *AccountInfo `json:"account,omitempty"`
 }
 
 // OrgInfo represents organization info from OAuth response
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -334,11 +334,18 @@ func (a *Account) GetCredential(key string) string {
 	}
 }
 
+// maxPlausibleUnixSeconds 用于区分秒级与毫秒级 Unix 时间戳。
+// 1e11 秒 ≈ 公元 5138 年，任何真实的秒级时间戳都远小于它；
+// 而当前的毫秒级时间戳约为 1.7e12，必然大于它。
+const maxPlausibleUnixSeconds = int64(1e11)
+
 // GetCredentialAsTime 解析凭证中的时间戳字段，支持多种格式
 // 兼容以下格式：
 //   - RFC3339 字符串: "2025-01-01T00:00:00Z"
 //   - Unix 时间戳字符串: "1735689600"
 //   - Unix 时间戳数字: 1735689600 (float64/int64/json.Number)
+//   - Unix 毫秒时间戳: 1735689600000（如 Claude Code 的 ~/.claude/.credentials.json
+//     里的 expiresAt，导入凭据时可能被原样写入）
 func (a *Account) GetCredentialAsTime(key string) *time.Time {
 	s := a.GetCredential(key)
 	if s == "" {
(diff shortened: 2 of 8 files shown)


xiongnemo:

PR #5490, opened 2026-08-10
fix(openai): preserve Anthropic prompt cache intent

kmiaojun:
## Summary

Anthropic Messages clients can mark reusable prompt prefixes with `cache_control`, but the Anthropic-to-Responses conversion currently drops that intent. This can turn stable agent prefixes into repeated cache misses when the downstream conversation identity changes.

This change adds an account-scoped prompt-cache capability mode for OpenAI API-key upstreams. Existing accounts remain unchanged unless an administrator explicitly opts in.

## What changed

- Add `accounts.extra.openai_prompt_cache_mode` with four strict modes:
  - `off`: retain the existing behavior; missing or invalid values also resolve to `off`
  - `stable_key_only`: derive a stable, tenant-isolated `prompt_cache_key`, then remove all translated breakpoint fields
  - `implicit_breakpoints`: send the stable key and exactly representable breakpoints without `prompt_cache_options`
  - `explicit_only`: send the stable key, representable breakpoints, and `prompt_cache_options.mode=explicit`
- Convert Anthropic `cache_control: { "type": "ephemeral" }` only where Responses has an exact representation:
  - structured system text to `input_text`
  - user text to `input_text`
  - user image to `input_image`
- Derive the stable cache identity from the downstream API-key ID, mapped upstream model, request-shaping fields, and the converted reusable prefix through the first supported breakpoint.
- Keep conversation identity independent from cache identity so unrelated conversations cannot share `previous_response_id` state.
- Omit cache key/options/breakpoints on continuation requests using `previous_response_id`; restore the configured mode if the existing full-replay path is triggered after a missing previous response.
- Preserve the account Extra field in scheduler snapshots.

## Compatibility boundaries

- The feature is opt-in and applies only to OpenAI-platform API-key accounts.
- OAuth, Grok, missing/invalid configuration, and existing accounts remain on the legacy path.
- Tool definitions, assistant blocks, `tool_use`, and `tool_result` cache markers are not relocated because Responses cannot represent those boundaries exactly.
- Anthropic TTL values are not forwarded because Anthropic and Responses TTL semantics are not equivalent.
- There is no hostname, gateway, account, or model-name allowlist.
- There is no response-driven mode change or unsupported-field retry.
- `stable_key_only` provides a safe outbound contract; actual cache hits still depend on upstream support and stable prompt contents.

## Tests

- `git diff --check origin/main...HEAD`
- targeted unit tests for conversion, mode resolution, stable-key isolation, legacy preservation, continuation/replay, OAuth/Grok boundaries, and scheduler metadata
- `make -C backend test-unit`

All tests pass on current `main`.


kmiaojun:

kmiaojun:
Updated this PR against `main` at `98d86915` (v0.2.4). The Anthropic-to-Responses conversion on that base still drops the representable `cache_control` intent addressed here.

- Resolved all four merge conflicts while retaining the current account-scoped upstream session isolation, SetupToken protocol handling, and GPT-6 model coverage.
- Added a negative mode-resolution test for OpenAI SetupToken accounts. The new cache modes remain opt-in and limited to OpenAI API-key accounts.
- Removed the unused conversion wrapper reported by `golangci-lint` in the first fork run. This is fixed in `47871f1d`.

Validation for the current head `47871f1d01ea292d88a004dd0b31b46ec01d2341`:

- [Fork CI run 34332255050](https://github.com/kmiaojun/sub2api/actions/runs/34332255050): all four jobs passed — shell checks, backend unit and integration tests, frontend checks, and golangci-lint.
- `git diff --check` against the synced main base passed.

The upstream PR workflows are still awaiting maintainer approval. Could a maintainer approve the workflows and review the updated changes? The CLA was already signed.

kmiaojun:
@Wei-Shaw 您好，想请您帮忙看一下这个 PR。8 月 10 日提交后目前还没有人工评审；现在已经同步 main、解决全部冲突，fork 的 4 项 CI 均已通过，测试链接与变更说明见上一条评论。

目前官方工作流仍等待维护者批准。方便时能否帮忙批准 CI 并安排评审？如果实现方向需要调整，或更适合拆分成较小的 PR，也请指出，我们会配合修改。谢谢！

PR #5491, opened 2026-08-10
fix(openai): retry rejected reasoning content

East5RingRoad-kyle:
## Summary

- retry an OpenAI Responses request when the upstream explicitly rejects a non-empty `reasoning.content` array
- trust only `array_above_max_length` with structured `input[N].content`
- preserve all non-reasoning items and reuse the existing retry-loop guards

## Production symptom

OpenAI returned HTTP 400 for an indexed reasoning content array; sub2api surfaced the failure as a generic 502. The same request shape failed across multiple models while unrelated Responses requests remained healthy.

## Tests

- `go test ./internal/service -run ReasoningContent -count=1`


East5RingRoad-kyle:

PR #5497, opened 2026-08-10
fix(openai): retry capacity errors after metadata preamble

cyhhao:
## Summary

- keep metadata-only Responses SSE events attempt-local until semantic output starts
- preserve pre-output capacity failover when upstream emits empty reasoning metadata before `server_is_overloaded`
- exclude native Responses keepalive comments from the handler's committed-output boundary
- retain protocol-correct terminal SSE errors when failover is exhausted after keepalives

## Problem

The capacity recovery added in #5398 works while no client output has started. Some Codex Responses streams, however, emit metadata-only events before the capacity error:

```text
response.created
response.output_item.added          # empty reasoning item
response.reasoning_summary_part.added # empty summary part
error: server_is_overloaded
response.failed
```

Those empty metadata events were flushed immediately and marked the attempt as client-visible output. The later overload event could therefore no longer use the existing same-account retry and account failover path, even though no semantic content had reached the client.

## Fix

Metadata-only events are staged per attempt until an event crosses the existing semantic-output boundary. A pre-output overload discards the staged preamble and returns the existing failover error. On successful output, the staged events are flushed in order before the first semantic event.

Keepalive comments remain client-visible for connection liveness but are excluded from the handler's adjusted written-size comparison. This keeps account failover replay-safe after a keepalive while ensuring final exhaustion is still encoded as a valid Responses SSE terminal event.

Passthrough rules, content-policy errors, usage accounting, response-model rewriting, and post-output error behavior remain unchanged. This PR does not change capacity retry timing; the 500 ms exponential backoff remains isolated in #5403.

## Tests

- metadata-only preamble followed by capacity overload
- large metadata preamble and keepalive-before-overload
- native and passthrough Responses paths
- handler-level account switching after a committed keepalive
- first-output timeout and adjusted keepalive-size regressions

### DIFF
--- a/backend/internal/handler/openai_gateway_handler.go
+++ b/backend/internal/handler/openai_gateway_handler.go
@@ -540,7 +540,7 @@ func (h *OpenAIGatewayHandler) Responses(c *gin.Context) {
 		// Forward request
 		service.SetOpsLatencyMs(c, service.OpsRoutingLatencyMsKey, time.Since(routingStart).Milliseconds())
 		forwardStart := time.Now()
-		// 用扣除 compact 心跳字节的口径快照：心跳注释不构成语义响应，
+		// 用扣除非语义心跳字节的口径快照：心跳注释不构成语义响应，
 		// 不能因心跳字节变化而放弃 failover 换号（#3887）。
 		writerSizeBeforeForward := service.OpenAICompactKeepaliveAdjustedWrittenSize(c)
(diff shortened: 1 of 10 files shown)


cyhhao:
@codex review

cyhhao:
@codex review

cyhhao:
@codex review

e06084:
这个修复什么时候发布～

PR #5522, opened 2026-08-11
fix: retry Codex capacity errors

zijiren233:
Fix Codex SSE requests that fail with:

`Selected model is at capacity. Please try a different model.`

Buffer initial lifecycle events so failover can run before the response is committed.


zijiren233:

PR #5532, opened 2026-08-11
Expose official pricing tiers and reference metadata in model plaza

allenxln:
## Summary

- parse explicit long-context price fields such as `*_above_256k_tokens` and `*_above_272k_tokens` into normalized pricing tiers
- carry catalog-provided `pricing_reference_model` and `pricing_reference_note` metadata through the pricing and model-plaza service layers
- expose optional `tiers`, `reference_model`, and `reference_note` fields in the public model-plaza response
- render long-context official prices and explicit reference-model notes in the model-plaza table
- add English and Chinese labels plus backend/frontend regression coverage

## Why

The model plaza currently exposes only one official price tier. The pricing catalog already contains higher long-context rates for models such as GPT and Qwen, but those fields are ignored. Some public aliases also use a canonical model as a price reference; without explicit metadata, the UI can only present that value as if it were an independent official price.

This change keeps the pricing catalog as the only source of public reference metadata. It does not infer or expose channel `model_mapping` values, and it does not change the billing path, channel pricing, or group multipliers.

## API compatibility

All new response fields are optional. Existing catalogs and clients continue to receive and render the current base official prices.

## Validation

- `go test -tags unit ./internal/service ./internal/handler`
- `vitest run src/components/modelPlaza/__tests__/PlazaModelPricingTable.spec.ts`
- `vue-tsc --noEmit`
- ESLint on the modified frontend files

### DIFF
--- a/backend/internal/handler/model_plaza_handler.go
+++ b/backend/internal/handler/model_plaza_handler.go
@@ -35,13 +35,24 @@ func NewModelPlazaHandler(
 	}
 }
 
+type modelPlazaOfficialPricingTier struct {
+	MinInputTokens  int      `json:"min_input_tokens"`
+	InputPrice      *float64 `json:"input_price"`
+	OutputPrice     *float64 `json:"output_price"`
+	CacheWritePrice *float64 `json:"cache_write_price"`
+	CacheReadPrice  *float64 `json:"cache_read_price"`
+}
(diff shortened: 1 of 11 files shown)


allenxln:

allenxln:
recheck

PR #5534, opened 2026-08-11
feat: support upstream account type in CreateAccountModal

Loliyer520:
## Summary

Add an **upstream** account category option to the account creation modal, allowing admins to create upstream (reverse-proxy style) accounts directly from the UI instead of via API/DB only.

## Changes

- New "upstream" card in `CreateAccountModal.vue` account category selector
- Quota injection support for upstream account type
- Probe support for upstream accounts
- Grok branch skips upstream to avoid overwriting already-set `base_url`/`model_mapping`

## Why

Upstream accounts (accounts that proxy to another upstream gateway) previously required manual DB/API setup. This makes them first-class citizens in the admin UI.

## Test plan

- vue-tsc passes
- vite build succeeds
- Docker image built and deployed; UI shows the upstream option and account creation works end-to-end

### DIFF
--- a/frontend/src/components/account/CreateAccountModal.vue
+++ b/frontend/src/components/account/CreateAccountModal.vue
@@ -349,6 +349,33 @@
             </div>
           </button>
 
+          <button
+            type="button"
+            data-testid="openai-account-type-upstream"
+            @click="accountCategory = 'upstream'"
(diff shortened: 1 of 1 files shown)


Loliyer520:

PR #5562, opened 2026-08-12
fix(apicompat): normalize Chat function tool_choice for Responses

liulixin-lex:
## Summary
- normalize the nested Chat Completions function `tool_choice` shape before forwarding it to a Responses upstream
- preserve an already Responses-native choice, prefer its top-level `name` when both forms are present, and leave non-function choices unchanged
- cover the conversion with a real Chat `tools` declaration

## Problem
Chat Completions represents a forced function choice as `{"type":"function","function":{"name":"..."}}`, while the Responses API requires the function name at the top level. Passing the Chat shape through causes Responses-compatible upstreams to reject the request because `tool_choice.name` is missing.

This only changes the Chat Completions-to-Responses conversion path. Native `/v1/responses` requests remain unchanged.

## Testing
- `go test ./internal/pkg/apicompat -run "^TestChatCompletionsToResponses" -count=1`
- `go test ./internal/pkg/apicompat -count=1`


liulixin-lex:

PR #5565, opened 2026-08-12
feat(grok): add grok-4.6 to the text model catalog

cuixiaorui:
## Why

Grok 4.6 is already live on the xAI API and is the default model in Grok Build / Grok CLI. Sub2API's Grok account mapping is a whitelist (`xai.DefaultModelMapping`). Requests for `grok-4.6` are treated as unsupported and clients report *selected model may not exist or you may not have access*.

## What

Register `grok-4.6` as a first-class Grok **text** model, using the same pattern as `grok-4.3`:

- catalog + aliases in `xai/models.go` (`grok-4.6`, `grok-4.6-latest`, plus `xai/` `x-ai/` `grok/` prefixes via existing helper)
- official fallback price ($2 / $0.50 cached / $6 per MTok, same as 4.5)
- reasoning-effort support
- frontend whitelist / mapping preset / Use Key configs
- settings datalist option

**Not changed on purpose**

- `DefaultTextModel` stays `grok-4.5`
- Claude Code generated env (`ANTHROPIC_MODEL`) stays `grok-4.5`
- no migration, no default remapping of `grok` / `grok-latest`

Operators who want 4.6 as the implicit default can set `grok_default_text_model`.

## Tests

- `go test -tags=unit ./internal/pkg/xai/ ./internal/service/ ./internal/handler/ ./internal/handler/admin/` (filtered Grok / models cases)
- `vitest` `useModelWhitelist.spec.ts` + `UseKeyModal.spec.ts`

Verified against a live SuperGrok OAuth account: `model=grok-4.6` succeeds upstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -734,7 +734,7 @@ Sub2API supports both Grok subscription accounts through xAI OAuth and standard 
 - Public Claude-compatible target: `/v1/messages`, converted to xAI Responses and returned as Anthropic Messages output for Claude CLI style clients
 - Public Chat Completions targets: `/v1/chat/completions` and `/chat/completions`, forwarded to the account-type-specific xAI upstream
 - Codex CLI style Responses WebSocket ingress is accepted on the Responses targets and bridged to xAI HTTP/SSE Responses upstream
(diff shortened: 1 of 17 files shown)


Yefori-Go:
lgtm

PR #5566, opened 2026-08-12
feat: add batch subscription admin actions

Su-cyber-art:
## Summary

- add cross-page subscription selection and batch actions for validity adjustment, quota reset, revoke, restore, and permanent deletion
- return per-subscription succeeded, skipped, or failed results so stale or incompatible rows do not block a batch
- allow revoked subscriptions to be permanently deleted while retaining usage logs through the existing `ON DELETE SET NULL` relationship
- add idempotency protection, destructive-action confirmation, bilingual UI copy, and backend/frontend coverage

## Validation

- `go test -tags=unit ./...`
- `go test -tags=unit ./internal/service ./internal/handler/admin ./internal/server ./internal/server/middleware`
- `go build ./cmd/server`
- `pnpm lint:check`
- `pnpm typecheck`
- `pnpm test:run` (224 files, 1542 tests)
- `pnpm build`

## Notes

- The focused repository integration test compiles and enters the integration harness locally, but was skipped because Docker is unavailable. CI will run the repository integration suite with PostgreSQL.


Su-cyber-art:
Superseded by #5568 so the source branch can use the conventional feat/admin-subscription-bulk-actions name. The commit and diff are unchanged.

PR #5570, opened 2026-08-12
fix(frontend): show persisted Grok subscription plans

### LINKED ISSUE #5569: bug: Grok subscription badge ignores persisted usage snapshot
## Problem

Grok subscription/plan badges in the admin account list can remain empty or stale even after the backend has persisted quota probe results.

The backend writes quota metadata to `extra.grok_usage_snapshot`, but `AccountsView` reads `extra.grok_quota_snapshot`, a key with no backend writer. The existing frontend fixture uses the same incorrect key, so it does not catch the production mismatch.

Grok-only snapshot updates can also be discarded by incremental account refresh because the row replacement key only tracks runtime fields and OpenAI usage state.

## Expected behavior

- The badge should prefer the billing snapshot plan, then the canonical Grok usage snapshot tier.
- Existing legacy `grok_quota_snapshot` data may remain a fallback, but it should not override canonical data.
- Persisted Grok billing or usage snapshot changes should update the badge on the next normal account-list refresh.
- Malformed imported snapshot values should not reach the string-only badge contract.

## Affected version

Observed on `v0.1.175` (`5935e674a84341c3536e27e6a968384f67d9062b`).

## Scope

This is a frontend projection and incremental-refresh issue. The backend already persists and exposes the canonical snapshots.

X-T-E-R:
## Summary

- read Grok subscription tiers from the backend's canonical `grok_usage_snapshot`
- keep `grok_quota_snapshot` as a legacy fallback without allowing it to override canonical data
- include Grok billing and usage snapshots in incremental account-row refresh decisions
- reject blank or malformed plan values before they reach the badge component
- add regression coverage for canonical/legacy precedence, malformed data, and timer-driven refresh

## Root cause

The backend persists quota headers under `extra.grok_usage_snapshot`, while the account-list badge read `extra.grok_quota_snapshot`. The frontend fixture repeated the incorrect key, masking the mismatch. Incremental refresh also did not compare Grok snapshot state, so a Grok-only update could retain a stale row.

## Verification

- focused Vitest: 2 files, 18 tests passed
- frontend lint passed
- frontend typecheck passed
- CI-critical frontend suite: 13 files, 163 tests passed
- `git diff --check` passed
- frozen install completed with pnpm 9.15.4 and did not modify dependency manifests or the lockfile

The badge updates when the persisted account row is received on the next normal incremental/list refresh; this PR does not patch the probe response directly into account-row state.

Fixes #5569

### DIFF
--- a/frontend/src/utils/__tests__/accountUsageRefresh.spec.ts
+++ b/frontend/src/utils/__tests__/accountUsageRefresh.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest'
-import { buildOpenAIUsageRefreshKey } from '../accountUsageRefresh'
+import { buildGrokUsageRefreshKey, buildOpenAIUsageRefreshKey } from '../accountUsageRefresh'
 
 describe('buildOpenAIUsageRefreshKey', () => {
   it('会在 codex 快照变化时生成不同 key', () => {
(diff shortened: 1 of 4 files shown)


X-T-E-R:

PR #2300, opened 2026-05-08
fix(antigravity): drop .sandbox subdomain on daily-cloudcode-pa base URL

ryneivy:
Fixes #2164: chat completion requests through Antigravity OAuth accounts return the stub "This version of Antigravity is no longer supported" with zero token usage, regardless of client version metadata.

Root cause: ForwardBaseURLs() rotates daily first, and the daily host was set to daily-cloudcode-pa.sandbox.googleapis.com, which is no longer the production daily endpoint and returns the version-gating stub for any request. The actual Antigravity client uses daily-cloudcode-pa.googleapis.com (no .sandbox subdomain).

Verified: claude-opus-4-6, claude-sonnet-4-6, gpt-oss-120b-medium all return real inference (non-zero token usage) after the fix.

PR #5625, opened 2026-08-14
fix(antigravity): use official daily endpoint

sweetcornna:
## Summary

- replace the legacy `daily-cloudcode-pa.sandbox.googleapis.com` host with the current official Antigravity daily endpoint
- add regression coverage so the daily endpoint cannot silently drift back to the sandbox hostname
- keep endpoint-selection behavior unchanged; explicit `GATEWAY_ANTIGRAVITY_FORWARD_BASE_URL=daily` continues to select `BaseURLs[1]`

## Context

The current Antigravity desktop client sends model and quota requests to `daily-cloudcode-pa.googleapis.com`. With a Google AI Pro (`g1-pro-tier`) OAuth account that still had available quota, requests routed to the prod endpoint returned `429 RESOURCE_EXHAUSTED`, while the official daily endpoint returned successful inference.

This is the narrow hostname-alignment part of #5611 and complements #5612, which changes endpoint selection for paid accounts. It overlaps the one-line fix in #2300 but also adds a regression assertion and validation against current v0.1.176 behavior.

## Test plan

- [x] `go test ./internal/pkg/antigravity`
- [x] build a Linux/ARM64 v0.1.176 server with this patch
- [x] deploy with `GATEWAY_ANTIGRAVITY_FORWARD_BASE_URL=daily` and Antigravity UA 2.1.4
- [x] verify `gemini-2.5-flash` through the public gateway returns HTTP 200 instead of 429

Related to #5611, #5537, #2300, and #5612.

### DIFF
--- a/backend/internal/pkg/antigravity/oauth.go
+++ b/backend/internal/pkg/antigravity/oauth.go
@@ -54,7 +54,7 @@ const (
 
 	// Antigravity API 端点
 	antigravityProdBaseURL  = "https://cloudcode-pa.googleapis.com"
-	antigravityDailyBaseURL = "https://daily-cloudcode-pa.sandbox.googleapis.com"
+	antigravityDailyBaseURL = "https://daily-cloudcode-pa.googleapis.com"
 )
 
 var userAgentVersionPattern = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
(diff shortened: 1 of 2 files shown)


sweetcornna:

PR #5641, opened 2026-08-14
fix(openai): preserve remote compaction v2 responses endpoint

### LINKED ISSUE #5624: 远程压缩端点报404
responses/compact 接口报404，无法正常调用远程压缩

InCerryGit:
## Protocol fix

- Replace the beta-feature header gate introduced by `84bb7d070` with the actual remote compaction v2 wire contract: a bare Responses request with `stream: true` and a `compaction_trigger` input item.
- Keep native v2 requests on the streaming `/responses` endpoint with their original body intact.
- Preserve legacy behavior for non-streaming body-signal requests and explicit `/responses/compact` calls, including nested compact subpaths.

## Root cause

Codex remote compaction v2 appends `compaction_trigger` to a normal streaming Responses request. The existing beta-header gate promoted otherwise valid native requests to `/responses/compact`, but that separate legacy endpoint is not available on affected upstreams and returned 404. Endpoint selection now follows the protocol shape instead of optional client metadata.

## Scheduling consistency

- Native v2 requires Responses capability but is independent of the legacy compact probe, `openai_compact_mode`, and `compact_model_mapping`.
- Legacy `/responses/compact` requests retain the existing compact capability tiers, model mapping, and endpoint-specific error.
- Channel restrictions evaluate the same mapped model that `Forward` sends, including `/responses/compact/...` subpaths.
- Passthrough, raw-chat fallback, and profit-control scheduling remain aligned with the selected wire path.

This incorporates and corrects the compact routing and scheduling work from #5580. It does not use the client-version heuristic from #5590.

## Verification

- `go test ./internal/handler -run 'TestNormalizeOpenAIResponsesCompactRequest|TestOpenAIResponsesCompactionRoutingFlags|TestIsOpenAILegacyCompactPath' -count=1`
- `go test ./internal/handler -count=1`
- `go test ./internal/service -run 'Compact|Compaction|OpenAIAccountScheduler|UpstreamModelRestricted' -count=1`
- End-to-end manual compaction on the exact `a8b9ea22b` image: HTTP 200; the compaction assistant finished without error; the session marker and linked checkpoint each increased once; checkpoint items contained `compaction`; no legacy compact path, upstream 404, unsupported error, cancellation, or stream error was observed. For `auto: false`, the response boolean was `false`, which OpenCode 1.18.18 defines as workflow stop rather than compaction failure.
- Production rollout completed in the requested `sub2api-20` then `sub2api-40` order. Both containers run the same immutable image and remain healthy with unchanged PostgreSQL, Redis, and Traefik containers.

Closes #5624

### DIFF
--- a/backend/internal/handler/openai_gateway_compact_body_signal_test.go
+++ b/backend/internal/handler/openai_gateway_compact_body_signal_test.go
@@ -25,40 +25,63 @@ func newCompactBodySignalTestContext(t *testing.T, path string, body []byte) *gi
 
 func TestNormalizeOpenAIResponsesCompactRequest_RemoteV2StaysOnResponses(t *testing.T) {
 	h := &OpenAIGatewayHandler{}
-	body := []byte(`{
-		"model":"gpt-5.6-sol",
-		"stream":true,
-		"store":true,
-		"prompt_cache_key":"pck-signal-1",
-		"reasoning":{"effort":"max","context":"all_turns"},
-		"input":[
-			{"type":"message","role":"user","content":"hello"},
-			{"type":"compaction_trigger"}
-		]
-	}`)
-	c := newCompactBodySignalTestContext(t, "/v1/responses", body)
-	c.Request.Header.Set("x-codex-beta-features", "responses_websockets_v2, remote_compaction_v2, another_feature")
+	tests := []struct {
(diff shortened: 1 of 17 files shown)


InCerryGit:

PR #5169, opened 2026-08-01
feat(admin): automatically retest and recover failed account models

apple-ouyang:
## Summary

- add an `error_recovery` scheduled-test mode with a simple minute interval (5 minutes by default) and an optional advanced 5-field cron schedule
- discover system-recorded account and model errors, test affected models independently, and stop retrying each model as soon as it recovers
- let operators multi-select probe models; error recovery defaults to all available models and stores the full selection canonically as `model_ids: []`
- enable auto recovery by default while preserving an explicit probe-only opt-out, and add click/keyboard-accessible help describing the recovery boundary
- clear only the successful model's unchanged JSON state and immediately synchronize the scheduler snapshot

## Safety boundaries

- only structured errors recorded by the system are eligible
- `AICredits`, manually disabled or unschedulable accounts, expired accounts, elapsed historical windows, and bare `status=error` states are not retried or recovered
- model-level errors always test their actual `model_id`; a non-empty `model_ids` list acts only as a failed-model allowlist
- account-level structured errors are probed once with the first selected model
- model recovery uses a PostgreSQL JSONB compare-and-delete so a concurrent new error or sibling-model update cannot be cleared accidentally
- account-level recovery is also conditional on the observed account version
- each account can have at most one error-recovery plan, and a process-local guard prevents overlapping runs of the same plan

## Database and compatibility

- migration `192_add_scheduled_test_error_recovery.sql` adds the trigger and retry schedule fields and backfills `scheduled_test_results.model_id`
- migration `193_add_scheduled_test_error_model_ids.sql` adds `model_ids TEXT[] NOT NULL DEFAULT '{}'`
- repository writes normalize nil/full selections to an empty PostgreSQL array, never SQL `NULL`
- existing scheduled plans retain their current behavior through the default `scheduled` trigger mode
- no dependencies or additional worker systems are introduced; the existing per-minute scheduled-test runner is reused

## Validation

- `go test ./internal/service ./internal/repository ./internal/handler/admin -count=1` (5,361 tests passed)
- `go test -tags=unit ./internal/service ./internal/repository ./internal/handler/admin -count=1` (8,187 tests passed)
- PostgreSQL integration tests for non-empty and canonical empty `model_ids` create/update round-trips (2 tests passed)
- focused PostgreSQL integration coverage for conditional per-model recovery, account recovery, sibling-model success, and same-reset rewrites
- `ScheduledTestsPanel.spec.ts` (8 tests passed) and `ModelMultiSelect.spec.ts` (1 test passed)
[...]

PR #5014, opened 2026-07-28
feat(admin): show token throughput in usage records

apple-ouyang:
## Summary

- add a dedicated `Token/s` column to the admin usage records table
- calculate output throughput as `output_tokens * 1000 / duration_ms`
- render `-` for invalid/zero durations, non-positive output tokens, and image-generation rows

This is intentionally limited to the admin usage page. It does not change the shared user usage table or the backend/database contract.

## Relation to #4453

#4453 adds throughput inside a shared token detail tooltip. This PR keeps the requested admin record metric as an independently scannable column and avoids changing the shared component.

## Verification

- `pnpm --dir frontend exec vitest run src/views/admin/__tests__/UsageView.spec.ts` (11/11)
- `pnpm --dir frontend exec eslint src/views/admin/UsageView.vue src/views/admin/__tests__/UsageView.spec.ts`
- `pnpm --dir frontend run typecheck`
- `pnpm --dir frontend run build`
- local browser verification on `/admin/usage`, with no console errors

Closes #5012

PR #5643, opened 2026-08-14
fix(openai): persist model circuit after consecutive failures

apple-ouyang:
## Summary

- persist a 10-minute account-and-model circuit after three distinct requests receive transient upstream failures
- deduplicate retries by request ID and keep transient 400, 401/403, 429/529, and pool-mode handling on their existing paths
- preserve failures that arrive during recovery with a pre-DB generation snapshot plus the existing database value CAS

## Behavior

The durable circuit applies to `500/502/503/504/520/521/522/523/524`, scoped by account ID and canonical model. Internal retries for one request count once. Scheduled recovery clears the circuit only when both the observed database limit and in-memory mutation generation are unchanged.

## Tests

- `go test -tags unit ./internal/service -count=1` (7742 passed)
- focused race tests for transient circuit and recovery paths (28 passed)
- deterministic regression coverage for failures during `GetByID` and canonical model case normalization
- `git diff --check`

## Stack

This PR is stacked on the local patch chain and currently overlaps #5014, #5169, and #5432. The new review boundary is commit `d78f27cb377f0bbddb35c68aff785c5df8dd9762` over `d00f5711a45c38687bb0376fbe097dd891cb3f63`; the cumulative GitHub diff will shrink as those dependencies land.

### DIFF
--- a/PROGRESS.md
+++ b/PROGRESS.md
@@ -23,14 +23,14 @@
 - ownership: backend paths declared by the approved Spec
 - worktree/branch: `/Users/admin/code/opensource/sub2api/.worktrees/openai-fallback-only-sticky-backend` / `codex/openai-fallback-only-sticky-backend`
 - worker: `backend_worker`
-- reviewer: `/root/frontend_review`
+- reviewer: `/root/backend_review`
 - started_at: `2026-08-08T10:45:00Z`
-- finished_at: pending
-- verification: pending RED/GREEN and focused tests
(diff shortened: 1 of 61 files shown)


apple-ouyang:
Closing this stacked PR until its prerequisite patches land. The branch and worktree are retained; the circuit-breaker patch will be rebuilt on the latest upstream main as a standalone PR afterward.

PR #5656, opened 2026-08-14
feat(i18n): add Spanish (es) locale

albert210596:
## Summary

Adds a complete **Spanish (`es`)** locale to the panel, bringing the locale count to three (`en` / `zh` / `es`).

Spanish is the second-most-spoken native language in the world, and the panel had no Latin-script locale other than English. This makes Sub2API usable for operators and end users across Spain and Latin America without relying on browser auto-translation (which mangles the tables, tooltips and the driver.js onboarding tour).

## What's in it

- **`frontend/src/i18n/locales/es/`** — all 16 message files, mirroring the `en/` tree file for file and key for key.
- **`frontend/src/i18n/index.ts`** — `es` registered in `LocaleCode`, `localeLoaders`, `isLocaleCode`, browser-language detection (`navigator.language` starting with `es`) and `availableLocales`. `LocaleSwitcher` picks it up automatically from `availableLocales`, so no component change was needed.
- **`frontend/src/i18n/__tests__/localesMessageCompile.spec.ts`** — `es` added to the precompile matrix, so an unescaped `{` in Spanish copy fails in CI instead of blowing up a component tree at runtime.
- **`DateRangePicker.vue` / `KeyUsageView.vue`** — these two branch on `locale.value === 'zh' ? 'zh-CN' : 'en-US'` for `toLocaleDateString`; Spanish now maps to `es-ES` instead of silently falling back to `en-US`.

## Translation notes

- Every key present in `en` exists in `es`, and no key exists in `es` that is not in `en` — checked programmatically over the flattened trees (**8,871 keys**, exact match on both sides).
- All `{placeholder}` tokens were diffed key-by-key against `en` — **identical in every string**. Escaped literals (`{'@'}`, `{'{'}billing_header{'}'}`), pluralization pipes and the HTML inside the onboarding-tour strings were preserved.
- Product and protocol nouns are deliberately left untranslated (OAuth, Setup Token, Refresh Token, API key, endpoint names, model IDs, `x-api-key`, RPM/RPD, Claude Code, Codex, Bedrock, Vertex…), matching how these terms are actually used by Spanish-speaking operators.
- Spanish uses neutral, non-regional wording (`ordenador`/`computadora` and similar splits avoided) so it reads correctly on both sides of the Atlantic.

## Verification

- `vue-tsc --noEmit` — clean.
- `vitest run src/i18n/__tests__/localesMessageCompile.spec.ts` — 3/3 pass (`zh`, `en`, `es`).
- Full `vitest run` — same result as the pre-change baseline on the same machine (the handful of failures there reproduce on unmodified `main` and are unrelated to i18n).
- Built the production Docker image from this branch and confirmed the Spanish strings are present in the embedded frontend, then loaded the panel and switched to **Español** in the language menu.

Happy to adjust any wording if maintainers or other Spanish-speaking users prefer different terminology.


albert210596:

PR #5658, opened 2026-08-14
fix(openai): honor OpenCode Go usage reset durations

william-drakemond:
## Summary

- Recognize OpenCode Go `GoUsageLimitError` responses as account usage limits.
- Parse reset durations from messages such as `Resets in 2 days` and composite values such as `Resets in 4hr 59min`.
- Keep the existing 429 fallback behavior for malformed payloads and unrelated error types.

## Problem

OpenCode Go usage-limit responses do not provide the reset through the OpenAI fields already handled by sub2api, such as `resets_at` or `resets_in_seconds`. The parser therefore returned no reset time. API-key accounts then fell through to the configured 429 cooldown, which defaults to 5 seconds, became schedulable again, and repeatedly hit the same exhausted account before its quota actually reset.

OpenCode rolling-window messages can contain more than one duration component. Parsing only the first component would turn `4hr 59min` into `4h` and release the account up to 59 minutes too early.

## Fix

`parseOpenAIRateLimitResetTime` now recognizes `GoUsageLimitError` and, only for that error type, parses the reset duration from its human-readable message.

The duration parser:

- Requires the `Resets in` marker.
- Supports seconds, minutes, hours, days, and weeks, including common abbreviations.
- Accumulates contiguous components such as `4hr 59min`.
- Rejects non-positive, malformed, and overflowing values so the established fallback remains in control.

The computed timestamp flows through the existing 429 path. That path persists the rate limit with `SetRateLimited` and notifies the runtime blocker, so API-key accounts stay out of scheduling until the parsed reset time. Unknown error types continue to return no parsed reset.

## Tests

Added coverage for:

- Day and hour reset messages.
- Composite hour/minute reset messages.
- Rejection of unrelated error types.
- The full API-key 429 path, including persistence and runtime blocking.

Validated with:

- `golangci-lint v2.9.0 run ./internal/service --timeout=10m` (`0 issues`)

### DIFF
--- a/backend/internal/service/openai_account_runtime_block_fastpath_test.go
+++ b/backend/internal/service/openai_account_runtime_block_fastpath_test.go
@@ -114,6 +114,33 @@ func TestOpenAI429RetryDelayHonorsBoundedRetryAfter(t *testing.T) {
 	require.Equal(t, openAIOAuth429MaxRetryDelay, openAIOAuth429SameAccountRetryDelay(http.Header{"Retry-After": []string{"90"}}, deadline))
 }
 
+func TestOpenAI429FastPath_OpenCodeGoUsageLimitUsesMessageResetDuration(t *testing.T) {
+	repo := &rateLimit429AccountRepoStub{}
+	rateLimitService := NewRateLimitService(repo, nil, &config.Config{}, nil, nil)
+	svc := &OpenAIGatewayService{rateLimitService: rateLimitService}
+	rateLimitService.SetAccountRuntimeBlocker(svc)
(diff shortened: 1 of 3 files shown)


william-drakemond:

Wei-Shaw:
维护审计反馈：修复逻辑本身有价值，现阶段主要阻塞是分支基线过旧。当前 backend-security 和 frontend-security 的失败来自旧依赖/旧安全基线，而 main 已更新对应版本。

请将分支 rebase 到最新 main（或合并最新 main），解决依赖与审计基线差异后重新触发全套 CI。完成后可按新结果快速复核。

PR #5664, opened 2026-08-15
fix(ops): score TTFT against a band that fits reasoning traffic

ngcongtunglam209:
## Problem

The ops dashboard reported two findings that looked independent:

```
Time to first token elevated (4741ms)
Overall health score low (61)
```

They are the same finding. `first_token_ms` is stamped at the first *visible* output event (`openai_gateway_response_handling.go:589` via `openAIStreamDataStartsVisibleOutput`), so for reasoning models it includes the whole thinking phase whenever the upstream sends no reasoning-summary deltas. `computeBusinessHealth` graded that number on a 1s→3s band and zeroed `ttftScore` above 3s.

Production traffic cannot satisfy that band. Measured over 24h:

| reasoning_effort | p50 range |
|---|---|
| (none) | 932–1323ms |
| medium | 1554–3604ms |
| high | 1289–5854ms |
| xhigh | 3594–5187ms |
| max | 6609–8037ms |

`ttftScore` was therefore pinned at 0, capping the overall score at 65 regardless of relay behaviour. The score was measuring its own calibration.

## Change

- TTFT band 1s→3s becomes **3s→30s**
- Grade **p95** instead of p99 (hourly buckets with a dozen samples make p99 indistinguishable from max, so one slow request swung the score); p99 kept as fallback
- Smart Diagnosis panel threshold 500ms→3000ms on the same statistic — the old trigger fired permanently
- `ttftHighAction` copy corrected in en + zh: it blamed pre-processing time for a metric dominated by upstream thinking

## Verification

- `go test -tags unit -run 'TestOps|TestCompute|TestDashboard' ./internal/service/ ./internal/repository/` — pass
- `go vet ./internal/service/` — clean
- `vue-tsc --noEmit -p tsconfig.json` — clean
- `vitest run src/i18n/__tests__` — 56 pass

Against real data, p95 4741ms now scores **94**, while the two observed incident windows (p99 ~69s) still drive `ttftScore` to 0 — the band keeps catching real events without flagging normal reasoning traffic.

## Not in this PR

Two per-account anomalies surfaced during the investigation and are worth a separate look:

- Accounts 3 and 7 are ~4x slower than accounts 2/6/14 on `gpt-5.6-sol` at the *same* `reasoning_effort=high` (p50 5290/5854ms vs 1289–1689ms, n=110/175)
- Account 14 shows a p95 of 47818ms at `xhigh` against account 2's 9937ms, matching an incident window on `channel_id=1`


ngcongtunglam209:
Opened against the wrong base repository by mistake — this belongs on my fork. Closing; sorry for the noise.

PR #5671, opened 2026-08-15
fix gpt-56-sol wm

jianjianai:


jianjianai:

LingyeNBird:
Hi! Nice catch on preserving the `-wm` variant — the normalization change itself looks correct.

I tested it against the billing path and found one gap: `normalizeKnownOpenAICodexModel("gpt-5.6-sol-wm")` now returns `""`, which makes `getFallbackPricing` (billing_service.go) miss the model entirely — it falls through the switch and returns `nil`, surfacing as `ErrModelPricingUnavailable`. So requests to `gpt-5.6-sol-wm` would be billed at nothing / fail cost calculation.

The fix is to add a price-card fallback alongside the normalization change:

```go
// billing_service.go, getFallbackPricing
switch normalized {
case "gpt-5.6-sol", "gpt-5.6-sol-wm":
    return s.fallbackPrices["gpt-5.6-sol"]
}
```

I implemented both parts (normalization + billing fallback + default model list) in my fork if you want to compare or cherry-pick:
https://github.com/LingyeNBird/sub2api/tree/feat/gpt-5.6-sol-wm

Happy to adjust if you prefer a different shape.

jianjianai:
确实，计费也应该补一下

gaoconggit:
大佬们，加油

jianjianai:
https://github.com/Wei-Shaw/sub2api/pull/5672

PR #5687, opened 2026-08-16
fix(openai): complete Codex state and compact probing

lyy0709:
## Summary

Follow-up to #5668, limited to the Codex state-machine, remote-compaction capability probe, and beta/default upgrade gaps that remained after that PR.

- bind each exact opaque `x-codex-turn-state` value to the OAuth account that actually delivered it, persist only a scoped digest across instances, and strip only known cross-account echoes
- record provenance only after the downstream header/body write succeeds; cover native and passthrough, streaming and non-streaming, including already-committed compact keepalives
- keep OpenAI API-key traffic transparent: relay the header but never create or enforce OAuth provenance
- make the compact probe follow Codex RemoteCompactionV2 collection semantics: require `response.completed`, count only `response.output_item.done`, require exactly one compaction item, permit unrelated done items, and ignore bytes after the first terminal
- preserve the official `compaction_summary` serde alias, while removing added-only fallback and compaction deduplication that could hide malformed streams
- add tri-state, versioned, expiring probe snapshots; bounded response reads; cancellation-detached persistence; monotonic database updates; and a migration that invalidates snapshots created by the old permissive probe
- normalize beta-feature headers case-insensitively and preserve native-v2/session defaults without changing API-key Responses routing

Official Codex reference audited at `49db349ffdd888f5e3c91abf9b7519d8631e6e9a`:
- [RemoteCompactionV2 collector](https://github.com/openai/codex/blob/49db349ffdd888f5e3c91abf9b7519d8631e6e9a/codex-rs/core/src/compact_remote_v2.rs)
- [RemoteCompactionV2 request construction](https://github.com/openai/codex/blob/49db349ffdd888f5e3c91abf9b7519d8631e6e9a/codex-rs/core/src/compact_remote_v2_attempt.rs)
- [Compaction response item aliases](https://github.com/openai/codex/blob/49db349ffdd888f5e3c91abf9b7519d8631e6e9a/codex-rs/protocol/src/models.rs)

## Regression coverage

New and expanded tests cover:

- exact opaque state values, same-value recommit, API-key/session isolation, cross-instance lookup, cancellation, write failures, keepalive-committed headers, and all four native/passthrough stream modes
- LF/CRLF/lone-CR SSE, BOM, comments, multi-line data, no-colon fields, EOF dispatch, `[DONE]`, terminal ordering, added-only output, duplicate compaction items, aliases, incomplete/truncated/oversized/read-error bodies, and 2xx/error status matrices
- supported/unsupported/unknown transitions, protocol version/TTL/skew boundaries, concurrent 24-writer monotonic convergence, unknown diagnostics, duplicate-account cleanup, and migration behavior/idempotence
- beta token casing, duplicate removal, override preservation, native-v2 forcing, OAuth defaults, API-key transparency, and probe identity parity

## Verification

- `cd backend && go test ./... -count=1`
- `cd backend && go test -tags=unit ./... -count=1`
- `cd backend && go test -tags=integration ./... -count=1`
- targeted `go test -race` for the changed service/repository paths
- `cd backend && golangci-lint run ./...`
- `make test`
- `pnpm --dir frontend run test:run` (223 files, 1547 tests)
- `pnpm --dir frontend run test:coverage` (223 files, 1547 tests)
- `make build`

No frontend, dependency, lockfile, or version metadata changes are included.


lyy0709:
Verification update for head `91f59a7f89eb166bfe870d430b375801f0f1ec2c`:

- [CI passed](https://github.com/lyy0709/sub2api-upstream/actions/runs/31922375913): shell, frontend, golangci-lint, Go unit tests, and Go integration tests all succeeded.
- [Security Scan passed](https://github.com/lyy0709/sub2api-upstream/actions/runs/31922375899).

The corresponding upstream runs are currently marked `action_required` because fork workflow execution needs maintainer approval.

PR #5733, opened 2026-08-17
fix(grok): strip all explicit nulls on Responses input items

Barrierml:
## Summary
xAI still returns 422 `data did not match any variant of untagged enum ModelInput` on large Grok `/v1/responses` requests after #4239.

#4239 only deleted `content: null` on `type=reasoning`. Production still saw this 422 on `POST /v1/responses` `model=grok-4.6` (account 117) after that sanitizer was already deployed. The remaining failures look like other explicit-null fields on other input item types.

This widens the existing sanitizer to delete every top-level JSON `null` on Responses `input` objects, while keeping non-null values.

## Test plan
- [x] `cd backend && go test -tags=unit ./internal/service -count=1 -run TestPatchGrokResponsesBody`
- [x] Existing reasoning `content: null` cases still pass
- [x] New case: `function_call` / `function_call_output` / `web_search_call` `status: null` is stripped
- [x] Non-null reasoning `content` is preserved

## Notes
This change is AI-assisted.

### DIFF
--- a/backend/internal/service/openai_gateway_grok.go
+++ b/backend/internal/service/openai_gateway_grok.go
@@ -822,14 +822,14 @@ func sanitizeGrokResponsesModelInput(body []byte) ([]byte, error) {
 			filtered = append(filtered, json.RawMessage(item.Raw))
 			continue
 		}
-		itemType := strings.TrimSpace(item.Get("type").String())
-		if itemType != "" {
-			if _, ok := grokResponsesSupportedInputTypes[itemType]; !ok {
-				changed = true
-				continue
-			}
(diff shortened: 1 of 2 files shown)


Barrierml:

Follow-up commit `285bd350c` is now on this branch:

- recursively strip nested JSON nulls inside Responses `input`
- drop OpenAI-only item types such as `item_reference` / `custom_tool_call`
- log item-type counts (no request body) when xAI still returns ModelInput 422

Production still saw the same 422 on ~251KB `grok-4.6` `/v1/responses` after the first top-level-null sanitizer.

PR #5808, opened 2026-08-19
fix(openai): decouple compact scheduling from ordinary quota

jinyaxuan:
## Summary

Legacy `/responses/compact` requests can be served by an account during the ordinary Codex quota/runtime window. The scheduler previously discarded those accounts before compact capability checks and lost `RequireCompact` across sticky, fresh, database, and `previous_response_id` rechecks.

## Changes

- Keep lifecycle, expiry, overload, temporary-block, model-level rate-limit, endpoint capability, parent-health, channel, threshold, proxy, and profit gates while bypassing ordinary account-wide quota/runtime gates only for compact.
- Query persistent `ListModelAvailabilityCandidates` for compact so transient quota state does not empty the candidate pool.
- Thread compact intent through legacy load-batch/fallback paths, advanced scheduler rechecks, sticky sessions, and `previous_response_id` routing.
- Defer compact-tier filtering until fresh/DB state is available, preserving `ErrNoAvailableCompactAccounts`.
- Avoid stale scheduler snapshot hydration after compact DB selection and add regression coverage across legacy/advanced and load-batch modes.

## Validation

- Targeted service regression: `go test ./internal/service -run 'Test(OpenAIAccountScheduler|OpenAIGatewayService_SelectAccountWithScheduler|OpenAICompact|AdvancedScheduler|OpenAIGatewayService_SelectAccountByPreviousResponseID)' -count=1`\n- The full service package was attempted; an unrelated proxy-quality test requires an IPv6 listener unavailable in the sandbox (`httptest` bind to `[::1]:0`).\n- The runtime block store currently does not retain block reasons, so compact bypasses all in-memory runtime blocks, not only 429; persistent expiry/overload/temp/model-level gates remain enforced.\n\nDownstream edge, anti-ban, branding, and CI changes are intentionally excluded.

### DIFF
--- a/backend/internal/service/antigravity_quota_scope.go
+++ b/backend/internal/service/antigravity_quota_scope.go
@@ -51,6 +51,26 @@ func (a *Account) IsSchedulableForModelWithContext(ctx context.Context, requeste
 	return true
 }
 
+// IsSchedulableForCompactModelWithContext keeps legacy /responses/compact
+// independent from the ordinary account-wide quota window while retaining
+// lifecycle, overload, temporary-block, and model-level checks.
+func (a *Account) IsSchedulableForCompactModelWithContext(ctx context.Context, requestedModel string) bool {
+	if a == nil || !a.IsActive() || !a.Schedulable {
+		return false
+	}
+	now := time.Now()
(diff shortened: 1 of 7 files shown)


jinyaxuan:

PR #5809, opened 2026-08-19
test(tlsfingerprint): require opt-in for external checks

riven9566-source:
## Summary

- share the existing external-network test gate across unit and integration build tags
- require `TLSFINGERPRINT_NETWORK_TESTS=1` before the TLS fingerprint integration tests contact `tls.peet.ws`

## Why

The integration suite contacted a third-party fingerprint service by default even though these tests are documented as manual network checks. That makes otherwise unrelated integration verification depend on the service's current TLS certificate and availability.

Deterministic ClientHello/profile tests continue to run normally. Explicitly enabling the network checks still performs the real request with normal certificate verification.

## Validation

- `go test -count=1 -tags=unit ./internal/pkg/tlsfingerprint`
- `go test -count=1 -tags=integration ./internal/pkg/tlsfingerprint`
- `git diff --check`

No production code changed, and this does not use `InsecureSkipVerify` or otherwise weaken TLS verification.


riven9566-source:
Withdrawn: this test-only proposal is no longer needed.

PR #5811, opened 2026-08-19
test: align group rollup fixture timezone

riven9566-source:
## Summary

- align the group-usage rollup integration fixture's PostgreSQL session timezone with its default `Asia/Shanghai` rollup state
- keep the existing UTC and DST-specific cases free to override the transaction timezone with `SET LOCAL`

## Root cause

The integration harness opens PostgreSQL sessions in UTC, while two rollup-trigger fixtures initialize and assert their watermark using the Shanghai calendar date. During the first eight hours of each Shanghai day, the trigger therefore sees the previous UTC date and the tests fail by one day.

This is a fixture-only correction. Production code, migrations, schema behavior, and TLS/runtime settings are unchanged. The application production connection already sets PostgreSQL `TimeZone` from the configured server timezone.

## Validation

- all `TestGroupUsage` repository integration tests pass, including the two midnight-boundary cases and the DST override case
- related service and repository GroupUsage unit tests pass
- `git diff --check` passes


riven9566-source:
Withdrawn: this test-only proposal is no longer needed.

PR #5838, opened 2026-08-19
fix(frontend): align admin role selector styling

xuhaihan:
## Summary

- replace the native role selector with the shared Select component
- align admin user editing with the existing form styling and dark mode behavior

## Validation

- `cd frontend && corepack pnpm test:run`
- `cd frontend && corepack pnpm build && corepack pnpm lint`


xuhaihan:
recheck

PR #5848, opened 2026-08-19
fix(openai): strip ChatGPT-internal-unsupported fields at the upstream egress

live400:
Fixes the 400 reported in #5803.

## What we observed

On one gateway running v0.1.178, **119 distinct client requests in 8 hours** were rejected by
`chatgpt.com/backend-api/codex/responses` with:

```json
{"error":{"code":"invalid_parameter",
          "message":"prompt_cache_retention is not supported on this model",
          "param":"prompt_cache_retention","type":"invalid_request_error"}}
```

That was **61% of all client-visible failures** on that gateway. A parameter rejection cannot be
rescued by switching accounts, so every one of them reached the client as a 400. 109 hit
`/v1/responses` and 10 hit `/v1/chat/completions`.

## Why the existing strips did not catch it

The field lists are not missing anything — that part of #5803's analysis is correct. v0.1.178
already strips this field in five places, most unconditionally, and the affected accounts take a
path that does:

| Site | Gate |
|---|---|
| `openai_codex_transform.go` (`openAICodexOAuthUnsupportedFields`) | OAuth only, unconditional |
| `request_body.go` ← `passthrough.go` | passthrough + OAuth, unconditional |
| `openai_gateway_forward.go` | **inside `if !isCodexCLI`** |
| `chat_completions.go` (`cursorResponsesUnsupportedFields`) | Responses-shape short-circuit |
| `grok.go` | grok |

We also ruled out, on the affected accounts: no `base_url`/`pool_mode` (talking to chatgpt.com
directly, not a second-tier relay), passthrough disabled, and **not a retry-path regression** —
each `client_request_id` logged exactly one 400, so the field was present on the very first
attempt. We also checked the `if requestView.HasPatches()` branch in `Forward` that discards the
decoded map: `HasPatches()` is `!patchesDisabled && len(patches) > 0` and `markDecodedModified()`
calls `disablePatch()`, so the decoded modifications are not lost there.

Request bodies are not persisted, so we could not determine statically which branch let the field
through. Rather than keep guessing at branches, this collapses the cleanup to the one point every
request must pass, and adds a diagnostic so the leaking branch can be identified from production.

## Changes

1. **Egress strip.** `applyChatGPTInternalEgressStrip` is called from both upstream request
   builders (`buildUpstreamRequest` and `buildUpstreamRequestOpenAIPassthrough`) right before
   `http.NewRequest`, next to the existing `normalizeDeepSeekResponsesRequestBody` egress
   normalization. It applies **only to OpenAI OAuth accounts** — API-key accounts talk to the
   Platform API or third-party compatible upstreams where these fields are legal. The field list
   reuses `openAIChatGPTInternalUnsupportedFields` so there is still one source of truth.

2. **Diagnostic.** If the egress actually removes something, an upstream branch leaked, so it logs
   at Warn with the egress name and the field names — **never values and never the request body**.

3. **Retry on `invalid_parameter`.** The rejected-field retry already existed, but
   `isExplicitOpenAIResponsesFieldRejection` only accepts code `unknown_parameter` /
   `unsupported_parameter` or a message containing "unknown/unsupported parameter". The real
   response is `invalid_parameter` with "X is not supported on this model", so both gates missed
   it; and its param handling only knew `max_output_tokens` and `input[N].namespace`. Rejections
   of whitelisted optional fields now retry once regardless of code. Safety comes from the
   whitelist rather than the code: it holds only fields with no generation semantics
   (`prompt_cache_retention`, `prompt_cache_options`, `safety_identifier`, `stream_options`,
   `metadata`, `user`), so a rejected `model` or `input[74].name` is still reported to the client
   verbatim instead of being silently rewritten. A nested param such as
   `prompt_cache_options.ttl` strips the whole top-level object.

Deliberately **not** done: normalizing to `prompt_cache_options.ttl = "30m"`. `forward.go`
historically stripped `prompt_cache_options` alongside `prompt_cache_retention`, which suggests
the internal endpoint rejected it too; adding a new field without confirming internal support
would be an experiment run in production. #5803 makes the same point about the public Platform
API docs not proving internal Codex support.

## Incidental finding

**API-key account + Codex CLI is a real blind spot** in the current cleanup:
`applyCodexOAuthTransform` only covers OAuth, and the `forward.go` strip sits under
`if !isCodexCLI`, so neither covers that combination. This surfaced while writing the retry test —
using a `curl/8.0` UA never reproduced, because that goes down the `!isCodexCLI` branch where the
field is already gone. Reproducing this class of leak requires a Codex CLI user agent. That
combination is now covered by the retry.

## Tests

`openai_upstream_egress_strip_test.go`, 9 cases. The egress case builds its probe body **from
`openAIChatGPTInternalUnsupportedFields` itself**, so adding a field to that list cannot silently
go untested; an explicit `prompt_cache_retention` case is kept pinned alongside it. There is an
end-to-end case for OAuth + Codex CLI + GPT-5.6, a case feeding a dirty body straight into
`buildUpstreamRequest` to simulate "every branch leaked", and negative cases asserting that
semantic params and `input[74].name` are untouched.

`go test ./internal/service/ -run 'OpenAI|Codex|Passthrough|Cursor|Grok'` passes on top of current
main.

## Not covered

WS v2 builds its own upstream target (`buildOpenAIResponsesWSURL` + ws payload) and does not go
through `buildUpstreamRequest`, so the egress strip does not apply there; that path already has a
test asserting the field is stripped. The passthrough path deliberately does not get the retry —
passthrough means proxying the client's request as-is, and rewriting it before retrying would
break that contract; the egress strip already covers OAuth there.

### DIFF
--- a/backend/internal/service/openai_gateway_forward.go
+++ b/backend/internal/service/openai_gateway_forward.go
@@ -1073,6 +1073,10 @@ func (s *OpenAIGatewayService) buildUpstreamRequest(ctx context.Context, c *gin.
 	// previous_response_id，避免携带状态字段被上游拒绝。
 	body = normalizeDeepSeekResponsesRequestBody(account, body)
 
+	// 出口兜底：ChatGPT internal 不支持的顶层字段在此做最后一次清理，防止
+	// 上游任一转换分支遗漏或在清理后重建 body 导致 invalid_parameter 400。
+	body = applyChatGPTInternalEgressStrip(ctx, account, body, openAIEgressForward)
+
 	req, err := http.NewRequestWithContext(ctx, "POST", targetURL, bytes.NewReader(body))
 	if err != nil {
 		return nil, err
--- a/backend/internal/service/openai_gateway_passthrough.go
+++ b/backend/internal/service/openai_gateway_passthrough.go
@@ -484,6 +484,11 @@ func (s *OpenAIGatewayService) buildUpstreamRequestOpenAIPassthrough(
 	// DeepSeek 原生 Responses 端点为无状态实现（见 normalizeDeepSeekResponsesRequestBody）。
 	body = normalizeDeepSeekResponsesRequestBody(account, body)
 
+	// 出口兜底：见 applyChatGPTInternalEgressStrip。透传路径虽已在
+	// normalizeOpenAIPassthroughOAuthBody 清理过，但其后还有多处 body 重建
+	// （工具适配、指纹改写、模型改写），这里做最终保证。
+	body = applyChatGPTInternalEgressStrip(ctx, account, body, openAIEgressPassthrough)
+
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, targetURL, bytes.NewReader(body))
 	if err != nil {
 		return nil, err
--- a/backend/internal/service/openai_responses_rejected_field_retry.go
+++ b/backend/internal/service/openai_responses_rejected_field_retry.go
@@ -17,7 +17,28 @@ const maxOpenAIResponsesRejectedFieldRetries = 6
 var (
 	openAIResponsesRejectedNamespaceParamPattern = regexp.MustCompile(`(?i)^input\[(\d+)\]\.namespace$`)
 	openAIResponsesRejectedMessageParamPattern   = regexp.MustCompile(`(?i)(?:unknown|unsupported)[ _-]+parameter\s*(?::|=|is)?\s*["']?(max_output_tokens|input\[\d+\]\.namespace)(?:["']|\b)`)
+	// ChatGPT internal Codex 端点点名拒绝顶层字段时用的是另一种文案，例如
+	// "prompt_cache_retention is not supported on this model"（code 为
+	// invalid_parameter），与上面的 "unknown/unsupported parameter: X" 形态不同。
+	openAIResponsesRejectedUnsupportedOnModelPattern = regexp.MustCompile(`(?i)^([a-z0-9_]+)\s+is\s+not\s+supported\s+(?:on|with|for|by)\s+this\s+model`)
 )
+
+// openAIResponsesStrippableRejectedParams 是被上游以 400 点名拒绝后，可以安全剥离
+// 并重试一次的顶层可选字段。
+//
+// 判据是「删掉它等价于客户端从没发过这个字段」：这些字段只影响提示缓存策略、
+// 调用方标识与用量统计口径，不参与生成语义，剥掉不会改变模型输出。
+//
+// 刻意不含 model / input / instructions / tools / reasoning 等语义关键字段——
+// 那些被拒时必须把错误如实返回客户端，而不是悄悄改写请求再重试。
+var openAIResponsesStrippableRejectedParams = map[string]struct{}{
+	"prompt_cache_retention": {},
+	"prompt_cache_options":   {},
+	"safety_identifier":      {},
+	"stream_options":         {},
+	"metadata":               {},
+	"user":                   {},
+}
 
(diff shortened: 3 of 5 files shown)


Wei-Shaw:
## 代码审计发现（自动化审计 + 独立复核确认）

出口剥离本身经得起反驳（账号门控与 targetURL 选择逐行同构、透传契约未破坏、`sjson.DeleteBytes` 非原地改写故调用方 body 不污染、5 个剥离字段逐一核对无误伤 `prompt_cache_key`）。以下两点需合并前知情（均经独立复核 CONFIRMED）。

### 🟡 需决策

**M1：retry 剥离不按账号类型门控，且绕过 code 白名单**
新流程把 `stripRejectedOptionalParam` 提到 `isExplicitOpenAIResponsesFieldRejection`（code 判据）之前，`normalizeOpenAIResponsesRejectedFieldRetryBody(statusCode, body, responseBody)` 无 account 参数（调用点 `openai_gateway_forward.go:920` 也不传）。后果：**API-key 账号打第三方上游时**，任何 400 只要 `error.param` 命中 6 元白名单（`user`/`metadata`/`safety_identifier`/`stream_options`/`prompt_cache_*`），哪怕原因是取值格式非法（`invalid_value`）而非字段不支持，网关都会静默删掉该字段同号重试——抹掉 `user`/`safety_identifier` 的滥用归因，并把客户端的真实参数错误掩盖成成功。
PR 自带测试直接坐实：`newOpenAIRejectedFieldTestAccount()` 是 APIKey + 第三方 base_url；另一负面用例的响应体 `code:"invalid_value"` 说明生产上存在这类 400，它不被剥离的唯一原因是 param 不在白名单里、而非 code 不对。
**修法**（复核修正）：不是加 account 门控（该 retry 的目标场景就是 APIKey+Codex CLI 盲区，OAuth-only 会废掉 PR 目的），而是加 **code/文案判别**——要求 code ∈ {invalid_parameter, unknown_parameter, unsupported_parameter} 且文案含 "not supported"，把取值错误挡在外面如实回传客户端。

**M2：名单漂移，`prompt_cache_options` 三处不同源**
retry 白名单与 `openai_gateway_forward.go:496` 的既有剥离都含 `prompt_cache_options`，但单一事实来源 `openAIChatGPTInternalUnsupportedFields` 与 `openAICodexOAuthUnsupportedFields` 都不含它；而 `:496` 挂在 `!isCodexCLI` 下。净结果：**OAuth + Codex CLI 组合下该字段飞行前无处剥离、新出口也兜不住**，只能靠 400 后 retry 白付一次往返。与"清单收敛到唯一出口"的表述不符。

### 需知情

**CI 从未在当前 head 跑过**。本 PR 来自 fork，`gh api runs?head_sha=6fe7c6469` 显示 CI 与 Security Scan 均 `action_required`（等待批准）、CLA Assistant `failure`，`mergeStateStatus=UNSTABLE`。5 个新测试无任何真实执行记录，PR 描述的本地测试通过无 CI 佐证。

### 🔵 跟进

- **L1** 出口剥离不改调用方 body，`stripRejectedOptionalParam` 又用该 body 判存在性——OAuth 请求因非该字段原因收到带白名单 param 的 400 时，会剥一个上游没收到的字段并重试，第二次 wire body 逐字节相同必然再失败（受 body 哈希去重与 6 次上限护栏，仅浪费一次往返）。
- **L3** 出口对每个 OAuth 请求循环 5 次 `gjson.GetBytes(...).Exists()`，稳态下每请求 5 遍全 body 扫描；可降为单次 `ForEach`。
- 与 #5842 无冲突：OpenAI OAuth 账号结构上不可能 adaptive（`GetAPIProtocol` 有 `!IsCNProvider()` 早退），不存在绕过出口的组合。

PR #5889, opened 2026-08-20
feat(openai): add strict raw Responses forwarding mode

heathermhuang:
## Summary

Add an opt-in `strict_raw` mode for OpenAI API-key upstreams on `/v1/responses` and `/v1/responses/compact`. This supports Responses-native upstreams that need the original request and JSON/SSE response bytes preserved, while retaining Sub2API's authentication, scheduling, concurrency and usage handling.

- Branch before compatibility transformations; preserve the requested model and response bytes.
- Support Bearer authentication or an explicit private-upstream `none` setting, with account validation and UI controls.
- Filter downstream credentials, cookies and transport headers; keep strict mode HTTP/SSE-only.
- Observe usage and response IDs without rewriting payloads.
- Preserve existing normal and legacy passthrough HTTP continuations. Strict continuations retain a separate binding to the original strict account and cannot fall through to a different or reconfigured account.

The implementation is generic. Browser sessions, consumer authentication and bridge automation remain outside Sub2API. Setup references: [strict upstream guide](https://github.com/Wei-Shaw/sub2api/blob/33f746fbda14311fb806f57cbba654059b6e7af0/docs/openai-strict-responses-upstream.md) and [web bridge example](https://github.com/Wei-Shaw/sub2api/blob/33f746fbda14311fb806f57cbba654059b6e7af0/docs/openai-web-bridge-quickstart.md).

## Verification

- Regression tests reproduce the previously reported normal/passthrough continuation failure and now pass.
- Handler tests cover a successful strict two-turn flow and reject continuation after owner disable, reconfiguration or deletion without contacting a replacement upstream.
- State-store tests cover persistence across instances, group isolation, scheduler-binding invalidation, cache misses and cache failures.
- Existing strict transport, header filtering, usage, scheduler and tenant-ownership tests remain in place.
- Full upstream CI: unit and integration suites, frontend checks, Go lint, security scans, shell checks and CLA.

Current reviewed head: `33f746fbda14311fb806f57cbba654059b6e7af0`. Runtime deployment is separate from this PR's source and CI verification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ Sub2API is an AI API gateway platform designed to distribute and manage API quot
 - **Built-in Payment System** - Supports EasyPay, Alipay, WeChat Pay, and Stripe for user self-service top-up, no separate payment service needed ([Configuration Guide](docs/PAYMENT.md))
 - **Admin Dashboard** - Web interface for monitoring and management
 - **Composite Groups** - Admin routing layer that resolves requested models to concrete providers for multi-provider groups ([Operator Guide](docs/COMPOSITE_GROUPS.md))
+- **Strict Responses Upstreams** - Byte-faithful request and SSE forwarding for upstreams that natively implement OpenAI Responses ([Deployment Guide](docs/openai-strict-responses-upstream.md))
 - **External System Integration** - Embed external systems (e.g. ticketing) via iframe to extend the admin dashboard
 
 ## Ecosystem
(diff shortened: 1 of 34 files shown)


heathermhuang:
Update: the continuation compatibility finding is fixed in `33f746fbd`, and the full upstream CI suite passes on that head.

Normal and legacy passthrough continuations keep their existing behavior. Strict continuations retain the original strict account binding across scheduler invalidation and fail closed after owner disable, reconfiguration or deletion. Added regression tests cover these cases, a successful two-turn strict request, cross-instance persistence and cache failure handling.

@Wei-Shaw could you review the updated opt-in transport contract? I still consider this valuable for generic Responses-native upstreams that require byte-preserving forwarding. The PR description now reflects the final behavior and current verification; the earlier compatibility finding has been addressed.

PR #5913, opened 2026-08-20
fix(deepseek): route Responses account tests through OpenAI probe

xuhaihan:
## Summary

- Route CN-provider accounts using the Responses protocol through the OpenAI Responses account-test path while preserving Adaptive and Chat Completions routing.
- Use the protocol-compatible API key accessor for Responses probes.
- Append `/responses` to the configured DeepSeek base path: the official root resolves to `/responses`, while relays can use `/v1/responses` by configuring their base URL with `/v1`.
- Cover official and relay-path behavior with account-test and URL-building tests.

This fixes connection tests that could use the wrong protocol path even though normal gateway traffic was configured for native Responses.

## Validation

- `make test-unit`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/backend/internal/service/account_test_service.go
+++ b/backend/internal/service/account_test_service.go
@@ -289,6 +289,8 @@ func (s *AccountTestService) TestAccountConnection(c *gin.Context, accountID int
 		switch account.GetAPIProtocol() {
 		case APIProtocolAdaptive:
 			return s.testCNProviderAdaptiveConnection(c, account, modelID, prompt)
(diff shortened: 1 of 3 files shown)


xuhaihan:
Fixed the failing unit job in 01a008394: removed hostname-based `/v1` inference, preserved the existing configured-base-path contract, and updated neutral relay tests to configure `/v1` explicitly. `make test-unit` passes locally and all refreshed GitHub checks are green. recheck

PR #5914, opened 2026-08-20
fix(openai): gate prompt cache fields by capability

### LINKED ISSUE #5821: 178 版本频繁遇到prompt_cache_retention is not supported on this model
178 版本频繁遇到prompt_cache_retention is not supported on this model

Image

cwj2001:
## Background

Issue #5821 reports upstream rejections for `prompt_cache_retention`. The field cannot be globally stripped: legacy OpenAI models and GPT-5.6+ use different cache-control shapes, while OAuth/Codex, Grok, and unknown compatibility upstreams must not receive unsupported fields.

### Field compatibility policy

The gateway now evaluates the **final mapped upstream model** at every Responses egress boundary. `prompt_cache_key` is never changed.

| Upstream/account condition | `prompt_cache_retention` | `prompt_cache_options` |
| --- | --- | --- |
| OpenAI API Key + explicit `prompt_cache_retention` capability + supported legacy model | Preserve | Strip |
| OpenAI API Key + GPT-5.6 or newer | Strip | Preserve client value |
| OAuth/Codex, Grok, unknown model, unsupported account, or missing capability | Strip | Strip |

The legacy allowlist is intentionally narrow and accepts only documented model families plus strict `YYYY-MM-DD` snapshots. GPT-5.6+ is parsed numerically with structured suffix validation, so arbitrary aliases are not mistaken for supported models.

### Account configuration

- Adds `prompt_cache_retention` to `credentials.openai_capabilities`.
- This is an explicit, default-deny opt-in for OpenAI API Key accounts using a legacy supported model.
- No database migration is required; existing accounts keep their previous filtering behavior.
- Adds the default-off setting to account create, edit, and bulk-edit UI.

### Request-path coverage

Applies the shared final-model sanitizer to:

- standard and passthrough HTTP Responses requests;
- Grok Responses and WebSocket HTTP bridge requests;
- raw WebSocket ingress;
- WebSocket passthrough initial and follow-up `response.create` frames;
- nested `session.update` fields, including frames that inherit the current mapped session model.

The HTTP hot paths keep raw request bytes unless one of the cache-control fields is actually present.

## Tests

- `pnpm run typecheck`
- `pnpm exec vitest run src/components/account/__tests__/CreateAccountModal.spec.ts src/components/account/__tests__/EditAccountModal.spec.ts src/components/account/__tests__/BulkEditAccountModal.spec.ts`

The backend coverage includes legacy/token default-deny behavior, GPT-5.6+, OAuth/Grok/nil/unknown accounts, strict model suffixes, `prompt_cache_key` preservation, HTTP hot paths, WebSocket initial/follow-up frames, and session-model changes.

Fixes #5821

### DIFF
--- a/backend/internal/service/account.go
+++ b/backend/internal/service/account.go
@@ -103,7 +103,8 @@ const (
 	// （openai_responses_supported / openai_responses_mode），而非
 	// credentials["openai_capabilities"] 配置集。仅用于生图意图的 /v1/responses
 	// 调度，避免把请求调度到会在 forward 阶段被降级为 Chat Completions 的账号（#4417）。
-	OpenAIEndpointCapabilityResponses OpenAIEndpointCapability = "responses"
+	OpenAIEndpointCapabilityResponses            OpenAIEndpointCapability = "responses"
+	OpenAIEndpointCapabilityPromptCacheRetention OpenAIEndpointCapability = "prompt_cache_retention"
 )
 
 const openAIEndpointCapabilitiesCredentialKey = "openai_capabilities"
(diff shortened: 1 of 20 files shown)


cwj2001:

PR #5923, opened 2026-08-20
fix(openai): gate Responses optional fields by official host

baryon:
## Problem

Responses optional fields currently sit on both sides of an incompatible boundary:

- Official OpenAI Platform API requests may legitimately use `prompt_cache_retention`, `safety_identifier`, and `prompt_cache_options`.
- ChatGPT internal, Grok, DeepSeek, Kimi, Zhipu, and arbitrary OpenAI-compatible endpoints may reject one or more of them.

The original global strip ran before account/upstream routing, so official API-key requests silently lost supported cache and safety metadata. Replacing that with a check for `PlatformOpenAI + APIKey` is still too broad because a custom `base_url` account has exactly the same platform and credential type while targeting a third-party endpoint.

After removing the global strip, several default HTTP and WebSocket paths also had no proactive cleanup. Relying on a 400 retry is not equivalent: error wording varies, one retry only removes one named field, and the first failure can interrupt streaming behavior.

## Root cause

Account type does not identify the actual trust boundary. The missing discriminator is the final upstream host:

- empty `base_url` means the official default;
- `api.openai.com` is official;
- every other host is an OpenAI-compatible upstream and must be treated as potentially incompatible.

Field cleanup was also spread across passthrough, OAuth, Grok, and WebSocket branches, leaving default forwarding and mixed-shape paths inconsistent.

## What changed

- Centralize the three Responses optional fields and official-host predicate.
- Preserve them only for OpenAI API-key accounts whose `base_url` is empty or has host `api.openai.com`.
- Strip all three before the first upstream request for custom OpenAI-compatible addresses and non-OpenAI/internal accounts.
- Apply the same policy to standard HTTP forwarding, raw WebSocket ingress, WS v2 initial/follow-up `response.create` frames, WS-to-HTTP bridge requests, and Cursor Responses-shaped bodies.
- Keep OAuth/Codex, Grok, and `/alpha/search` unsupported-field lists synchronized, including `prompt_cache_options` and `safety_identifier` where previously omitted.
- Retain bounded retry for official model-specific 400 responses, deleting only the explicitly rejected top-level optional field and preserving the others.

## Related open work

- #5848 protects ChatGPT OAuth HTTP egress and expands rejected-field retry. This PR additionally covers API-key/custom-host classification, compatible providers, Cursor, alpha search, and WebSocket boundaries.
- #5914 introduces model/capability-based prompt-cache handling. This PR addresses a separate endpoint-identity boundary and includes all three optional fields. In particular, #5914 currently keeps `prompt_cache_options` for a third-party `PlatformOpenAI + APIKey` account, while this change intentionally strips optional fields for any non-`api.openai.com` host before the first request.

## Tests

- `go test ./...`
- `go test -tags=unit ./internal/service`
- `go build ./cmd/server`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`

Regression coverage includes official/default and explicit official hosts, custom and lookalike hosts, Grok/Kimi/Zhipu/DeepSeek, OAuth transforms, Cursor mixed-shape requests, alpha search, normal and passthrough WebSockets, the WS HTTP bridge, and bounded single-field retry.

### DIFF
--- a/backend/internal/service/openai_alpha_search.go
+++ b/backend/internal/service/openai_alpha_search.go
@@ -443,13 +443,12 @@ func openAIAlphaSearchInboundHeader(c *gin.Context, key string) string {
 	return strings.TrimSpace(c.GetHeader(key))
 }
 
-var openAIAlphaSearchUnsupportedBodyFields = [...]string{
+var openAIAlphaSearchUnsupportedBodyFields = append([]string{
 	// Codex alpha/search 是 SearchRequest 独立协议，不是 /responses 子请求。
 	// 新版 Codex/第三方代理可能把 Responses 公共字段误带到搜索请求里；ChatGPT
 	// alpha/search 会对这些字段返回 Unknown parameter（例如 prompt_cache_key）。
 	"prompt_cache_key",
-	"prompt_cache_retention",
-}
+}, openAIResponsesOptionalFields...)
 
 func sanitizeOpenAIAlphaSearchBody(body []byte) ([]byte, error) {
 	if len(body) == 0 {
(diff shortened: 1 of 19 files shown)


baryon:

PR #5926, opened 2026-08-20
feat(gateway): generate complete routed Codex model catalogs

baryon:
## Why

Codex consumes `model_catalog_json` as a capability and routing contract, not merely as a list of model slugs. A slug-only routed catalog hides supported reasoning levels, can reject valid image inputs, can advertise unsupported image inputs, and can expose automatic or media-only models in `/model`.

Account `model_mapping` is the authority for model availability, but it is not a reliable capability database. New compatible-upstream models can appear before Sub2API has any local name-based metadata, leaving reasoning levels, input modalities, and context limits missing or wrong. Third-party Responses providers also commonly omit `/models`, so a configured model must remain usable without making catalog generation depend on that endpoint.

Composite catalogs aggregate exact account aliases. Without matching request-side ownership resolution, those aliases appear selectable in Codex but remain unroutable. Catalog generation, account capability snapshots, and alias routing therefore need to land together.

API key users also need a complete download-and-configure flow. The generated file must describe the effective route without embedding the API key.

## What changed

- Treat explicit account `model_mapping` entries as authoritative for OpenAI mixed groups and generate their Codex catalog locally before account selection, avoiding any runtime dependency on a third-party `/models` endpoint.
- Keep live upstream discovery and account failover only when the group has no explicit mappings.
- Extend the existing account-page **Sync upstream models** action to return normalized capability metadata in addition to model IDs.
- Prefer complete capability fields returned directly by the account upstream; when fields are missing, match the account API Base URL to the Models.dev provider registry and fill only missing fields.
- Persist complete capability snapshots in account `extra`; incomplete enrichment returns the structured `upstream_model_metadata_incomplete` warning and never replaces an existing snapshot.
- If `/models` returns 404 or 405 and the account already has concrete model mappings, use the configured upstream model IDs for capability enrichment. Do not use this fallback for 401, 429, 5xx, network failures, empty mappings, or wildcard targets.
- Never copy account authentication or header overrides to Models.dev requests. The account proxy and TLS profile may still be reused.
- Make generated Composite and routed catalogs consume saved account capability snapshots. Catalog membership still uses currently schedulable accounts; advertised capabilities for those models intersect all active group members that map the alias, including temporarily unschedulable ones, using the shared reasoning levels, input modalities, and smallest known context window.
- Keep account configuration authoritative: capability metadata can describe configured models, but it cannot add models to a group or make an unconfigured account eligible.
- Advertise a single `none` reasoning choice for visible non-reasoning models so current Codex clients can select them directly.
- Remove that catalog-only `none` placeholder before forwarding to third-party compatible Responses upstreams, including the HTTP WebSocket bridge, while preserving official OpenAI API key and OAuth request semantics.
- Convert official OpenAI API key `/v1/models` results into complete Codex `ModelInfo` entries and exclude image-generation, embedding, realtime, audio, TTS, and moderation models.
- Generate complete route-aware catalogs for Composite, Grok, DeepSeek, Anthropic, and other routed groups.
- Advertise provider-specific reasoning levels, including GPT-5.6 variants, Grok 4.6 `xhigh`, DeepSeek, and Claude effort levels.
- Derive image input support from the effective Responses route and account capability snapshot; unknown or conflicting capabilities fail closed.
- Filter GPT Image, Gemini image-generation, and Grok Imagine models from the Codex agent picker without changing ordinary `/v1/models` or media APIs.
- Hide `codex-auto-*` unless an enabled custom model list selects the exact slug.
- Compute the client ETag from the final filtered catalog so conditional requests match the body Codex receives.
- Add an authenticated **Use Key** action that downloads `codex-models.json` and emits matching Unix and Windows `model_catalog_json` configuration without including the API key in the file.
- Resolve unique exact Composite account-model ownership before built-in model detection, while keeping explicit routes authoritative and failing closed on cross-platform alias ambiguity.
- Preserve the public alias until the selected account applies its own model mapping, so every exact alias exposed by the catalog is also request-routable.

## How to use

1. Create or edit the provider account and configure its model whitelist or exact `model_mapping`. These settings decide which public models the account can serve.
2. In the account model section, click **Sync upstream models**. A complete upstream catalog is saved directly; an ID-only catalog is enriched from the Models.dev provider whose API Base URL matches the account.
3. During account creation, run the preview sync before submitting. After the account is created, the UI automatically invokes the formal account-ID sync so the capability snapshot can be persisted.
4. If the UI reports that model IDs synced but capability metadata is incomplete, the model list was updated but no incomplete snapshot replaced the previous one. The upstream must expose complete metadata or have a matching provider registry entry; Sub2API does not guess capabilities from another provider with the same model slug.
5. Add the account to the intended OpenAI or Composite group and configure the group models/routes as needed.
6. In **API Keys -> Use Key**, open the **Codex CLI** tab and click **Fetch model catalog**.
7. Download `codex-models.json` and place it at the path shown by the generated config (`~/.codex/codex-models.json` on macOS/Linux or `%USERPROFILE%\.codex\codex-models.json` on Windows).
8. Apply the generated environment-variable command and merge the generated user-level `config.toml` into `~/.codex/config.toml` or `%USERPROFILE%\.codex\config.toml`.
9. Restart Codex. Use `/model` in Codex CLI, or the model picker in the ChatGPT desktop app / Codex IDE extension, to switch among the routed entries.
10. Sync the account again and replace the downloaded catalog after changing account models, Base URLs, group routes, or upstream capabilities.

The generated config chooses the platform-preferred model when that slug exists in the downloaded catalog; otherwise it chooses the first valid entry, so `model` and `review_model` never point at a model omitted by the same catalog.

## Capability boundaries

- This config flow targets local Codex surfaces that read `config.toml`: Codex CLI, the ChatGPT desktop app's local Codex experience, and the Codex IDE extension. It does not change Codex Cloud or ChatGPT web model availability.
- OpenAI mixed groups use schedulable account `model_mapping` keys as the complete catalog whenever any explicit mapping exists. Accounts without mappings do not contribute live upstream models.
- Catalog membership still uses currently schedulable accounts. Advertised reasoning levels, input modalities, and context windows for those models intersect all non-deleted, `status=active` group accounts that map the alias, including members that are only temporarily unschedulable. Rate-limits, pauses, and health failures must not widen the client contract; deleting an account may.
- Routed alias failover is a clean contract only when clients send standard `tools` and omit `code_mode_only`. Private Codex tool shapes such as `use_responses_lite` and `code_mode_only` are out of scope for this PR.
- A 404/405 `/models` fallback reuses only concrete configured upstream IDs. It does not manufacture capability metadata and does not hide authentication, rate-limit, server, or network failures.
- Models.dev lookup is provider-specific by normalized API Base URL. Same-slug metadata from a different provider is intentionally ignored because supported reasoning levels and limits can differ.
- Anthropic, Gemini, and Antigravity remain distinct catalogs: Anthropic defaults contain Claude models only; Gemini defaults contain Gemini models only; Antigravity intentionally exposes its own Claude and Gemini compatibility surface.
- Composite catalogs aggregate only schedulable provider platforms. Exact aliases keep their public slug while inheriting metadata from a unique upstream target. Conflicting targets fail closed.
- Exact aliases discovered from account mappings are scheduled only to accounts that declare that alias; empty mappings cannot capture the request. Explicit Composite routes remain authoritative.
- Dedicated image/video generation, embedding, realtime, audio, TTS, moderation, and automatic models are excluded from the Codex agent picker as applicable.
- Image input is advertised only when the effective Responses route or saved account capability confirms it. Unknown aliases and ambiguous routes remain text-only.
- `none` is a Codex catalog selection value for non-reasoning models. It is omitted for third-party compatible Responses requests, not rewritten into another effort level.

## Validation

- `go test -tags=unit ./...` (54 packages)
- `go test ./internal/service ./internal/handler/admin -count=1`
- `go vet ./internal/service ./internal/handler/admin`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.9.0 run --timeout=30m` (`0 issues`)
- `make build` (server version `0.1.179`)
- `pnpm typecheck`
- `pnpm lint:check`
- `pnpm build`
- `pnpm test:run` (240 files, 1705 tests)

Supersedes #5927 and #5940.

OpenAI documents `model_catalog_json` as a startup-loaded catalog path and states that the ChatGPT desktop app, Codex CLI, and IDE extension share the same local `config.toml` layers:

- https://learn.chatgpt.com/docs/config-file/config-reference#configtoml
- https://learn.chatgpt.com/docs/models#configure-your-default-local-model

## Follow-up review update (2026-08-26)

The latest review pass added the following hardening:

- Expand exact group model selections covered by wildcard account mappings without publishing wildcard slugs or falling back to a live account `/models` request.
- Include the current concrete `model_mapping` in create-account preview sync and run the formal account-ID capability sync after creation whenever concrete mappings are present.
- Preserve the `ultra` reasoning level during upstream metadata normalization.
- Offer the Codex catalog/config flow for every routed group supported by the backend, including Anthropic, Gemini, Antigravity, Kimi, and Zhipu.
- Always compute a strong client ETag from the final response body, even when the upstream manifest has no ETag and local completion does not otherwise change the body.
- Rebuild API-key client manifests from the cached upstream source body so capability precedence is strictly: **upstream explicit fields > current `upstream_model_metadata` snapshot > local descriptor fallback**.
- Replace the saved model/capability snapshot on each successful formal sync, so upstream additions appear and removed models disappear. A stale metadata entry cannot reintroduce a model absent from the current upstream model list.
- Bound the processed manifest and retained upstream source body to the existing combined 1 MiB cache-entry budget.

### Dynamic sync boundaries

- With a working `/models` endpoint, model additions and removals are reflected after the next catalog refresh/formal sync. ID-only model lists use local descriptors only until capability sync supplies provider-specific metadata.
- A 404/405 fallback can only use concrete configured mapping targets. It cannot discover unconfigured additions or prove that an upstream model was deleted.
- Known reasoning levels are `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`, and `ultra`. A future unknown level remains fail-closed until the Codex client contract explicitly supports it.

### Fresh validation

The results below supersede the earlier test counts in this description:

- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`
- `pnpm exec vitest run` — 246 files, 1767 tests
- `go vet ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.13.0 run --timeout=30m` — 0 issues
- `make build`
- `go test ./...`


baryon:
SCR-20260821-nyfh
SCR-20260821-nvov
SCR-20260821-nxij

zanyatta:
Tested this branch against a **real third-party Responses upstream** (Volcengine Ark "Agent Plan", `https://ark.cn-beijing.volces.com/api/plan/v3`, serving `glm-5.3`) rather than a mock. Built the branch into an image and ran it against sub2api v0.1.179's database. Sharing results since this upstream is probably hard for you to obtain.

Setup: one group with `platform=openai`, containing a Codex OAuth account (`priority=1`) and an API-key account pointing at Ark (`priority=50`, `model_mapping={"glm-5.3":"glm-5.3"}`).

## What works

- `MergeGroupConfiguredCodexModels` does exactly what we needed: `glm-5.3` shows up in the Codex model picker **alongside** the upstream OAuth models, with correct ETag recomputation.
- Request routing stays orthogonal: `gpt-5.6-sol` → OpenAI, `glm-5.3` → Ark, tool calls intact on both.
- `use_responses_lite: false` on the generated entry is the right call. With `true`, tools get placed in an `additional_tools` developer item, which non-OpenAI upstreams silently drop — the model then narrates that it ran a command without emitting any tool call. Failing that way is worse than erroring, so this default matters.
- Failing closed to `input_modalities: ["text"]` matches reality here. Ark rejects images for this model with `400 InvalidParameter: Model only support text input`.
- `apply_patch_tool_type: null` is also correct — Ark accepts the freeform custom tool without erroring, but the model never invokes it.

## Issue 1 (blocking): a 404 from an API-key upstream's `/models` fails the whole manifest request

Many third-party Responses-compatible upstreams implement `/responses` but **not** `/models`. Ark is one: `GET {base_url}/models` returns `404`.

`FetchCodexModelsManifest` builds the URL as `base_url + /models` for API-key accounts, and the retry predicate is:

```go
retryable: (resp.StatusCode == http.StatusUnauthorized && !request.useAPIKeyUpstream) ||
    resp.StatusCode == http.StatusTooManyRequests ||
    (resp.StatusCode >= http.StatusInternalServerError && resp.StatusCode < 600),
```

`404` is not retryable, so `CodexModels` returns `502` to the client instead of excluding that account and trying the next one. Observed in production logs:

```
GET /v1/models  status=502  account=7(api-key/Ark)  latency=5148ms
{"error":{"message":"codex models manifest upstream error 404: 404 Not Found","type":"upstream_error"}}
```

The OAuth account in the same group was healthy, schedulable, higher priority, and would have served the manifest fine. Because the model-picker refresh runs every ~3 minutes in Codex, any scheduling round that happens to land on the API-key account breaks the model list for that client.

Suggestion: treat `404` (and probably `405`) from an API-key upstream's `/models` as "this account cannot serve a manifest" and add it to `failedAccountIDs`, rather than as a terminal error. An upstream that lacks the endpoint is a static property of that account, not a request failure.

## Issue 2: generated metadata for routed models is inaccurate in ways that degrade silently

The entry generated for `glm-5.3`:

```
context_window             = 272000
max_context_window         = 272000
supported_reasoning_levels = []
default_reasoning_level    = null
shell_type                 = "unified_exec"
```

Two problems:

**`context_window` defaults to the OpenAI value.** Ark accepted a request with **680,022 input tokens** (measured, single request), and the vendor documents 1M for this model. Codex derives its auto-compaction threshold as `context_window * 90%`, so a 272000 default makes it compact at ~245k instead of ~944k. That wastes roughly three quarters of the usable context and adds an extra summarization round-trip each time — with no error surfaced anywhere. Users just perceive the model as forgetful.

**`supported_reasoning_levels` is empty.** Ark documents exactly three levels for this route (`low` / `medium` / `high`) and rejects `none` / `minimal` with `400 InvalidParameter`. An empty list gives the client nothing to offer in the picker.

I understand why: the catalog builder only learns the model **exists** from the `model_mapping` key, and there is no capability table for this provider (unlike `pkg/claude/effort_catalog.go` / `pkg/xai/models.go`) because the account's platform is `openai`, not one of the recognized CN providers.

Suggestion: allow per-account overrides for the generated descriptor — at minimum `context_window` and `supported_reasoning_levels`. Operators know their upstream's real limits; today there is no way to express that, and both defaults fail in the direction that degrades quality silently rather than loudly.

Alternatively, when the upstream returns a body that already carries a top-level `models` field, that body is passed through verbatim today — documenting that as the supported escape hatch would also work, since an operator can then front their upstream with a static `/models` endpoint. That path incidentally fixes Issue 1 too.

## Supporting measurements

| Check | Result |
|---|---|
| `GET {ark}/api/plan/v3/models` | `404` |
| Single request with 680,022 input tokens | accepted |
| `input_image` in payload | `400 InvalidParameter: Model only support text input` |
| `reasoning.effort` = `none` / `minimal` | `400 InvalidParameter` |
| `reasoning.effort` = `low` / `medium` / `high` | accepted (documented range) |
| `reasoning.effort` = `xhigh` / `max` / arbitrary string | accepted but undocumented — no guarantee |
| Tools in `additional_tools` item (lite shape) | accepted, **tools silently dropped**, no tool call emitted |
| Tools in standard `tools` field | accepted, tool call emitted correctly |

baryon:
@zanyatta Thank you for testing this against a real Ark account and for documenting the failure mode and capability measurements in detail. I pushed `4eb0f08ad` to this PR.

For Issue 1, the fix uses an account-configuration-first rule instead of retrying `404`/`405` from `/models`:

- If any schedulable account in an OpenAI group has explicit `model_mapping` entries, those public keys become the complete authoritative Codex catalog.
- The manifest is generated locally before account selection, so neither the OAuth account nor the API-key account is asked for `/models` in that case.
- If the group has no explicit mappings, the existing upstream manifest discovery, cache, and account failover behavior remains unchanged.
- Composite groups keep their existing account-first catalog behavior.

The regression test reproduces the relevant shape: an OpenAI group containing an OAuth account plus an Ark-compatible API-key account with `model_mapping={"glm-5.3":"glm-5.3"}`, while the API-key `/models` response is `404`. It now asserts `200`, a catalog containing `glm-5.3`, and zero upstream manifest calls.

One intentional behavior difference from the branch you tested: when `glm-5.3` is the only configured public mapping, the picker now contains only `glm-5.3`; OAuth models are no longer merged in implicitly. To expose GPT models alongside it, those public model mappings need to be configured as well. This keeps the Codex catalog consistent with the administrator-defined account model list and the ordinary model-list priority rules.

Issue 2 is intentionally not addressed by this change. Unknown model metadata still uses the existing conservative fallback. Model-specific context windows and reasoning levels should be owned by the provider/account configuration layer; GLM support can be improved separately without teaching the generic catalog builder to guess provider capabilities.

Could you retest the updated branch against the same Ark setup when convenient? The key checks are that `/v1/models?client_version=...` returns `200`, Ark `/models` is never called, and request routing to `glm-5.3` remains unchanged.

baryon:
@zanyatta I followed up on the capability-metadata part in `e39fce270` while keeping the account-configuration-first behavior from the previous fix.

The updated flow is:

- Account `model_mapping` remains the authority for availability and routing. Capability metadata cannot add models or make an unconfigured account eligible.
- The existing account-page **Sync upstream models** action now parses reasoning levels, input modalities, and context/output limits from a complete upstream model response.
- If the upstream returns only IDs, Sub2API matches the account API Base URL to the Models.dev provider registry and fills missing fields, then persists a complete account capability snapshot.
- Generated OpenAI/Composite Codex catalogs read that snapshot. When multiple schedulable accounts can serve the same public model, the catalog exposes only their shared capabilities and uses the smallest known context window.
- If `/models` returns 404 or 405, but the account already has concrete `model_mapping` targets, those configured upstream IDs are used to continue capability enrichment. This fallback does not apply to 401, 429, 5xx, network errors, empty mappings, or wildcard targets.
- If enrichment is still incomplete, the API returns `upstream_model_metadata_incomplete`, the frontend reports partial success, and an existing snapshot is preserved rather than overwritten.

One important boundary for the Ark setup you tested: the reported Ark Base URL currently has no matching provider metadata entry, and Ark itself returns 404 from `/models`. This PR therefore does **not** guess `glm-5.3` capability values from another provider with the same slug, because those providers expose different effort sets. The account sync will retain the configured model ID and return the incomplete-metadata warning, not incorrectly advertise `low/medium/high`. A provider/account-specific metadata source for that Ark route remains separate work, or a complete static `/models` response in front of Ark can supply the fields directly.

The original manifest failure remains fixed independently: with explicit mappings, `/v1/models?client_version=...` is generated locally and does not call Ark `/models` at request time.

The PR description now includes the account sync and Codex setup steps, the 404/405 behavior, and these capability boundaries. Thanks again for the production measurements; they exposed the distinction between model availability and provider-specific capability metadata clearly.

zanyatta:
@baryon Thanks for both follow-ups — `4eb0f08` fixed the manifest `502` in our setup exactly as described, and the Models.dev enrichment in `e39fce2` is a sensible boundary to draw.

We retested `e39fce2` against the same Ark account, then went one step further: we want a **single public model name backed by multiple upstreams** (OpenAI OAuth as primary, Ark as fallback), with the client never seeing the upstream names. That is the "account-configuration-first" model this PR introduces, taken to its natural conclusion — and it surfaced four things. Three are small code fixes (patch attached below), one is a UX gap that I think matters for mergeability.

## 1. OpenAI groups never read the capability snapshot

The commit message says *"Generated OpenAI/Composite Codex catalogs read that snapshot"*, but only the Composite/Grok path does. `BuildGroupConfiguredCodexModelsManifest` (the OpenAI-group path) starts from `BuildCodexModelsManifest(nil)` and feeds `mergeConfiguredCodexModelsManifest`, which emits `newConfiguredCodexModelDescriptor(modelID)` — the bare conservative descriptor. `groupCodexModelMetadata` (the function that reads `upstream_model_metadata` and intersects) is only reached via `BuildCodexModelsManifestForGroup`.

Reproduction: OpenAI group, API-key account with `model_mapping={"my-coder":"glm-5.3"}` and a complete snapshot under `extra.upstream_model_metadata.models["glm-5.3"]` (three reasoning levels, text-only, 1M context). `GET /v1/models?client_version=…` still returns `supported_reasoning_levels=[{"effort":"none"}]`, `context_window=272000`.

Fix we applied: a small helper on `OpenAIGatewayService` that mirrors the Composite path (`codexCatalogMetadataModels` → `groupCodexModelSupportsImageInput` → `groupCodexModelMetadata` → `buildCodexModelsManifest`) with `compositeRoutes=nil`, and `BuildGroupConfiguredCodexModelsManifest` calling it instead of `BuildCodexModelsManifest(nil)`. ~20 lines. After this, the same request returns `['low','medium','high']`, `['text']`, and the snapshot's context window.

## 2. Same alias → different upstream targets is treated as a misconfiguration

`groupCodexModelMetadata` short-circuits on `codexExplicitModelTargetsConflict` and returns `reasoningConflict=true, inputModalitiesConflict=true`, which wipes the reasoning levels and forces text-only — even when every account has a complete snapshot.

For a primary/fallback setup this is the *intended* shape: `my-coder → gpt-5.6-sol` on the OAuth account, `my-coder → glm-5.3` on the Ark account. The intersection logic in `intersectUpstreamModelMetadata` already does exactly the right thing for this case (common reasoning subset, minimum context window, modality intersection, and it flags a real conflict when one target has `reasoning=true` and another `reasoning=false`). The name-based short-circuit just prevents it from running.

Fix we applied: drop the `codexExplicitModelTargetsConflict` early return in `groupCodexModelMetadata` and let the candidates flow into the intersection. The existing test `TestBuildCodexModelsManifestForGroupUsesSafeFallbackForConflictingAliasTargets` then yields `["none"]` instead of `[]` for the no-snapshot case, which matches the single-`none`-choice convention you introduced in `e39fce2`; we updated the assertion accordingly.

Measured result with both snapshots present: `my-coder` advertises `low/medium/high`, `text`, `272000` — the safe intersection of a 6-level/text+image/272k OpenAI model and a 3-level/text/1M Ark model. Routing `my-coder` lands on OpenAI when the OAuth account is schedulable and on Ark when it is not, with the response `model` field reporting `my-coder` in both cases, and tool calls working on both.

## 3. `display_name` leaks the upstream name in the fallback state

`intersectUpstreamModelMetadata` takes `DisplayName` from the first candidate. Snapshots are keyed by the *upstream* model id, so when only the Ark account is schedulable the picker shows `glm-5.3` as the display name of `my-coder`. When both are schedulable it shows whatever the first account's snapshot says.

For an alias the public name should win. Suggest: when the model id is an alias (i.e. `explicitClaims` and `mappedModel != modelID`), set `DisplayName` to the alias itself unless an alias-level display name is configured somewhere — which leads to point 4.

## 4. The capability snapshot has no UI — operators cannot see or edit the catalog they are shipping

This is the one I'd flag for mergeability rather than correctness.

`extra.upstream_model_metadata` is now the source of truth for what Codex clients see (reasoning levels, modalities, context window, display name). But the only frontend reference to it is the `upstream_model_metadata_incomplete` warning toast. There is no place in the account modal that shows the snapshot, no way to edit it, and no way to set alias-level fields like display name or description. An operator configuring `model_mapping` in the UI has no indication that a second, invisible structure decides the model's advertised capabilities.

In practice this bit us immediately: we wrote the snapshot via the admin API, got the display-name leak in §3, and could only diagnose it by querying the database. A reviewer will reasonably ask "how does an admin know what this catalog contains?"

Suggest, minimally: show the per-model snapshot (read-only is fine) next to the model whitelist in `EditAccountModal`, with the "Sync upstream models" result written into that view; and allow editing `display_name` / `description` / `context_window` / `supported_reasoning_levels` / `input_modalities` per model, since for providers not in Models.dev (Ark is one) there is no other source.

## Patch

The three code changes (§1–§3 minus the display-name part) plus test updates are ~60 lines. Happy to open a PR against your branch if that's easier than pasting a diff here. Files touched:

- `backend/internal/service/openai_codex_models_service.go` (helper + call site)
- `backend/internal/service/openai_codex_model_metadata.go` (remove short-circuit)
- `backend/internal/handler/openai_codex_models_handler.go` (404/405 from API-key `/models` → exclude account, try next; unchanged semantics for 400/401/403)
- corresponding `_test.go` files

`go test ./internal/handler/ ./internal/service/` passes in full.

baryon:
@zanyatta Thanks again for the detailed retest and for narrowing this down to the primary/fallback alias case. I pushed `fc589bce1` with the backend fixes from your follow-up.

- OpenAI groups with explicit mappings now build their local Codex catalog through the same account capability snapshot path used by Composite groups.
- A public alias mapped to different upstream targets on the same platform now intersects complete snapshots instead of treating the different target IDs as a conflict. If a required snapshot is missing, the catalog remains conservative. Cross-platform ambiguity still fails closed.
- Routed aliases retain the public alias as `display_name` and use a generic routed description, so account ordering or fallback state no longer exposes the upstream model identity.
- API-key `/models` responses with 404/405 now fail over to another selected account for manifest discovery. Responses with 400/401/403, and OAuth 404/405 responses, remain non-retryable.
- The configured catalog merge no longer re-adds media-only models filtered out by the catalog builder.

Regression coverage now includes both account orders, complete-snapshot intersection, missing-snapshot conservative behavior, cross-platform ambiguity, alias identity preservation, and API-key 404/405 failover. `go test ./internal/service ./internal/handler -count=1` passes.

I did not include the snapshot viewer/editor mentioned in point 4. I agree that it is a real operator UX gap, especially for providers without a Models.dev entry, but I kept it as a separate frontend follow-up rather than expanding this PR further.

Could you retest `fc589bce1` against the same OpenAI-primary / Ark-fallback setup when convenient? The main checks are the advertised capability intersection, the public alias remaining stable in fallback-only state, and `/models` 404/405 no longer breaking catalog refresh.

zanyatta:
@baryon Retested `fc589bce1` on a clean instance (fresh DB, fresh containers) against the same Ark setup. **All four points check out.** Details below, plus one design side-effect I think is worth a look.

## Retest results

One `platform=openai` group, two API-key accounts both pointing at Volcengine Ark, the same public alias mapped to different upstream models, with complete capability snapshots written by hand (Ark has no Models.dev entry, so there is no other way to populate them):

| Account | Upstream | Snapshot: levels / modalities / context |
|---|---|---|
| A | `kimi-k3` | low,medium,high / text+image / 1,000,000 |
| B | `glm-5.3` | low,medium,high / text / 272,000 |

`GET /v1/models?client_version=0.150.0` across four scheduling states:

| State | display_name | levels | modalities | context |
|---|---|---|---|---|
| A+B | **the alias** | low,medium,high | **text** | **272000** |
| B only | **the alias** | low,medium,high | text | 272000 |
| A only | **the alias** | low,medium,high | text,image | 1000000 |
| A+B again | **the alias** | low,medium,high | text | 272000 |

- **§1 fixed.** OpenAI groups now read the snapshot. Before this commit the same request returned `[{effort:"none"}]` and `context_window=272000`.
- **§2 fixed.** Different upstream targets behind one alias intersect instead of being treated as a conflict: `text+image ∩ text = text`, `min(1M, 272k) = 272k`.
- **§3 fixed.** `display_name` stays the public alias in all four states; `description` is the generic routed text. No upstream identity leaks in fallback-only state.
- **§4 (the original Issue 1) fixed.** With the Ark account (404 on `/models`, higher priority) and a second API-key upstream that does serve `/models`, the manifest request returns **200** on every attempt, using the second account's manifest. This used to be a hard 502.

Beyond the manifest, a real `POST /v1/responses` on the alias completes normally: `status: completed`, the response `model` field reports the alias, and `usage.output_tokens_details.reasoning_tokens = 29` confirms the reasoning level is actually applied upstream.

## One side-effect: the advertised contract widens when an account is temporarily out of rotation

Row three above: while B is unschedulable, the alias advertises `text+image` and `1,000,000`. When B comes back, it narrows again to `text` / `272,000`.

Codex caches the manifest (ETag + 300s TTL, ~3-minute refresh). So there is a window where a client picked up the wider manifest, B returns to rotation, and the client sends an image — or plans for a 1M context — against a request that now routes to B. Ark answers `400 InvalidParameter: Model only support text input`.

This is the hard kind of failure to diagnose, because neither the catalog nor the routing is wrong on its own; they are just inconsistent in time. And temporary unschedulability (rate-limit cooldown, manual pause, failed health check) is normal operation, not an exceptional state.

Suggestion: intersect over **all non-deleted accounts in the group that map the alias**, not only the currently schedulable ones. Narrowing when an account is deleted is correct — the contract genuinely changed. Widening when one is briefly out of rotation is not. A weaker alternative is "allow narrowing, never widen automatically", but that needs extra persisted state and is less clean.

## A boundary beyond the catalog: the client's tool shape

This is not a defect in the PR — it sits outside what a relay can fix — but it changes what
"one public alias, multiple upstreams" can actually deliver, so it seems worth recording here.

The catalog side now works. The request side has a second gate: **Codex ships two
OpenAI-private tool shapes**, and non-OpenAI upstreams do not understand either of them.

- `tool_mode: code_mode_only` — the model gets a single freeform `execute` tool plus a lark
  grammar and writes source code that calls other tools, instead of emitting standard function
  calls (`codex-rs/protocol/src/openai_models.rs`, `core/src/tools/code_mode/`). Note that
  `CodeMode` falls back to `Direct` when unsupported, but **`CodeModeOnly` does not**.
- `use_responses_lite: true` — tools move into an `additional_tools` developer item instead of
  the top-level `tools` field.

Measured on a real Codex client against one group holding an OpenAI OAuth account plus an Ark
account, same task each time:

| catalog entry | via OpenAI | fallback to Ark |
|---|---|---|
| `lite=true` + `tool_mode=code_mode_only` | ✅ | ❌ **tools silently dropped; the model narrates that it ran the command** |
| `lite=false` + `tool_mode=code_mode_only` | ✅ | ❌ model asks the user to run the commands manually |
| `lite=true` + no `tool_mode` | ✅ | ❌ |
| `lite=false` + no `tool_mode` | ✅ | ✅ tools execute for real |

The practical consequence: **whether fallback works depends on which model the user picked.**
In the catalogs we see, the `gpt-5.6-*` entries carry `code_mode_only` + `lite` and the
`gpt-5.5` / `gpt-5.4*` entries carry neither — so the older entries fall back cleanly while the
newest ones cannot, and in our traffic the newest one accounts for ~98% of calls.

The first row is the dangerous one: it fails silently. The user sees the model claim it ran a
command, nothing happened, and the request is still billed. That is worse than an error.

I mention it because operators enabling this PR's alias feature for failover will likely hit it
without an obvious diagnosis. A note in the docs — "a routed alias only fails over cleanly if
the client's catalog entry uses standard tool shapes" — would probably save people a lot of time.
Detecting it server-side looks impractical: the relay would have to inspect the tool payload
shape and know the target's dialect.

## Where we ended up

For transparency, since you invested three rounds on our feedback: we are moving away from the
one-alias-many-upstreams model, for the reason above rather than anything wrong with this PR.
Presenting a single model identity across heterogeneous vendors means something has to translate
the client's private tool shapes, and that translation has to keep chasing a client we neither
control nor get advance notice from. We are instead exposing distinct sources and switching
between them at the product layer, accepting a visible switch instead of an invisible one.

None of that diminishes the PR: the catalog generation, the snapshot intersection, and the
404/405 failover are all correct and useful on their own, and the retest above stands.

## Minor, unrelated to this PR

`credentials.openai_capabilities` silently accepts `["responses"]`, which is not a member of that capability set (`/v1/responses` eligibility comes from `extra.openai_responses_supported`). The account then shows `active` + `schedulable` while every request fails with `no available accounts` → 503. `admin_service_bulk_update` rejects the same value with `OPENAI_ENDPOINT_CAPABILITIES_INVALID`; the create path does not. Happy to open a separate issue if that is useful.

baryon:
@zanyatta Thanks for the retest and the catalog-widening catch.

Implemented in this PR:

- Catalog membership still uses currently schedulable accounts (`ListSchedulableByGroupID`).
- Advertised capabilities for those models now intersect all non-deleted, `status=active` group accounts that map the alias (`ListByGroup`), including rate-limited / paused / health-failed members.
- `ListByGroup` failures fall back to the schedulable set so a listing error does not fail the client request.
- An extra public model that exists only on an unschedulable account is not added to the catalog.

Out of this PR, as discussed:

1. Private Codex tool shapes (`use_responses_lite` / `code_mode_only`) — alias failover is a clean contract only with standard `tools` and without `code_mode_only`. Documented in the PR body.
2. `credentials.openai_capabilities: ["responses"]` create-path validation — separate issue.
3. Snapshot viewer/editor UI — later frontend work.

Wei-Shaw:
## Follow-up code audit

I re-checked the current PR head (`77b6c5bb0cda5d56690941d3b2095d7b493b1e43`) against the implementation and call paths. The following are confirmed correctness defects or implementation omissions, not feature requests. CI is green; I did not rerun the test suite.

### 1. High: the API-key manifest cache is mutated by group-specific policy

The API-key cache returns the shared `*CodexModelsManifest` pointer on fresh and stale hits ([openai_codex_models_service.go#L1539-L1563](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L1539-L1563)). The cache key is based on account/credential/upstream identity and does not include the group ([openai_codex_models_service.go#L2032-L2046](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L2032-L2046)).

After fetching it, the handler mutates that object in place for API-key metadata completion and group model filtering/ETag handling ([openai_codex_models_handler.go#L79-L99](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/handler/openai_codex_models_handler.go#L79-L99), [openai_codex_models_service.go#L154-L188](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L154-L188), [openai_codex_models_service.go#L1842-L1858](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L1842-L1858)).

This is reproducible when one API-key account is attached to multiple groups and those groups have different custom model lists. Group A can contaminate the cached body and ETag returned to group B. A matching conditional request can also set the shared object to `NotModified=true` with an empty body. Concurrent requests can mutate the same object without synchronization.

Please keep the cached upstream manifest immutable and deep-copy it before applying metadata completion or any group-specific merge, or use a separate cache for final group results. A regression test should cover the same account in two groups with different filters and concurrent requests.

### 2. High: DeepSeek groups without explicit mappings fall through to Claude defaults

For a non-Composite group, `codexModelIDsForGroup` uses `GetAvailableModels` and falls back to `defaultModelIDsForPlatform` when no mapping exists ([gateway_handler.go#L1179-L1213](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/handler/gateway_handler.go#L1179-L1213)). `defaultModelIDsForPlatform` has no `PlatformDeepseek` case; the default branch returns Claude model IDs ([gateway_handler.go#L1427-L1467](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/handler/gateway_handler.go#L1427-L1467)).

Therefore a DeepSeek group with an empty `model_mapping` can advertise Claude slugs in its Codex manifest. The new Use Key flow explicitly exposes a DeepSeek Codex tab and prefers `deepseek-v4-pro` ([UseKeyModal.vue#L463-L468](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/frontend/src/components/keys/UseKeyModal.vue#L463-L468), [UseKeyModal.vue#L1178-L1187](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/frontend/src/components/keys/UseKeyModal.vue#L1178-L1187)); when that slug is absent, the selector chooses the first returned slug ([UseKeyModal.vue#L648-L651](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/frontend/src/components/keys/UseKeyModal.vue#L648-L651)). The generated DeepSeek config can consequently point at a Claude model and fail at the DeepSeek upstream.

Please add an explicit DeepSeek default policy, or fail closed instead of returning another provider's model IDs.

### 3. High: generated OpenAI configs hardcode an unsupported reasoning effort

The OpenAI and OpenAI-WebSocket config generators select the first valid catalog model dynamically ([UseKeyModal.vue#L906-L916](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/frontend/src/components/keys/UseKeyModal.vue#L906-L916), [UseKeyModal.vue#L1217-L1227](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/frontend/src/components/keys/UseKeyModal.vue#L1217-L1227)), but always write `model_reasoning_effort = "xhigh"`.

The catalog builder's generic routed descriptor advertises only `none` for an unknown/custom model ([openai_codex_models_service.go#L334-L344](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L334-L344)). This occurs specifically in the mixed OpenAI/API-key scenario that this PR enables. The downloaded catalog and generated `config.toml` therefore disagree, and compatible upstreams may reject `xhigh` or behave inconsistently.

Please derive the default effort from the selected descriptor, or omit the setting for non-reasoning models; do not emit an effort that is absent from `supported_reasoning_levels`.

### 4. Medium: wildcard mapping keys can leak into generated Codex catalogs

`GetAvailableModels` adds every mapping key to the model list ([gateway_service.go#L1426-L1432](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/gateway_service.go#L1426-L1432)). The generic Codex filter removes media and `codex-auto-*` entries but does not remove wildcard IDs ([openai_codex_models_service.go#L45-L72](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L45-L72)). The OpenAI-specific catalog path already skips IDs containing `*` ([openai_codex_models_service.go#L222-L243](https://github.com/Wei-Shaw/sub2api/blob/77b6c5bb0cda5d56690941d3b2095d7b493b1e43/backend/internal/service/openai_codex_models_service.go#L222-L243)), so this is an inconsistent omission.

A slug such as `foo-*` is not a concrete Codex model and can be selected by the client even though Composite ownership resolution is exact-match based. Please apply the same wildcard exclusion to all generated Codex catalog paths.

I did not classify the snapshot viewer/editor gap, private Codex tool-shape limitations, or the Models.dev dependency as defects in this PR; those are documented boundaries or follow-up product work.

baryon:
@Wei-Shaw Thanks for the follow-up audit. Addressed in `195b21970` on this branch.

1. **API-key manifest cache** — `codexModelsManifestForClient` now returns a deep copy. The shared cache entry stays immutable; metadata completion, group filtering, and `NotModified`/empty-body handling only mutate the copy. Regression covers the same account in two groups with different custom lists, plus concurrent requests.

2. **DeepSeek Codex fallback** — `defaultModelIDsForPlatform` is unchanged, so Claude Code still sees Claude slugs on CN groups. Codex catalog generation uses `defaultCodexModelIDsForPlatform`, which returns `deepseek-v4-pro` and `deepseek-v4-flash` for DeepSeek instead of falling through to Claude.

3. **Generated OpenAI `config.toml` effort** — `model_reasoning_effort` is derived from the selected catalog descriptor. If the default is `none`, missing, or not in `supported_reasoning_levels`, the setting is omitted rather than hardcoded to `xhigh`.

4. **Wildcard mapping keys** — `FilterCodexModelIDsForGroup` now drops IDs containing `*`, matching the OpenAI configured-catalog path.

`go test ./internal/service ./internal/handler -count=1` passes. Frontend coverage is in `codexCatalogConfig.spec.ts` and `UseKeyModal.spec.ts`.

baryon:
Follow-up review fixes are pushed in `2abce65031`. The PR description now documents the dynamic model/capability precedence, add/delete synchronization behavior, fallback boundaries, cache memory bound, and fresh validation results (246 frontend files / 1767 tests plus `go test ./...`).

PR #5927, opened 2026-08-20
fix(openai): hide image generators from Codex model picker

baryon:
## Why

Custom OpenAI API key upstreams may return text models and dedicated image-generation models from the same `/v1/models` endpoint. The Codex conversion currently treats every returned ID as an agent model, so `gpt-image-*` entries appear in `/model` even though Codex cannot use them as text/agent models.

## What changed

- Filter `gpt-image-*` slugs while adjusting API-key Codex model manifests.
- Preserve malformed/unknown manifest entries and unknown fields.
- Keep OAuth ChatGPT manifests unchanged.
- Leave ordinary `/v1/models`, Images API, and Responses image tools unchanged.
- Reuse the adjusted body for the client ETag while retaining the upstream ETag for revalidation.

This PR is independent of #5926 and is based directly on the current `upstream/main`.

## Validation

- `go test ./...`
- `go vet ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.9.0 run --timeout=30m ./...` (`0 issues`)
- `pnpm run lint:check`
- `pnpm run typecheck`
- `pnpm run test:run` (237 files, 1666 tests)
- `make build`

### DIFF
--- a/backend/internal/service/openai_codex_models_service.go
+++ b/backend/internal/service/openai_codex_models_service.go
@@ -567,10 +567,10 @@ var apiKeyCodexModelsWithoutResponsesLite = map[string]struct{}{
 	"gpt-5.6-luna":  {},
 }
 
-// adjustAPIKeyCodexModelsManifest prevents Codex from selecting Responses
-// Lite for custom API key providers. Those clients do not install web.run in
-// Lite mode, so the affected model manifests must advertise the full Responses
-// path. Return the original body when no targeted true value is present.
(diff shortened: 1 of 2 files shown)


baryon:
Superseded by #5926. The complete routed catalog now includes this media-model filtering together with reasoning levels, route-aware image capability, Auto-model opt-in, final ETag handling, and the API-key download flow. Keeping the behavior in one PR avoids two overlapping implementations of the Codex manifest path.

PR #5940, opened 2026-08-20
fix(composite): route exact account model aliases

baryon:
## Why

Composite model catalogs aggregate exact account `model_mapping` keys, but request routing only considered persisted Composite routes or model-name detection. A custom or unprefixed alias could therefore appear in `/models` and Codex model catalogs while remaining unroutable.

## What changed

- keep explicit Composite routes authoritative
- resolve a unique exact account-model owner before built-in model detection
- allow duplicate claims on the same platform and fail closed on cross-platform ambiguity
- ignore wildcard and empty model mappings for ownership
- preserve the public alias until the selected account applies its own mapping
- reuse the existing short-lived model-list cache and invalidation path
- preserve detector routing for known models when ownership lookup fails

## Validation

- `go test ./... -count=1`
- `go vet ./...`
- `golangci-lint run ./...`
- `CGO_ENABLED=0 go build ./...`
- `pnpm typecheck`
- `pnpm lint:check`
- `pnpm build`
- `pnpm test:run` (237 files, 1666 tests)

### DIFF
--- a/backend/internal/service/composite_model_route.go
+++ b/backend/internal/service/composite_model_route.go
@@ -23,7 +23,18 @@ const (
 
 	CompositeRouteSourceExplicit = "route"
 	CompositeRouteSourceDetector = "detector"
+	CompositeRouteSourceAccount  = "account_model"
 )
+
+// CompositeModelOwnership identifies the concrete provider that exposes a
+// public model through an account-level exact mapping.
+type CompositeModelOwnership struct {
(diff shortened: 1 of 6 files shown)


baryon:
Consolidated into #5926 so generated Composite model aliases and their request-side routing land atomically. Commit abd6979dc is now part of the #5926 branch; closing this PR as superseded.

PR #6016, opened 2026-08-21
fix(securityaudit): log prompt_guard.config_loaded only on change

### LINKED ISSUE #5732: [Bug] Prompt Guard 每 5 秒重复输出 prompt_guard.config_loaded
## 问题描述

Prompt Guard 配置保持不变时，服务仍会每 5 秒输出一条 `prompt_guard.config_loaded`。我在 2026-08-17 的日志中统计到约 1.2 万条同类记录。

这不是环境变量配置错误，而是当前刷新与日志逻辑组合后的确定性行为。此前 #5310 出现过相同日志，但该 Issue 被提报者按“环境变量问题”自行关闭；当前 `upstream/main`（`e330c243a`）仍可从源码稳定复现。

## 根因

1. [`refreshLoop`](https://github.com/Wei-Shaw/sub2api/blob/e330c243a8f142f8963d784916da0093ab7084ee/backend/internal/securityaudit/prompt_config_store.go#L434-L447) 每 5 秒调用一次 `Reload(ctx)`。
2. [`Reload`](https://github.com/Wei-Shaw/sub2api/blob/e330c243a8f142f8963d784916da0093ab7084ee/backend/internal/securityaudit/prompt_config_store.go#L108-L144) 即使配置内容完全没有变化，也会无条件执行：

```go
LogInfo(EventConfigLoaded, map[string]any{
    "config_version": storage.ConfigVersion,
    "status":         "loaded",
})
```

因此单实例每天理论上最多会产生约 17,280 条无变化的加载日志。

## 影响

- 持续增加日志文件及日志采集成本；
- 掩盖真正的配置变化、降级和安全告警；
- 容易被误判为配置不断重载或服务异常。

## 期望行为

5 秒兜底刷新仍然更新运行态快照，但 `config_loaded` 只在以下情况记录：

- 首次成功加载；
- `config_version` 变化；
- 全局 `risk_control_enabled` 状态变化。

## 建议修复

`Reload` 已经读取了 `previous := m.snapshot.Load()`，可以复用该快照判断是否需要记录：

```go
func shouldLogConfigLoaded(previous *activeConfigSnapshot, storage storageConfig, active ActiveConfig) bool {
    return previous == nil ||
        previous.storage.ConfigVersion != storage.ConfigVersion ||
        previous.active.RiskControlEnabled != active.RiskControlEnabled
}
```

只有该函数返回 `true` 时才调用 `LogInfo(EventConfigLoaded, ...)`。已有的无效 Token 日志仍保持“仅变化时记录”的逻辑。

## 回归测试建议

- 连续两次以相同配置调用 `Reload`，`config_loaded` 只出现 1 次；
- 切换全局风控开关后增加 1 次；
- 提升配置版本后再增加 1 次。

上述修复和回归测试已在本地通过。

YogaSakti:
## Problem

`ConfigManager.refreshLoop` reloads the Prompt Guard config every 5 seconds (`backend/internal/securityaudit/prompt_config_store.go:436`), and `Reload` logged `prompt_guard.config_loaded` on every successful load, whether or not anything had changed (`prompt_config_store.go:141` before this change).

One instance therefore emits up to ~17,280 identical `config_loaded` lines per day, which is what #5732 reports (~12k lines counted in a single day of logs). The noise inflates log storage and hides the lines that matter — real config changes, degraded reloads and token warnings.

The same file already draws this distinction one function below: `logInvalidTokenEndpoints` is documented as "warns once per change (not on every 5s refresh)" (`prompt_config_store.go:159`). `config_loaded` was the outlier.

## Change

**`backend/internal/securityaudit/prompt_config_store.go`**

- `Reload` now emits `config_loaded` only when the reload carries news, via a new `shouldLogConfigLoaded(previous, storage, active)`:
  - first successful snapshot (`previous == nil`);
  - `ConfigVersion` changed — every admin save bumps it by one under the advisory lock in `UpdateConfig` (`prompt_config_store.go:280`), so the version is a reliable change signal for the audit config itself;
  - `RiskControlEnabled` changed — the global gate is a separate setting (`SettingKeyRiskControl`) fed into `ActiveFromStorage` and does not touch `ConfigVersion`, so it needs its own clause.
- `clearLoadError` now reports whether a load error was pending, and `Reload` logs `config_loaded` on that transition too. Without this clause, a service that recovers from a stretch of `config_reload_degraded` would go silent instead of confirming the recovery, since the version is usually unchanged. Redis invalidation reloads (`subscribeLoop`) are unaffected: they arrive with a new version and still log.

**`backend/internal/securityaudit/prompt_config_test.go`**

- `TestConfigLoadedIsLoggedOnlyWhenSomethingChanged` captures `slog` output the way `prompt_logging_test.go:17` does and walks one manager through: first load, two unchanged reloads, a risk-control flip, a version bump, a failed reload, and a recovery — asserting the running `config_loaded` count at each step.

## Verification

```
cd backend
go test -tags=unit ./...          # ok, 53 packages
go test -tags=integration ./...   # ok, 47 packages
golangci-lint run ./...           # 0 issues (v2.9.0, as pinned in backend-ci.yml)
gofmt -l internal/securityaudit/  # clean
```

The new test fails on unpatched `main` — reverting only `prompt_config_store.go` gives:

```
--- FAIL: TestConfigLoadedIsLoggedOnlyWhenSomethingChanged
    Error: Not equal: expected: 1  actual: 3
    Messages: TTL refreshes of an unchanged config must stay silent
```

## What the test would catch

A regression that turns `config_loaded` back into a 5-second heartbeat, and equally a regression in the other direction: dropping the risk-control clause, the version clause, or the recovery clause each flips one of the four assertions.

Closes #5732

### DIFF
--- a/backend/internal/securityaudit/prompt_config_store.go
+++ b/backend/internal/securityaudit/prompt_config_store.go
@@ -136,12 +136,26 @@ func (m *ConfigManager) Reload(ctx context.Context) error {
 	previous := m.snapshot.Load()
 	m.snapshot.Store(&activeConfigSnapshot{storage: cloneStorageConfig(storage), active: cloneActiveConfig(active), loadedAt: now})
 	m.configUntrusted.Store(false)
-	m.clearLoadError()
+	recovered := m.clearLoadError()
 	m.logInvalidTokenEndpoints(previous, active)
-	LogInfo(EventConfigLoaded, map[string]any{
-		"config_version": storage.ConfigVersion, "status": "loaded",
-	})
+	// refreshLoop calls Reload every 5s, so logging every successful load turns
+	// config_loaded into a heartbeat that buries real config changes.
+	if recovered || shouldLogConfigLoaded(previous, storage, active) {
+		LogInfo(EventConfigLoaded, map[string]any{
+			"config_version": storage.ConfigVersion, "status": "loaded",
+		})
+	}
 	return nil
+}
+
+// shouldLogConfigLoaded reports whether a successful reload carries news: the
+// first snapshot, a new config version (every admin save bumps it under the
+// advisory lock in UpdateConfig) or a flip of the global risk control gate,
+// which lives in its own setting and so leaves the version untouched.
+func shouldLogConfigLoaded(previous *activeConfigSnapshot, storage storageConfig, active ActiveConfig) bool {
+	return previous == nil ||
+		previous.storage.ConfigVersion != storage.ConfigVersion ||
+		previous.active.RiskControlEnabled != active.RiskControlEnabled
 }
 
 // logInvalidTokenEndpoints warns once per change (not on every 5s refresh)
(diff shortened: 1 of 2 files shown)


YogaSakti:

PR #6042, opened 2026-08-21
feat(openai): DSML leak guard - detect and heal deepseek-v4 tool calls leaked as plain text

weicheng59:
**Problem.** deepseek-v4 intermittently emits DSML tool-call blocks as plain `delta.content` text instead of structured `tool_calls` deltas: the opening tag is eaten serving-side, the parameter body and closing tags (`…</｜DSML｜tool_calls>`, usually with a `<thinking>` head) stream out as assistant text, the turn ends `finish_reason:"stop"`, and no `tool_calls` delta ever arrives. Clients see the agent "stop mid-task". Once a leaked turn enters history the model imitates it, so the failure cascades. Same phenomenon reported across the ecosystem: opencode#24566, cherry-studio#14714, vllm#41240.

**Mechanism.** Response side, raw chat-completions lane (enforce): a content-hold state machine lets reasoning stream live while holding the text channel through a 128-byte sniff window; a suspicious prefix (`<thinking>`, `<｜`) or the DSML marker holds to the terminal frame. Leak verdict (marker seen, no structured tool call) discards the held frames — including finish/usage/[DONE] — re-POSTs the identical body to the same account, and continues the same client SSE with the retry. Retry exhaustion, resend failure, hold caps (1 MiB / 5 min), or mid-stream abort all release the held output unchanged: held output is preserved. Upstream truncation still returns a typed stream-read error with retained usage, rather than reporting an incomplete response as success. SSE comment keepalives protect long holds; each retry re-scopes terminal tracking, the silent-refusal detector and response-model observation; billing uses the final attempt's usage with fallback to the prior attempt on truncation. Tool rounds with leaked leading text are scrubbed (marker reassembled across chunk boundaries) and released without retry. Native Responses lane: observe-only detection with an ops event at stream finalization. Request side: leaked DSML fragments are scrubbed from assistant-role history strings (user/tool untouched) on the chat-completions lanes including the anthropic-protocol lane — this cures the cascade. Every action lands a `Kind=dsml_leak` ops upstream-error event, persisted even on 200 responses.

**Config.**
```yaml
gateway:
  dsml_guard_enabled: true
  dsml_guard_models: ["deepseek-v4"]  # prefix match on the upstream model id
  dsml_guard_max_retries: 1           # exhausted = release held output unchanged
  dsml_guard_observe: false           # detect + log only, zero byte changes
```
The detector activates only for gated models on streaming requests carrying a non-empty `tools` array.

**Tests.** 28 unit cases with synthesized fixtures (no production captures): heal-by-retry client byte sequence, byte-invisibility on clean streams and in observe mode, retry-exhaustion release, split-marker detection and scrub across chunk boundaries, silent-refusal failover across retries, response-model observation reset, usage fallback on truncated retries, history-scrub role isolation, activation gating, mid-hold abort fail-open, Responses-lane observe verdicts, and a scrub pure-function table.

**Soak evidence.** A production deployment of this patch ran observe-then-enforce on live deepseek-v4 traffic. In a ~3-hour observe window covering 308+ requests, 4/4 response-side leaks were detected with zero false positives — 4/4 correlation against an independent rescan of raw traffic captures — and 76 requests carried leaked fragments in assistant history. In the following ~14-hour enforce window, 3 leaks were detected and retried: 2 healed by the same-account retry (client received a clean stream), 1 exhausted retries and was released unchanged (fail-open). Zero panics, zero client-visible errors, and leaked-marker-in-request counts dropped to zero once the history scrub went live.

**Current validation.** Merged upstream main at 32682a4f8. Targeted unit tests for DSML, raw streaming and silent refusal passed. Truncation regressions assert typed errors, retained partial output and prior-attempt usage fallback; a discarded attempt's terminal cannot mask a truncated retry.

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -999,6 +999,15 @@ type GatewayConfig struct {
 	// OpenAIPassthroughAllowTimeoutHeaders: OpenAI 透传模式是否放行客户端超时头
 	// 关闭（默认）可避免 x-stainless-timeout 等头导致上游提前断流。
 	OpenAIPassthroughAllowTimeoutHeaders bool `mapstructure:"openai_passthrough_allow_timeout_headers"`
+	// DSMLGuardEnabled: deepseek-v4 DSML 泄漏防护总开关。拦截「工具调用漏成正文」的坏流并原地重试，
+	// 同时从请求历史的 assistant 消息中剥离已泄漏的 DSML 碎片（治级联）。
+	DSMLGuardEnabled bool `mapstructure:"dsml_guard_enabled"`
+	// DSMLGuardModels: 防护生效的上游模型前缀列表（对 upstreamModel 做前缀匹配）。
+	DSMLGuardModels []string `mapstructure:"dsml_guard_models"`
+	// DSMLGuardMaxRetries: 判定泄漏后对同账号的最大原地重试次数。
+	DSMLGuardMaxRetries int `mapstructure:"dsml_guard_max_retries"`
+	// DSMLGuardObserve: true 时只检测记录、不扣流不重试不清洗（灰度用）。
+	DSMLGuardObserve bool `mapstructure:"dsml_guard_observe"`
 	// OpenAICompactModel: /responses/compact 上游使用的模型。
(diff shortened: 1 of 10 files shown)


weicheng59:

weicheng59:

PR #6061, opened 2026-08-21
fix(ops): avoid mixing cgroup and host memory metrics

shunwang-crypto:
## Problem

When Sub2API runs in **Docker + cgroup v2 with no memory limit set**, the ops dashboard reports a wildly wrong system memory usage — e.g. `74 MB / 23.4 GB (0.3%)` — even though the host is actually using ~16 GiB / 23 GiB.

In that environment:

```
/sys/fs/cgroup/memory.current = 64573440       # ~61 MiB, container usage
/sys/fs/cgroup/memory.max      = max            # no limit
/proc/meminfo MemTotal         = 24550208 kB    # host total
```

## Root cause

In `backend/internal/service/ops_metrics_collector.go`:

- `readCgroupMemoryBytes()` reads `memory.current` fine but, since `memory.max = "max"`, returns `(used=<container>, total=0, ok=true)`.
- `collectSystemStats()` then sets `memoryUsedMB` from the **container** value, but — unable to derive a cgroup total — fills `memoryTotalMB` from the **host** via gopsutil.
- Percentage becomes `container_used / host_total` = `~60 MB / 23 GB ≈ 0.3%`.

So the metric mixed two different sources: **container "used" ÷ host "total"**.

## Fix

Introduce `resolveMemoryStats(...)`, which produces a single **self-consistent** `(used, total, percent)` trio from **one** source and never mixes them:

1. Use **cgroup** metrics only when the cgroup exposes **both** a current usage **and** a concrete limit (`memory.max != "max"`, i.e. `total > 0`).
2. Otherwise (`memory.max = "max"` or incomplete cgroup data), all three values fall back **entirely** to host/gopsutil.
3. Container `used` is never divided by host `total`.

CPU metric behavior is intentionally left unchanged (cgroup attempt → host fallback).

## Behavior

| Scenario | Result |
|---|---|
| `memory.current` valid + `memory.max = max` | all host metrics (correct host %) |
| `memory.current = 512 MiB` + `memory.max = 2 GiB` | `~25%` from cgroup |
| no cgroup (bare metal) | all host metrics |

## Tests

Adds `backend/internal/service/ops_metrics_collector_memory_test.go` covering:

- cgroup usage but `memory.max = max` → falls back entirely to host (the reported bug)
- explicit container limit (512 MiB / 2 GiB) → ~25% from cgroup
- no cgroup → host metrics
- no data → all `nil`
- host without total → keeps gopsutil percentage

```
$ go test ./internal/service/ -run 'ResolveMemoryStats|OpsMetricsCollector|CollectConcurrencyQueueDepth'
ok  github.com/Wei-Shaw/sub2api/internal/service
```

All new tests pass and existing collector tests remain green.

### DIFF
--- a/backend/internal/service/ops_metrics_collector.go
+++ b/backend/internal/service/ops_metrics_collector.go
@@ -601,24 +601,11 @@ func (c *OpsMetricsCollector) collectSystemStats(ctx context.Context) (*opsColle
 
 	sampleAt := time.Now().UTC()
 
-	// Prefer cgroup (container) metrics when available.
+	// CPU: prefer cgroup (container) metrics, falling back to host metrics when
+	// cgroup CPU accounting is unavailable.
 	if cpuPct := c.tryCgroupCPUPercent(sampleAt); cpuPct != nil {
 		out.cpuUsagePercent = cpuPct
 	}
-
-	cgroupUsed, cgroupTotal, cgroupOK := readCgroupMemoryBytes()
-	if cgroupOK {
-		usedMB := int64(cgroupUsed / bytesPerMB)
-		out.memoryUsedMB = &usedMB
-		if cgroupTotal > 0 {
-			totalMB := int64(cgroupTotal / bytesPerMB)
-			out.memoryTotalMB = &totalMB
-			pct := roundTo1DP(float64(cgroupUsed) / float64(cgroupTotal) * 100)
-			out.memoryUsagePercent = &pct
-		}
-	}
-
-	// Fallback to host metrics if cgroup metrics are unavailable (or incomplete).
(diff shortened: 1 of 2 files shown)


shunwang-crypto:

Wei-Shaw:
维护审计反馈：改动方向和主体逻辑没有发现阻塞，但当前仍不能合并，因为 golangci-lint 在本 PR 代码中报 QF1001（backend/internal/service/ops_metrics_collector.go 的 resolveMemoryStats 附近）。

请按 staticcheck 建议简化该布尔判断，并在最新 main 上跑通 golangci-lint。这个缺口修复后可继续进入合并复核。

PR #6062, opened 2026-08-21
fix(openai): complete Codex Responses bridge and native search compatibility

laoxiechuzheng:
Fixes #5883

## Status: rebased on latest official main - mergeable

Rebased on the latest official `main` at `b5827cfd5` (v0.1.183 line). Conflicts with upstream #6078 (typed tool-call item IDs) are resolved: the official retyping path is kept, and this PR layers the restored-custom-tool input normalization on top. Head source: `424dfc266`. GitHub reports **MERGEABLE** and the CLA check passed. Fork CI is fully green at `424dfc266` (shell / test / golangci-lint / frontend / security scan).

## What this PR adds

1. Chat fallback continuation and tool-state bridge - DeepSeek reasoning/continuation state across Chat fallback turns, msg_*/chatcmpl_* id normalization, owner-bound response IDs, tool declarations restored from previous_response_id, agent_message task text preserved, type-aware function/custom/namespace collision handling.
2. Native Claude hosted search over Responses - server_tool_use / web_search_tool_result converted to real web_search_call, no orphan function_call_arguments deltas, pending search close + late-result dedup, placeholder input:{} replaced instead of {}{...}.
3. Namespace custom children (functions.exec) and web_search alias - flatten/restore with namespace + Custom type mapping, sparse tool-call indexes, request-level web_search to web.run alias.
4. Grok / GLM / Claude code-mode exec compatibility - narrowly scoped normalization for declared Codex custom exec tools only; GLM normalization enabled only when the final upstream model is GLM.
5. Claude thinking budget vs max_tokens - explicit client max_output_tokens caps the thinking budget; otherwise default max_tokens is raised above thinking.budget_tokens. Regression tests cover high/max/default/explicit.
6. Undeclared tool repair and Alpha Search isolation - one-shot protocol repair for undeclared streamed tool calls; transient Alpha Search 5xx no longer cools down ordinary Responses accounts; Chat fallback keeps encrypted reasoning replay.
7. Claude 5 adaptive thinking - the Claude 5 family (opus/sonnet/fable/haiku-5, including dated suffixes) converts reasoning effort to output_config.effort + thinking.type=adaptive; older models keep enabled+budget_tokens with the max_tokens reconciliation. Fixes the upstream 400 where thinking.enabled is rejected for these models.
[...]

PR #6078, opened 2026-08-22
fix(openai): keep restored tool-call item IDs typed

### LINKED ISSUE #6071: [BUG]Responses API 工具调用类型被错误转换：custom_tool_call 使用了 fc_ ID，导致 Codex 续聊失败
### 问题描述

使用 Responses API 中转服务时，返回的工具调用条目类型与 ID 前缀不一致：

```text
条目类型：custom_tool_call
条目 ID：fc_...
API 要求：ctc_...

Codex 在后续续聊时会把该历史条目重新发送给 OpenAI Responses API，随后请求失败：
{
  "error": {
    "code": "invalid_value",
    "message": "Invalid 'input[216].id': 'fc_09f77ac43cf7db36016a8920e7934487d0a9fd08e9db431d47'. Expected an ID that begins with 'ctc'.",
    "param": "input[216].id",
  }
}

观察到的响应条目
中转返回或保存的条目类似：
{
  "type": "custom_tool_call",
  "id": "fc_...",
  "call_id": "call_...",
  "name": "exec",
  "input": "..."
}

这里存在类型不匹配：

function_call       -> fc_
custom_tool_call    -> ctc_
message              -> msg_
reasoning            -> rs_

影响
一旦该条目进入 Codex 的本地历史：
1. 当前任务无法继续；
2. 每次发送新消息都会重新提交该错误条目；
3. OpenAI Responses API 拒绝整个请求；
4. 用户只能修改本地会话历史或新建任务；
5. 工具调用结果本身可能仍然有效，但因为条目 ID 类型错误而无法续聊。

期望行为
请确保中转服务在转换工具调用时保持类型和 ID 前缀一致：
function_call       使用 fc_
custom_tool_call    使用 ctc_


请重点检查以下逻辑：
- OpenAI Responses API 与 OpenAI 兼容格式之间的工具调用映射；
- function_call 和 custom_tool_call 的类型转换；
- 历史消息重放时是否错误复用了原始 fc_ ID；
- 是否把 response item 的 id 错误地当成了工具结果的 call_id；
- HTTP 和 WebSocket 两条响应路径是否使用了不同的转换逻辑。
如果中转服务无法保证 ID 前缀正确，建议在输出前：
- 保持原始条目类型，不要把 function_call 改成 custom_tool_call；
- 或删除错误的 response item id，但保留正确的 call_id；
- 不要把 fc_... 直接改名成 ctc_...，除非这是重新生成的合法条目 ID。

建议增加的回归测试
请增加以下测试：
function_call + fc_... => 保持 function_call
custom_tool_call + ctc_... => 保持 custom_tool_call
custom_tool_call + fc_... => 拒绝、修正或删除错误 id，不得原样转发
同时验证工具结果使用：call_id
进行关联，而不是使用 response item 的：id

补充信息
该问题发生在 Codex Desktop 续聊阶段，但根因可能来自中转服务返回了类型与 ID 前缀不匹配的历史条目。
相关错误并不是普通网络错误，而是 Responses API 对 response item 类型和 ID 前缀进行严格校验后拒绝请求。

### PR DESCRIPTION
**Pull request by YogaSakti**
Closes #6071.

## Problem

`restoreClientToolValue` raises an upstream `function_call` back to `custom_tool_call` / `tool_search_call` but never touches the item's `id`, so the client receives a `custom_tool_call` carrying the upstream's `fc_` ID. The Responses API treats the ID prefix as part of the item type, and the repository already encodes that rule — `openAIResponsesInputItemIDPrefix` (`internal/service/openai_responses_item_id.go:11-32`) requires `ctc` for `custom_tool_call` and `tsc` for `tool_search_call`.

The mismatch is silent until the item lands in the client's history. Any later replay to an upstream that validates item IDs fails the whole request:

```
Invalid 'input[216].id': 'fc_09f77...'. Expected an ID that begins with 'ctc'.
```

and every following message re-submits the same bad item, so the session cannot continue.

The mirror-image direction already enforced the rule: the lowering path reconciled `ctc_`/`tsc_`/`ctco_`/`tso_` IDs before handing an item to a function-only upstream. Only the raising direction was missing.

## Changes

**`backend/internal/pkg/apicompat/responses_client_tools.go`**

- New `retypedResponsesToolCallItemID` re-prefixes an item ID to match the type it is raised to, preserving the suffix so the ID stays stable and unique per upstream item. IDs with no known tool-call prefix are left alone rather than guessed at, and item types that constrain no prefix are untouched.
- Applied at all four sites that raise a `function_call`: `restoreClientToolValue` (non-stream), `restoreResponsesOutputClientTools` (terminal output), and the stream restorer's `response.output_item.added` / `response.output_item.done`.
- `responsesClientToolStreamCall` gains `clientItemID`. The upstream ID stays in `itemID` and keeps keying `r.calls`, so later `response.function_call_arguments.*` events still match the call; `clientItemID` is what goes out on the item events and on `response.custom_tool_call_input.delta` / `.done`.
- `dropInvalidLoweredFunctionItemID` is renamed to `normalizeLoweredFunctionItemID` and now maps a `ctc_`/`tsc_` **call** ID back to `fc_` instead of dropping it. Output item IDs (`ctco_`, `tso_`) still have no function-protocol counterpart and are still dropped.

**`backend/internal/pkg/apicompat/responses_client_tools_test.go`** — two assertions in `TestAdaptResponsesClientTools_LowersDeclarationsHistoryChoiceAndNamespaces` updated from "ID is dropped" to "ID is mapped back", matching the change above. Called out below because it is a deliberate behavior change, not a test fix.

**`backend/internal/pkg/apicompat/responses_client_tools_item_id_test.go`**, **`..._item_id_helper_test.go`** (new) — regression coverage.

## Why the lowering direction had to change too

Retyping only the raising direction is not enough, and I only found this because it broke an existing test. `TestOpenAIWSHTTPBridgeAPIKeyReusesClientToolMappingWhenFollowupOmitsTools` asserts that the WebSocket HTTP bridge replays the tool call to the upstream with the upstream's own `fc_custom_1` ID. The bridge replays *restored* items, so once the client-facing ID became `ctc_custom_1`, the old drop-on-lowering behavior removed it and the upstream saw a `function_call` with no ID.

Mapping `ctc_X` back to `fc_X` makes that round trip lossless, and it matches the convention `normalizeCodexCallIDForItemType` already uses for call IDs (`internal/service/openai_codex_transform.go:100-115`): the suffix is the identity, the prefix is the type tag.

**The trade-off, stated plainly.** An ID sub2api minted round-trips exactly. An ID that a client minted itself — a genuine `ctc_` from an upstream that speaks custom tools natively, replayed later through a function-only upstream — now reaches that upstream as a synthesized `fc_` ID where it previously reached it with no ID at all. That is the case the old comment on `shouldStripOpenAIResponsesInputItemID` warns about ("a fabricated ID may point at a different upstream object"). I could not observe whether any upstream resolves these IDs against stored state rather than only validating the prefix, so I have not claimed it is harmless.

If you would rather keep the strict drop, the alternative is to leave `normalizeLoweredFunctionItemID` dropping and instead teach the WS bridge's replay collector to record the pre-restore upstream item. That is a larger change in `openai_ws_http_bridge.go` and touches the failover replay path, so I did not take it unasked — say the word and I will redo it that way.

## Verification

```
cd backend
go test -tags=unit ./...          # pass
go test -tags=integration ./...   # pass
golangci-lint run ./...           # 0 issues
```

Red/green check, source reverted to `main` with the new test file kept:

```
--- FAIL: TestRestoreResponsesClientToolPayload_RetypesToolCallItemIDs
--- FAIL: TestRestoreResponsesOutputClientTools_RetypesToolCallItemIDs
--- FAIL: TestResponsesClientToolStreamRestorer_RetypesCustomToolCallItemID
--- FAIL: TestResponsesClientToolStreamRestorer_RetypesToolSearchCallItemID
```

The first failure reproduces the reported shape exactly — `"type":"custom_tool_call"` with `"id":"fc_abc123"`.

## What the tests would catch

A regression that lets a raised item keep a foreign prefix (the reported bug), one that changes the client-facing ID without the lowering side recovering it (the WS replay would silently lose the upstream ID), and one that starts rewriting IDs it should leave alone — `item_*`, unprefixed IDs, and item types that constrain no prefix are all pinned.

## Not in this PR

`chatcompletions_responses_bridge.go` mints every synthesized output item ID with `generateItemID()`, i.e. `item_*`, including the `custom_tool_call` items it builds at lines 1229 and 1983. A replay of those to an upstream that validates IDs would fail the same way. I left it alone because `item_*` is that bridge's convention for *every* item type, message and reasoning included, so changing it is a separate decision rather than the same root cause. Happy to follow up if you want it.

Also worth flagging: #6062 edits the same four `if` branches in `responses_client_tools.go` to resolve custom tool *names*. The changes are orthogonal — that PR touches `name`, this one touches `id` — but whichever lands second will need a small textual rebase.


Monika-shipship:
### 中文补充审查意见：建议增加 OAuth Responses WS v2 回归测试

这个 PR 的核心修复方向是正确的：恢复 `function_call` 为 `custom_tool_call` / `tool_search_call` 时，item ID 的类型前缀也必须同步变化；同时降低回上游时要保留可用的 `fc_` 映射，否则 WebSocket HTTP bridge 的 follow-up 会丢失上游 item ID。

我在本地 Sub2API `v0.1.179` 的 OAuth + `mixed` 路由上复现了另一个容易遗漏的场景：

- Codex 续聊历史中已有 `custom_tool_call.id = fc_...`
- 请求命中 Responses WebSocket v2
- 如果只在 `forwardOpenAIPassthrough()` 内清洗，`Forward()` 的 WS 分支会先发送请求，旧 ID 仍会到达上游
- 将清洗逻辑放到 `Forward()` 的 HTTP/WS 分流之前后，同样的真实请求返回 HTTP 200

建议在本 PR 或后续 PR 增加一个回归测试，至少覆盖：

1. OAuth 账号启用 Responses WS v2；
2. input 中有 `custom_tool_call.id = fc_...` 和对应 `custom_tool_call_output.call_id`；
3. 捕获发往 WS 上游的 `response.create`；
4. 断言无效 native item ID 被清掉，`call_id` 和工具输出仍保留；
5. 再验证新响应恢复出的 custom/tool-search item 分别使用 `ctc_`/`tsc_`。

当前 `main` 已经有公共路径的通用历史 ID 清洗代码，因此这里更像是防回归测试，而不是要求重复实现另一套清洗器。

PR #6107, opened 2026-08-23
fix(openai): disable parallel tool calls for Responses Lite

superman2003:
## Summary
- force Responses Lite requests to send `parallel_tool_calls: false` before account routing
- preserve tools while applying the Lite-only constraint to managed and passthrough API-key paths
- add regression coverage for missing/true values and both API-key forwarding modes

## Why
OpenAI rejects Responses Lite requests with `X-OpenAI-Internal-Codex-Responses-Lite` when `parallel_tool_calls` is omitted or true. The existing generic normalization intentionally retained `true` when tools were present, so Lite requests could fail with HTTP 400 after the recent Responses compatibility changes.

## Tests
- `go test -tags=unit ./internal/service -run 'TestNormalizeOpenAIResponsesLiteParallelToolCalls|TestOpenAIGatewayServiceForward_(DisablesParallelToolCallsForResponsesLiteAPIKey|NormalizesResponsesLiteToolsForOAuth)' -count=1`
- `git diff --check`

The full service package was also run. Its Lite/OpenAI coverage passed; an unrelated DNS-dependent channel monitor test fails because the test hostname resolves differently in the current environment.

### DIFF
--- a/backend/internal/service/openai_gateway_forward.go
+++ b/backend/internal/service/openai_gateway_forward.go
@@ -56,6 +56,16 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 	if legacyIngressChanged {
 		body = legacyIngressBody
 	}
+	responsesLite := isOpenAIResponsesLiteHeader(c.GetHeader(responsesLiteHeader))
+	if responsesLite {
+		liteBody, changed, liteErr := normalizeOpenAIResponsesLiteParallelToolCalls(body)
+		if liteErr != nil {
+			return nil, liteErr
+		}
+		if changed {
+			body = liteBody
+		}
+	}
 	// 在分流到 passthrough / Codex transform / 原生 ChatCompletions 之前统一修正
 	// 显式为 null 的工具 Schema type，否则 upstream 的 400 会被归一成可重试的 502，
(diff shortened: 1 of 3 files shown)


TsaiYongChuan:
■ {"error":{"code":"unsupported_value","message":"X-OpenAI-Internal-Codex-Responses-Lite requires `parallel_tool_calls` to be
false.","param":"parallel_tool_calls","type":"invalid_request_error"}}

superman2003:
@TsaiYongChuan Thanks for the report. Is this error from the current `main` branch, or did you test this PR branch at commit `478d27a3f8`?

If it still occurs with this PR applied, could you please provide:

- account type (OpenAI OAuth or API key)
- whether `openai_passthrough` is enabled
- the relevant request/upstream logs

This PR normalizes Responses Lite requests to explicitly send `parallel_tool_calls: false` before the managed/passthrough and OAuth/API-key routing paths, so those details will help identify whether another forwarding path is bypassing the normalization.

superman2003:
已由 #6084 修复并合入，后续 #6148、#6157 进一步补全相关路径；本 PR 已被覆盖，故关闭。

superman2003:
已由 #6084 修复并合入，后续 #6148、#6157 进一步补全相关路径；本 PR 已被覆盖，故关闭。

PR #6122, opened 2026-08-23
fix(deps): bump dompurify to patch sanitizer-bypass XSS advisories

aeonframework:
Automated dependency bump to address disclosed CVEs in DOMPurify.

- **Package:** `dompurify` `^3.3.1` -> `^3.4.14` (also pins the `mermaid`-transitive copy, previously stuck at `3.3.3`, via `pnpm.overrides`)
- **Severity:** high (sanitizer bypass -> stored/DOM XSS)
- **Advisories closed:** ~18 GHSAs on 3.3.1 / ~14 on 3.3.3, including [GHSA-cj63-jhhr-wcxv](https://github.com/advisories/GHSA-cj63-jhhr-wcxv) (CVE-2026-65913)

**Why this matters here specifically:** `frontend/src/utils/sanitize.ts` calls `DOMPurify.sanitize(svg, { USE_PROFILES: { svg: true, svgFilters: true } })`, and its output is rendered via `v-html` in `ImageUpload.vue`'s SVG upload/preview path. GHSA-cj63-jhhr-wcxv is specific to the `USE_PROFILES` code path: DOMPurify rebuilds `ALLOWED_ATTR` as a plain array and looks attributes up via `ALLOWED_ATTR[lcName]`, so if `Array.prototype` is polluted elsewhere on the page (e.g. `Array.prototype.onclick = true`), the polluted property is treated as an allow-listed attribute and event handlers survive sanitization -- turning the sanitizer itself into an XSS vector. Bumping closes this and the other 17 advisories on the same major line without changing the public API used here.

Verified against the live OSV API that `dompurify@3.4.14` has zero open advisories before opening this PR, and re-ran `osv-scanner` against the regenerated lockfile to confirm both the direct and transitive (`mermaid`) copies now resolve to `3.4.14` with 0 dompurify findings.

Confirmed via `gh api repos/Wei-Shaw/sub2api/contents/frontend/package.json` that `main` still pins `^3.3.1` at the time of filing (not already fixed upstream).

No code changes outside `frontend/package.json` / `frontend/pnpm-lock.yaml` (lockfile-only regen, `pnpm 9`, scoped diff).

**Not in scope:** `xlsx@0.18.5` also carries 2 open advisories (GHSA-4r6h-8v6p-xvw6, GHSA-5pgg-2g8v-p4x9), but SheetJS stopped publishing fixed versions to npm (fix is only available via `cdn.sheetjs.com`), so a standard `npm`/`pnpm` bump cannot resolve it -- flagging for awareness rather than bundling into this PR.

Detected by [osv-scanner](https://google.github.io/osv-scanner/) + manual verification.

---
Filed by [Aeon](https://github.com/aeonfun/aeon).

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -22,7 +22,7 @@
     "@vueuse/core": "^10.7.0",
     "axios": "^1.18.0",
     "chart.js": "^4.4.1",
-    "dompurify": "^3.3.1",
+    "dompurify": "^3.4.14",
     "driver.js": "^1.4.0",
     "file-saver": "^2.0.5",
     "marked": "^17.0.1",
@@ -63,7 +63,8 @@
     "overrides": {
       "js-cookie": "3.0.7",
       "form-data@<4.0.6": ">=4.0.6",
-      "postcss@<8.5.18": ">=8.5.18"
+      "postcss@<8.5.18": ">=8.5.18",
+      "dompurify@<3.4.14": ">=3.4.14"
     }
   }
 }
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -8,6 +8,7 @@ overrides:
   js-cookie: 3.0.7
   form-data@<4.0.6: '>=4.0.6'
   postcss@<8.5.18: '>=8.5.18'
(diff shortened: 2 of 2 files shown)


aeonframework:

PR #6128, opened 2026-08-24
refactor(cn-providers): introduce provider registry for extensible CN supplier onboarding

zichuanwangcloud-gif:
## Summary

The CN provider support (kimi/zhipu/deepseek, landed in v0.1.174–179) currently hardcodes the platform set in ~15 places: `IsCNProvider` switches, the default-endpoint matrix (`platform × protocol × mode`), balance/quota probe parsing, `AllowedQuotaPlatforms`, `AllowedSchedulingThresholdPlatforms`, scheduler snapshot buckets, model→platform detection, and admin platform lists. Adding one more supplier means touching all of them, and missing one is easy — the platform CHECK constraint incident fixed in migration 224 (new users got no quota rows) showed how these scattered enumerations drift apart.

This PR consolidates all of that into a **provider registry** so onboarding a new CN supplier is a single `RegisterCNProvider` call in one new file, with zero changes to existing code paths.

## Design

- **`CNProviderSpec`** = metadata + optional behavior hooks:
  - Metadata: default base-URL matrix keyed by `(protocol, mode)`, model-detect prefixes/aliases, LiteLLM vendor key.
  - `BalanceProbe` / `QuotaProbe` hooks own the platform-specific parts of probing (endpoint derivation, auth header shape, extra headers, response parsing). A **nil hook means "no such endpoint"** and the platform degrades to reactive 402/429 handling — exactly how zhipu balance works today, now expressed structurally instead of by special-casing.
  - `SupportsNativeResponses` replaces the deepseek-only special cases in routing decisions (adaptive protocol, responses probing, `max_output_tokens` preservation).
- **Derived lists**: `AllowedQuotaPlatforms`, `AllowedSchedulingThresholdPlatforms`, and the concrete-platform enumerations (scheduler snapshots, admin lists, composite matching) are rebuilt from the registry on registration, so a newly registered provider automatically appears in quota settings, threshold evaluation, monitor capability checks, and admin UIs.
- **Builtin providers** live in `cn_provider_{kimi,zhipu,deepseek}.go`, each self-contained (spec + probe implementations). Registration order is centralized in `cn_provider_builtin.go` to preserve the historical list ordering (Go executes same-package `init()` in filename order, which would silently reorder derived lists).

## Behavior compatibility

No behavior change intended:

- Same default endpoints for every `platform × protocol × mode` combination (fallback chain: exact `(protocol, coding)` → `(protocol, payg)` → mode-agnostic key).
[...]

PR #6130, opened 2026-08-24
refactor(cn-providers): drop platform enum CHECK constraints in favor of application-level validation

zichuanwangcloud-gif:
## Summary

> ⚠️ Stacked on #6128 (CN provider registry). Please review/merge that first; this branch will be rebased onto `main` afterwards.

Removes the platform enum CHECK constraints from `user_platform_quotas.platform` and `composite_model_routes.target_platform` (migration 229), so onboarding a new CN supplier no longer requires a constraint migration.

## Motivation

These constraints froze the platform list in the database layer and have already caused the same class of incident twice:

- migration 157's CHECK predated grok → new-user registration's bulk quota insert aborted entirely → **new users got zero quota rows** (= no limits);
- migration 224 had to repeat the identical fix when kimi/zhipu/deepseek landed.

With #6128, the authoritative platform set lives in the CN provider registry, and **every write path already validates at the application layer**:

- `user_platform_quotas`: `service.IsAllowedQuotaPlatform` in `setting_update.go` and both admin user handler entry points;
- `composite_model_routes`: `isConcreteRequestPlatform` in the group admin route writer.

The DB-level enums only duplicated that list, so they are dropped. The ent field validator is relaxed to non-empty + max-length with a comment pointing at the application-layer source of truth.

## Test plan

- [x] `go build ./...` / `go generate ./ent` / `golangci-lint run` — clean
- [x] `go test ./migrations/... ./internal/...` — all pass
- [x] New migration test asserts both constraints are dropped and no new enum constraint is introduced
- [x] Existing 224 migration test untouched (historical migration content unchanged)

### DIFF
--- a/backend/ent/runtime/runtime.go
+++ b/backend/ent/runtime/runtime.go
@@ -2392,7 +2392,6 @@ func init() {
 		fns := [...]func(string) error{
 			validators[0].(func(string) error),
 			validators[1].(func(string) error),
-			validators[2].(func(string) error),
 		}
 		return func(platform string) error {
 			for _, fn := range fns {
--- a/backend/ent/schema/user_platform_quota.go
+++ b/backend/ent/schema/user_platform_quota.go
@@ -1,8 +1,6 @@
 package schema
 
 import (
-	"fmt"
-
 	"entgo.io/ent"
 	"entgo.io/ent/dialect"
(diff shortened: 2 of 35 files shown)


zichuanwangcloud-gif:
recheck

zichuanwangcloud-gif:
recheck

PR #6139, opened 2026-08-24
feat(gateway): configure model list read limit

xz-dev:
## Summary

- add `gateway.models_list_read_max_bytes` with the existing 8 MiB limit as its default
- apply the shared limit to generic upstream model sync, Codex manifests, and Antigravity model-list reads
- detect oversized Codex manifests with a one-byte sentinel instead of reporting truncated JSON as an invalid envelope

## Problem

A valid Codex-compatible model manifest can exceed the current hard-coded 8 MiB read cap. In the reproduced case, the upstream returned 192 models in an 8,674,607-byte JSON document. `io.LimitReader` truncated it at 8 MiB, and the gateway returned a retryable 502 with `unexpected end of JSON input` even though the upstream response was valid.

The generic model-sync and Antigravity paths already use a `limit+1` sentinel. This change gives all three model-list paths one operator-configurable bound and brings the Codex path in line with the explicit oversize behavior.

This follows the existing `gateway.upstream_response_read_max_bytes` configuration pattern introduced in `bf0bbe0`.

## Compatibility

- the default remains 8 MiB, so existing deployments keep their current memory bound
- operators with larger manifests can set, for example, `gateway.models_list_read_max_bytes: 16777216`
- `codexModelsManifestCacheBodyLimit` remains 1 MiB
- no account catalog, model ID, or provider behavior changes

## Verification

- `make test-unit`
- `make test-integration`
- `golangci-lint v2.9.0 run --timeout=30m ./...`
- Wire v0.7.0 regeneration
- regression coverage for a valid Codex manifest above 8 MiB and below 16 MiB

### DIFF
--- a/backend/cmd/server/wire_gen.go
+++ b/backend/cmd/server/wire_gen.go
@@ -194,7 +194,7 @@ func initializeApplication(buildInfo handler.BuildInfo) (*Application, error) {
 	groupCapacityService := service.NewGroupCapacityService(accountRepository, groupRepository, concurrencyService, sessionLimitCache, rpmCache)
 	groupHandler := admin.NewGroupHandler(adminService, dashboardService, groupCapacityService)
 	claudeUsageFetcher := repository.NewClaudeUsageFetcher(httpUpstream)
-	antigravityQuotaFetcher := service.NewAntigravityQuotaFetcher(proxyRepository)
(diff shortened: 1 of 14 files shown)


xz-dev:

PR #6152, opened 2026-08-24
fix(payment): refresh balance after fulfillment

### LINKED ISSUE #6151: Payment result page does not refresh credited balance after fulfillment
## Problem

After a balance recharge is fulfilled, the payment result page can show a successful order while the navigation/user balance still shows the pre-payment value. The auth store only refreshes user data on its periodic interval, so users perceive this as delayed crediting even though backend fulfillment has completed.

## Expected behavior

When a balance order reaches `COMPLETED`, refresh the current user profile once so the credited balance is reflected immediately.

## Actual behavior

The payment result page resolves/polls order status but does not refresh the auth store after fulfillment.

## Proposed fix

- Refresh `authStore.user` once when a balance order reaches `COMPLETED`.
- Do not refresh at `PAID` or `RECHARGING`, because fulfillment may not have completed yet.
- Keep payment success authoritative if the best-effort profile refresh fails.
- Add regression tests for initial completion, pending-to-completed polling, and refresh failure.

ranxi2001:
## Summary

- refresh the authenticated user once after a balance order reaches `COMPLETED`
- avoid refreshing at `PAID`/`RECHARGING`, before balance fulfillment is guaranteed
- keep the successful payment result intact if the best-effort profile refresh fails
- add regression coverage for completed orders, pending-to-completed polling, and refresh failures

Fixes #6151

## Verification

- `vitest run src/views/user/__tests__/PaymentResultView.spec.ts` (14 tests)
- `vue-tsc --noEmit`
- ESLint on modified files
- `git diff --check`

### DIFF
--- a/frontend/src/views/user/PaymentResultView.vue
+++ b/frontend/src/views/user/PaymentResultView.vue
@@ -107,6 +107,7 @@ import {
   readPaymentRecoverySnapshot,
 } from '@/components/payment/paymentFlow'
 import { usePaymentStore } from '@/stores/payment'
+import { useAuthStore } from '@/stores/auth'
 import { paymentAPI } from '@/api/payment'
 import type { PublicOrderVerifyResult } from '@/api/payment'
 import type { OrderStatus, PaymentOrder } from '@/types/payment'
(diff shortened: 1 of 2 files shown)


ranxi2001:

PR #6160, opened 2026-08-25
feat: add OrcaRouter as a first-class OpenAI-compatible provider

kuswardhanietidims-svg:
Adding [OrcaRouter](https://www.orcarouter.ai) as a concrete platform lets Sub2API operators provision upstream API-key accounts that reuse the existing OpenAI gateway forwarding path instead of hand-filling a custom base URL. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

This mirrors how Kimi / Zhipu GLM / DeepSeek were wired in as OpenAI-compatible upstreams: a `PlatformOrcaRouter` constant with a default base URL (`https://api.orcarouter.ai/v1`), inclusion in the OpenAI-compatible account predicate and protocol API-key resolution, composite-route detection for `orcarouter/*` model prefixes, channel-monitor probe adapter, and the platform CHECK-constraint migrations. On the frontend, OrcaRouter appears in the create-account dialog, platform catalogs, badges, model icons, and channel-monitor dialogs, plus the en/zh i18n labels.

Verified locally: `go build ./...` and `go vet ./...` clean, backend unit + integration tests pass, frontend `vue-tsc` and the full Vitest suite (1744 tests) pass. I also ran a live completion through the resolved provider path (`GetOpenAIBaseURL()` → `DefaultOrcaRouterBaseURL`, Bearer key from `GetOpenAIProtocolAPIKey()`) against `https://api.orcarouter.ai/v1/chat/completions` with model `orcarouter/fusion-mini` and got HTTP 200.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.


kuswardhanietidims-svg:

PR #6170, opened 2026-08-25
perf(usage): speed up API key usage aggregation

cyhhao:
## Summary

Addresses the `POST /api/v1/usage/dashboard/api-keys-usage` part of #5858.

This PR extracts and completes the API-key rollup work from #5859 as an independently reviewable change. It intentionally excludes the minute-metrics and system-log pagination changes from that PR. The original `pcmid` commits and authorship are preserved in the branch history, followed by the complete set of review-driven correctness fixes.

## Changes

- Add `usage_apikey_daily_rollups` and produce group and API-key daily buckets together with one `GROUPING SETS` scan inside the existing day-chunked rollup transaction.
- Read complete historical days from the derived table while keeping partial boundary days and the unpublished tail on `usage_logs`.
- Fall back to raw data when the rollup watermark or timezone state is missing or stale, preserving correctness while the rollup catches up.
- Make the shared group/API-key rollup state safe across late writes, API-key-only changes, retention cleanup, cross-barrier updates, timezone changes, interrupted synchronization, and transaction-scoped fallback execution.
- Render the API-key list before usage statistics finish loading, show per-row placeholders, and abort stale requests when the page query changes or the component unmounts.

## Correctness Model

- Retention advances only after all pending closed days have been published; archived buckets stay immutable after source rows are removed.
- Invalidation covers both `group_id` and `api_key_id`, but never rewinds the watermark behind the archival barrier.
- The API-key table migration rewinds an existing watermark before it can be reused, forcing historical API-key buckets to be backfilled.
- Historical buckets are bounded by the requested interval, including non-midnight start/end times.
- Timezone changes preserve the archived boundary bucket and rebuild the retained side under the new day boundaries.
- Both the normal and caller-owned transaction paths drain the bounded backlog before reporting success.

## Verification

- `go test -tags=unit ./migrations ./internal/repository -count=1` on the final upstream-based head
- Full frontend Vitest suite: 244 files, 1745 tests passed
- Focused `KeysView` suite: 10 tests passed on the final upstream-based head
- Frontend ESLint, TypeScript typecheck, and production build passed

The Docker CLI is unavailable in the local environment, so the PostgreSQL integration suite was not rerun locally. The included integration coverage exercises the real `GROUPING SETS` statement, watermark invalidation, retention barriers, timezone changes, multi-day backlog processing, and transactional fallback; upstream CI should run it with PostgreSQL.

## Rollout Notes

- Migrations `231_group_usage_rollup_archival.sql` and `232_usage_apikey_daily_rollups.sql` are additive/idempotent and follow the current migration sequence.
- No API response shape changes.
- Until the first historical backfill completes, the query remains correct through its raw-data fallback, but may not receive the full performance benefit yet.


cyhhao:
@pcmid @Wei-Shaw FYI: this PR carries the API-key usage aggregation part of #5859 as a focused change and preserves the original commits and authorship. It also includes the follow-up correctness fixes around backfill, retention-barrier invalidation, timezone transitions, and transactional backlog processing.

The current head passes the full upstream CI suite. The same change set has also been running successfully under a real workload, and we have not observed regressions.

No urgency or review request - just wanted to keep you informed and thank you for the original work.

cyhhao:
@Wei-Shaw 你好，补充说明一下：目前用户 Keys 页面在用量明细数据较多时加载会非常慢，主要瓶颈是 API Key 用量统计需要扫描和聚合较大的明细范围，已经直接影响日常使用。

这个 PR 将该查询改为使用 API Key 日汇总，并补齐了回填、归档屏障、时区切换和并发一致性处理。当前完整 CI 已通过，相关变更也已在真实负载下验证稳定。我们目前比较急需这个 PR 来解决 Keys 页面的性能问题，麻烦有空时帮忙看一下，感谢！

PR #6186, opened 2026-08-25
fix(monitor-v2): correct composite platform NULLIF syntax

jinsc123:
## Summary

- pass the required empty-string argument to `NULLIF` when resolving a composite group's account platform
- strengthen the existing regression test to assert the syntactically valid expression

## Root cause

The composite-platform aggregation added a one-argument `NULLIF(TRIM(a.platform))`. PostgreSQL requires exactly two arguments, so the channel monitor V2 overlap aggregation failed with `syntax error at or near ")"` and its watermark stopped advancing.

## Validation

- `go test -mod=readonly ./internal/repository -run 'TestChannelMonitorV2' -count=1`
- prepared the complete aggregation statement successfully with PostgreSQL (`PREPARE` only; statement not executed)
- `git diff --check`

### DIFF
--- a/backend/internal/repository/channel_monitor_v2_aggregation.go
+++ b/backend/internal/repository/channel_monitor_v2_aggregation.go
@@ -255,7 +255,7 @@ WITH dedup AS (
     -- group errors aggregate under platform 'composite', which is never an
     -- enabled config platform, and are filtered out of every monitor v2 query.
     lower(CASE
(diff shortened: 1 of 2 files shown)


jinsc123:

Wei-Shaw:
https://github.com/Wei-Shaw/sub2api/pull/6193

PR #6193, opened 2026-08-25
fix(channel-monitor-v2): fix composite platform aggregation SQL

antTing:
## Bug

After upgrading to v0.1.182, Channel Monitor V2 aggregation fails every five minutes with:

```text
pq: syntax error at or near ")"
```

The Composite group platform resolution introduced an invalid PostgreSQL expression:

```sql
NULLIF(TRIM(a.platform))
```

`NULLIF` requires two arguments. Because usage aggregation, error aggregation, rollups, and watermark updates run in one transaction, this error rolls back the entire aggregation window and prevents Channel Monitor V2 from updating.

## Fix

Use a valid two-argument expression:

```sql
NULLIF(TRIM(a.platform), '')
```

A regression assertion was added to require the valid expression and reject the invalid single-argument form.

## Verification

```bash
go test ./internal/repository \
  -run 'TestChannelMonitorV2(ErrorAggregation|UsageSuccess|SQLTaxonomy)' \
  -count=1
```

Result:

```text
ok github.com/Wei-Shaw/sub2api/internal/repository
```

### DIFF
--- a/backend/internal/repository/channel_monitor_v2_aggregation.go
+++ b/backend/internal/repository/channel_monitor_v2_aggregation.go
@@ -255,7 +255,7 @@ WITH dedup AS (
     -- group errors aggregate under platform 'composite', which is never an
     -- enabled config platform, and are filtered out of every monitor v2 query.
     lower(CASE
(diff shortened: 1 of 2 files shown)


antTing:

PR #6198, opened 2026-08-25
fix(accounts): reauthorize OpenAI with refresh token

xuhaihan:
## Summary
- route refresh-token validation through the existing account reauthorization flow
- update the existing OpenAI or Antigravity account instead of creating a duplicate account
- preserve the existing Grok SSO and refresh-token paths

## Tests
- `pnpm exec vitest run src/components/admin/account/__tests__/ReAuthAccountModal.grok.spec.ts`
- `pnpm build`


xuhaihan:
recheck

PR #6200, opened 2026-08-25
fix(openai): normalize cross-provider reasoning replay

xuhaihan:
## Summary
- remove non-portable non-empty `reasoning.content` arrays before replaying history to OpenAI Responses
- preserve portable reasoning metadata, summaries, encrypted content, IDs, and opaque extensions
- apply the normalization consistently to HTTP and WebSocket forwarding paths

## Tests


xuhaihan:
recheck

PR #6201, opened 2026-08-25
fix(openai): honor Codex session-id header

xuhaihan:
## Summary
- accept the Codex-standard `session-id` header for sticky routing
- prefer it over the legacy `session_id` form in WebSocket session resolution
- keep reconnect hashes stable across warmup and business turns

## Tests


xuhaihan:
recheck

PR #6206, opened 2026-08-25
fix(openai): complete Responses bridge with review follow-up

xuhaihan:
## Context

This PR is a fast-forward continuation of #6062 from its current head `9e47eab6b`. It preserves the six existing #6062 commits and adds two focused review follow-up commits from `xuhaihan/sub2api`.

It was opened from this fork because the original cross-fork head cannot be updated by the current GitHub account. This is not an independent reimplementation, and it contains no unrelated internal/FAT/CCloud changes. If accepted, it supersedes #6062 while retaining its commit history and attribution.

## Review follow-up

- make namespace custom/function alias restoration type-aware, including declared custom `web_search` conflicts
- prevent streamed tool-repair retries after assistant text or reasoning has already become visible
- reset retry-only stream state safely and preserve accumulated usage/service-tier metadata
- persist omitted follow-up tool declarations in the shared gateway cache so continuations survive Pod changes
- scope Claude code-mode hints to the final upstream model and an actual declared custom exec tool
- add streaming, non-streaming, cross-instance continuation, collision, and cache regression coverage
- fix current Go 1.27 / golangci-lint v2.13 findings in the #6062 commit chain

## Validation

- `go test ./internal/pkg/apicompat ./internal/repository ./internal/service -count=1`
- `golangci-lint v2.13.0 run --timeout=30m ./...`
- `git diff --check upstream/main...HEAD`
- community sensitive-content scan


xuhaihan:
Closing this PR as superseded by #6062. The review follow-up covered here (type-aware custom/function alias restoration, streamed repair safety, continuation tool-state persistence, and scoped provider hints) has been incorporated into #6062's current implementation, while #6062 has continued with additional compatibility fixes. Keeping both branches open now creates a duplicate and increasingly divergent review surface.

PR #6277, opened 2026-08-27
fix(openai): preserve declared API-key namespace calls

### LINKED ISSUE #6276: fix(openai): preserve API-key namespace calls for compatible custom providers
## Summary

OpenAI API-key accounts with a custom Responses-compatible `base_url` drop `input[].namespace` from historical `function_call` items even when the same request declares a `type: "namespace"` tool. Namespace-aware upstreams then reject the continuation because the declaration remains namespaced while the call is rewritten as a default-namespace call.

## Environment

- sub2api: `v0.1.183`
- client: Codex Desktop using `/v1/responses`
- account: OpenAI `apikey`, custom `base_url`, forced native Responses
- upstream: Responses-compatible proxy with namespace-tool support

## Error

```text
Missing namespace for function_call 'automation_update'. It does not exist in the default namespace. Round-trip the model's function_call item with its namespace field included.
```

## Minimal request shape

```json
{
  "model": "gpt-5.6-sol",
  "tools": [
    {
      "type": "namespace",
      "name": "mcp__codex_app",
      "tools": [
        { "type": "function", "name": "automation_update", "parameters": { "type": "object" } }
      ]
    }
  ],
  "input": [
    {
      "type": "function_call",
      "call_id": "call_1",
      "name": "automation_update",
      "namespace": "mcp__codex_app",
      "arguments": "{}"
    },
    { "type": "function_call_output", "call_id": "call_1", "output": "ok" }
  ]
}
```

Sending this shape directly to the upstream succeeds. Sending it through an API-key account causes sub2api to remove `input[0].namespace`, and the upstream returns the error above.

## Root cause

`OpenAIGatewayService.Forward` calls `stripOpenAIResponsesInputNamespaces`. For API-key accounts, `shouldKeepOpenAIResponsesToolCallNamespaces` always returns `false`, so the cleanup removes namespace from tool-call history even though top-level namespace declarations are preserved.

This differs from the OAuth cases tracked in #4761 and #5068: this report is specifically about namespace-capable custom API-key providers.

## Proposed behavior

For non-compact API-key requests that explicitly declare at least one `type: "namespace"` tool, preserve namespace on direct tool-call history while continuing to strip residual namespace fields from messages, outputs, reasoning items, and unknown shapes.

Requests without namespace tool declarations retain the current standard Responses API behavior.

TYzzt:
## Summary

- preserve `namespace` on direct historical tool-call items for non-compact OpenAI API-key requests that explicitly declare a `type: "namespace"` tool;
- continue stripping residual namespace fields from messages, outputs, reasoning items, and unknown shapes; and
- add policy and end-to-end forwarding coverage for namespace-capable custom Responses providers.

Closes #6276.

## Root cause

API-key forwarding preserved top-level namespace tool declarations but unconditionally stripped `namespace` from every direct input item. A continuation could therefore reach a namespace-aware custom provider with a namespaced declaration and an unqualified historical `function_call`, causing:

```text
Missing namespace for function_call 'automation_update'. It does not exist in the default namespace.
```

## Behavior

| API-key request | Tool-call history |
|---|---|
| Non-compact, declares `type: "namespace"` | Preserve tool-call namespace |
| No namespace tool declaration | Keep existing cleanup |
| Compact | Keep existing cleanup |
| Residual namespace on non-call items | Keep existing cleanup |

This uses the request declaration as the capability signal. Standard Responses requests without namespace tools are unchanged, while a custom provider that accepts the namespace declaration receives the matching call identity it requires.

## Tests

```text
go test -tags unit ./internal/service -run Namespace -count=1
go test -tags unit ./internal/service -count=1
git diff --check
```

All passed with Go 1.27. The full service unit suite completed in 164.769s.

## Scope

- Four backend service/test files.
- No database, frontend, or account configuration changes.

### DIFF
--- a/backend/internal/service/openai_gateway_forward.go
+++ b/backend/internal/service/openai_gateway_forward.go
@@ -121,7 +121,7 @@ func (s *OpenAIGatewayService) Forward(ctx context.Context, c *gin.Context, acco
 	}
 	if shouldStripOpenAIResponsesInputNamespaces(account, wsDecision.Transport, passthroughEnabled) {
 		keepToolCallNamespaces := shouldKeepOpenAIResponsesToolCallNamespaces(
-			account, wsDecision.Transport, passthroughEnabled, compactPath,
+			account, wsDecision.Transport, passthroughEnabled, compactPath, body,
 		)
 		body, err = stripOpenAIResponsesInputNamespaces(body, keepToolCallNamespaces)
 		if err != nil {
--- a/backend/internal/service/openai_responses_namespace.go
+++ b/backend/internal/service/openai_responses_namespace.go
@@ -77,23 +77,46 @@ func shouldStripOpenAIResponsesInputNamespaces(account *Account, transport OpenA
 //     故 OAuth 非 compact 请求必须保留。
 //   - compact 端点的 schema 不含该字段，携带即 400 `Unknown parameter:
 //     input[N].namespace`（issue #4761 正文），故 compact 一律清理。
-//   - API Key 出口是标准 Responses API（api.openai.com 或自定义 base_url），同样
-//     不认识该字段，维持全量清理；否则只能退化成
-//     openai_responses_rejected_field_retry 的逐项删除，6 次上限根本盖不住长历史。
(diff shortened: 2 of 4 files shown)


TYzzt:

PR #6291, opened 2026-08-27
fix(grok): sanitize Codex Responses requests

Whisper-stark:
## Summary

- strip top-level `metadata` before forwarding Grok Responses requests to xAI
- preserve Claude Code cache identity from the original request while keeping `metadata` out of the upstream payload
- enable the existing Responses tool-parameter root sanitizer for `PlatformGrok`
- replace root `anyOf`/`oneOf` schemas containing non-object branches with a permissive object schema and disable `strict` for that tool
- preserve valid object-only unions unchanged

## Why

Codex Desktop can send OpenAI-compatible `metadata` and tool schemas that xAI Responses rejects. The first case returns a generic upstream 400. The second commonly affects `codex_app__automation_update` with `invalid_client_tool_schema`.

The compatibility changes are limited to the Grok Responses path. Chat Completions behavior is unchanged.

## Tests

- `go test -tags unit ./internal/service -count=1`
- `go test ./...`
- targeted metadata, cache-identity, retry, platform-boundary, object-only-union, and mixed-union regression tests

Refs #819
Refs #6250

### DIFF
--- a/backend/internal/service/openai_gateway_grok.go
+++ b/backend/internal/service/openai_gateway_grok.go
@@ -74,8 +74,14 @@ func (s *OpenAIGatewayService) forwardGrokResponses(
 		}
 	}
 	// Derive the identity from the request xAI will actually see. This makes
-	// Codex Responses Lite additional_tools part of the stable tool prefix.
-	cacheIdentity := resolveGrokCacheIdentity(c, patchedBody, "", upstreamModel)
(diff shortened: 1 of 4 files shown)


Whisper-stark:

Whisper-stark:

MilesNext:
Fixes #6250

PR #6063, opened 2026-08-21
fix(openai): fail raw CC streams truncated before any terminal chunk

nyx-invoiss:
## Summary

- Detect raw Chat Completions streams that end without any terminal signal and treat them as upstream failures instead of silent successes.
- Fail over to another account when nothing has reached the client yet; surface a typed stream read error once bytes are already on the wire.
- Preserve existing billing semantics for client cancel/disconnect.

## Problem

The raw Chat Completions passthrough returned a nil error whenever the upstream body ended. An upstream that cut the SSE stream mid-generation therefore surfaced as HTTP 200 with usage 0/0: the client kept a half-finished answer, the request was not counted as an SLA failure, and ops saw nothing.

A clean EOF carries no transport error, so without an explicit terminal check a truncated generation is indistinguishable from a completed one.

Observed against `opencode.ai/zen/go` (deepseek-v4-pro): 11 of 1346 streams in 24h recorded a first token but no final usage, alongside a concurrent `connection reset by peer` against the upstream edge.

## Fix

Track whether the upstream ever emitted a terminal signal — `[DONE]`, a usage chunk, or a `finish_reason` — and treat a stream that ends without one as an upstream failure.

Matching only `[DONE]` would misfire on compatible upstreams that EOF right after their last chunk, and would drop billing for tokens that were actually delivered, so all three signals count.

Behavior on truncation:

- **Nothing written to the client yet** — `UpstreamFailoverError`, retried on another account with no client-visible half stream.
- **Bytes already on the wire** — typed `openAIUpstreamStreamReadError` with the new `upstream_stream_truncated` code, so the handler appends an SSE error frame and marks the request an SLA failure.

Client cancel/disconnect keeps the previous semantics: there the upstream read error is indistinguishable from a truncation, so partial usage is still billed rather than blamed on the account.

## Tests

Added coverage for:

- Truncation after client output begins (fails the request).
- Empty stream before any output (triggers failover).
- Stream read error after output.
- Missing `[DONE]` with usage present, and with `finish_reason` present (both still succeed).
- Client disconnect and client cancel truncation (both still bill).
- Terminal-state tracker unit coverage.

Validated against this branch's base:

- `go test -tags=unit ./...` — 54 packages ok
- `go test -tags=integration ./...` — 48 packages ok
- `golangci-lint v2.9 run ./internal/service/... --timeout=15m` — `0 issues`
- Go 1.26.6, matching `backend/go.mod` and the CI version gate

PR #6298, opened 2026-08-27
fix(openai): fail raw CC streams truncated before any terminal chunk

nyx-invoiss:
## Summary

- Detect raw Chat Completions streams that end without any terminal signal and treat them as upstream failures instead of silent successes.
- Fail over to another account when nothing has reached the client yet; surface a typed stream read error once bytes are already on the wire.
- Preserve existing billing semantics for client cancel/disconnect.
- Drive the full Chat Completions handler so usage is committed once and the SSE error frame matches the protocol.

This supersedes #6063. Rebased onto current `main` (`efb46db0a`) and adds the handler end-to-end coverage requested in review.

## Problem

The raw Chat Completions passthrough returned a nil error whenever the upstream body ended. An upstream that cut the SSE stream mid-generation therefore surfaced as HTTP 200 with usage 0/0: the client kept a half-finished answer, the request was not counted as an SLA failure, and ops saw nothing.

A clean EOF carries no transport error, so without an explicit terminal check a truncated generation is indistinguishable from a completed one.

Observed against `opencode.ai/zen/go` (deepseek-v4-pro): 11 of 1346 streams in 24h recorded a first token but no final usage, alongside a concurrent `connection reset by peer` against the upstream edge.

## Fix

Track whether the upstream ever emitted a terminal signal — `[DONE]`, a usage chunk, or a `finish_reason` — and treat a stream that ends without one as an upstream failure.

Matching only `[DONE]` would misfire on compatible upstreams that EOF right after their last chunk, and would drop billing for tokens that were actually delivered, so all three signals count.

Behavior on truncation:

- **Nothing written to the client yet** — `UpstreamFailoverError`, retried on another account with no client-visible half stream.
- **Bytes already on the wire** — typed `openAIUpstreamStreamReadError` with the new `upstream_stream_truncated` code, so the handler appends an SSE error frame and marks the request an SLA failure.

Client cancel/disconnect keeps the previous semantics: there the upstream read error is indistinguishable from a truncation, so partial usage is still billed rather than blamed on the account.

## Review follow-up (#6063)

Wei-Shaw asked to close the loop on the handler path:

> 请把「部分输出 + 上游截断」作为端到端用例，验证 usage 只提交一次且错误帧符合协议；同时确保无输出截断与既有终态路径不回归。

`ChatCompletions` already submitted usage on both the error and success branches (`submitChatUsage` before `return` in each), and `ensureOpenAIStreamReadErrorResponse` already classified typed stream-read errors. The missing piece was an end-to-end test through the handler.

New handler coverage in `openai_chat_completions_raw_truncation_test.go`:

- **Partial output + upstream truncation** — HTTP 200 with the already-written chunks, exactly one `event: error` frame whose `error.code` is `upstream_stream_truncated`, one usage commit, and `GetOpsStreamError` counting toward SLA.
- **Empty stream before any output** — fail over to a second account before headers are committed; the client only sees the recovered stream; usage is committed once from the successful account.
- **Missing `[DONE]` with usage** and **missing `[DONE]` with `finish_reason`** — still succeed, one usage commit, no error frame.

Service-level truncation tests from #6063 are unchanged.

## Tests

- `go test -tags=unit ./internal/handler -run 'TestChatCompletions_(TruncatedRawStreamAfterOutput|EmptyRawStreamBeforeOutput|MissingDoneWithUsage|MissingDoneWithFinishReason)'`
- `go test -tags=unit ./internal/service -run 'TestForwardAsRawChatCompletions_(Truncated|EmptyStream|StreamReadError|MissingDone|ClientDisconnectTruncation|ClientCancelTruncation)|TestOpenAIRawStreamTerminalState'`

### DIFF
--- a/backend/internal/handler/openai_chat_completions_raw_truncation_test.go
+++ b/backend/internal/handler/openai_chat_completions_raw_truncation_test.go
@@ -1,0 +1,336 @@
+//go:build unit
+
+package handler
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/Wei-Shaw/sub2api/internal/config"
+	"github.com/Wei-Shaw/sub2api/internal/pkg/openai_compat"
+	middleware2 "github.com/Wei-Shaw/sub2api/internal/server/middleware"
+	"github.com/Wei-Shaw/sub2api/internal/service"
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
+)
+
+const rawCCTruncationChatBody = `{"model":"deepseek-v4-pro","messages":[{"role":"user","content":"hello"}],"stream":true}`
+
+// 维护者对 #6063 的闭环要求：把「部分输出 + 上游截断」接到 Chat Completions
+// handler，验证 usage 只提交一次、错误帧符合协议，并保住无输出截断与既有终态路径。
+
+func TestChatCompletions_TruncatedRawStreamAfterOutput_CommitsUsageOnceAndWritesTypedError(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	h := newRawCCTruncationHandler(t, []service.Account{
+		rawCCTruncationAccount(8101, 1),
+	}, map[int64]string{
+		8101: rawCCTruncatedAfterOutputSSE(),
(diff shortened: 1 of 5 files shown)


nyx-invoiss:
Closing this one — it was opened under the wrong GitHub account (`nyx-invoiss`). I'll reopen from `william-drakemond` on another machine.

The branch is still on the fork: `william-drakemond:fix/openai-raw-stream-truncation-e2e`.

PR #6300, opened 2026-08-27
Match real Codex CLI TLS fingerprint for OpenAI OAuth forwarding

JnyRoad:
## Summary
- OpenAI Codex OAuth outbound traffic went out over Go's default `crypto/tls` ClientHello, which is trivially distinguishable from the real Codex CLI: cipher suite table, `supported_groups`, and extension set all differ, and real Codex CLI (reqwest 0.12 + rustls 0.23, `aws_lc_rs` provider, compiled without the `http2` feature — confirmed from the official `openai/codex` source, `codex-rs/http-client`) never sends ALPN at all.
- Add a `tlsfingerprint.Profile` matching the real client's cipher suites / `supported_groups` / `ec_point_formats` / extension set, cross-checked against three independent packet captures of a real `codex-tui` session (see `specs/002-codex-tls-fingerprint/research.md` for the full evidence trail — this doc isn't part of the diff, kept locally).
- Real rustls clients also reshuffle their ClientHello extension order on every connection (three captures showed the same cipher/group/extension **set** each time, but three different extension **orders**) — a permanently fixed order is itself a distinguishing signal. Added a `RandomizeExtensionOrder` switch to `tlsfingerprint.Profile` that shuffles the extension list once per TLS connection; off by default so the existing Claude Code / Node.js profile is unaffected, on for the new Codex profile.
- Wired the profile into `doOpenAIUpstream`: an account's explicitly configured TLS fingerprint (via the existing `TLSFingerprintProfileService`) still wins if set; otherwise OpenAI Codex OAuth accounts get the new profile automatically instead of requiring a manual per-account opt-in. Non-OAuth / non-OpenAI accounts are unaffected.

## Test plan
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` — full repo, 0 failures (caught and fixed a real regression along the way: three `internal/handler` test fixtures embedded `service.HTTPUpstream` and only overrode `Do`, panicking once `doOpenAIUpstream` started calling `DoWithTLS` unconditionally — added matching `DoWithTLS` overrides)
- [x] TDD red→green cycle followed throughout (tests written and confirmed failing before each implementation step)
- [x] `wire_gen.go` regenerated via `google/wire` (not hand-edited) after adding the new DI wiring

### DIFF
--- a/backend/cmd/server/wire_gen.go
+++ b/backend/cmd/server/wire_gen.go
@@ -291,7 +291,7 @@ func initializeApplication(buildInfo handler.BuildInfo) (*Application, error) {
 	legacyEngine := securityaudit.NewLegacyModerationAdapter(contentModerationService)
 	coordinator := securityaudit.NewCoordinator(legacyEngine, promptService)
 	gatewayHandler := handler.ProvideGatewayHandler(gatewayService, openAIGatewayService, geminiMessagesCompatService, antigravityGatewayService, userService, concurrencyService, billingCacheService, usageService, apiKeyService, usageRecordWorkerPool, errorPassthroughService, contentModerationService, userMessageQueueService, configConfig, settingService, coordinator)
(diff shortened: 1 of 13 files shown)


JnyRoad:
> 
> 
> 
> 
> 

JnyRoad:
recheck

JnyRoad:

PR #6323, opened 2026-08-28
fix(monitor-v2): fix composite error aggregation SQL

ziyue67:
## Summary\n\n- Fix the composite-group error aggregation SQL so NULLIF receives its required empty-string comparison value.\n- Add a regression assertion for the corrected SQL template.\n\nThe malformed NULLIF caused PostgreSQL syntax errors during every Channel Monitor V2 bootstrap/recompute transaction, rolling back usage aggregation and leaving progress at 0%.\n\n## Validation\n\n- git diff --check\n- Go unit tests should run in CI; local Go runtime was unavailable in the development environment.


ziyue67:
Closing per request; this change will not be submitted upstream.

PR #6389, opened 2026-08-30
feat(pricing): add dynamic catalog and upstream price sync

lori-super:
## Summary

- add a dynamic public model-pricing catalog with provider notes, official-price sources, and admin controls
- add production-key upstream price monitoring with passive ledger reconciliation, exclusive active probes, cache-price inference, and per-request price-page discovery
- atomically apply channel pricing and display multipliers with 20% markup, half-up rounding, no-op detection, snapshot validation, and CAS-safe rollback
- add a managed upstream model catalog for free `/v1/models` discovery, explicit adoption, ignored/retired states, incomplete-scan protection, and 3-scan + 24-hour suspected-retirement handling
- add Cloudflare/OVH deployment support plus pricing preview and x5m5x probe/report tools

## Safety

- monitoring defaults to disabled and observe-only
- API keys are reused from existing account credentials and are never returned by the monitor API or persisted in monitor tables
- active probes acquire an exclusive finite-concurrency account lease, persist pending state before sending, and fail closed on overlap, timeout, billing-context drift, or incomplete attribution
- official display base prices remain unchanged; token display updates only the model multiplier
- model removal only changes probe scope and retains channel pricing, display pricing, mappings, and evidence history
- stale runs, account/config/catalog changes, shared multi-model rows, incomplete account coverage, and out-of-order snapshots are rejected

## Validation

- `go test ./... -run '^$' -count=1`
- targeted service/repository/handler/routes/migration/Wire tests for upstream pricing and model catalog
- `vue-tsc --noEmit`
- targeted Vitest suites for pricing API, monitor UI, sidebar, and navigation
- `node --check` for pricing/probe tools
- `git diff --check`
- secret-pattern scan of monitor and probe files

## Notes

- active probing requires a finite account concurrency setting and a production key not shared with programs outside this deployment
- newly managed models still require a single-model channel pricing row and a verified official display base price before automatic application can succeed
- this PR is intentionally not merged by automation


lori-super:
Closing because this PR was opened against the upstream repository by mistake; the intended target is the fork repository.

PR #6390, opened 2026-08-30
fix(openai): rearm persisted model circuit after expiry

apple-ouyang:
## Summary
- re-arm the persisted OpenAI account/model circuit after its cooldown expires
- deduplicate repeated upstream failures by request ID before advancing the durable streak
- make scheduled-test recovery model-scoped and conditional so stale results cannot clear a newer circuit

## Root cause
The transient runtime state could remain marked as persisted after the durable model rate-limit entry expired. Subsequent failures were therefore no longer eligible to open a new durable circuit, so repeated CII/AIINPUT failures could continue without the expected model-level circuit breaker.

## Validation
- `go test ./internal/service ./internal/repository`
- focused OpenAI transient circuit tests
- independent reviewer: PASS
- GitHub CI: test, golangci-lint, frontend, shell, and security scans passed

### DIFF
--- a/backend/internal/repository/account_repo.go
+++ b/backend/internal/repository/account_repo.go
@@ -2457,6 +2457,32 @@ func (r *accountRepository) ClearModelRateLimits(ctx context.Context, id int64) 
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


apple-ouyang:
Following up for maintainer review. The current required checks are green, and the PR is ready for feedback. Please let me know if any additional validation or changes are needed.

PR #6395, opened 2026-08-30
fix(frontend): clarify local timezone for expiry inputs

lylelllll:
## Summary

- Show the browser's IANA timezone next to account expiry inputs.
- Parse datetime-local values as browser-local wall-clock time before converting them to Unix seconds.
- Reuse the strict parser for redeem-code bulk updates and reject malformed or overflowing values.
- Add localized copy and timezone-aware tests.

## Context

datetime-local values do not carry timezone metadata. Treating a copied wall-clock value as another timezone can shift the resulting expiry timestamp. This change makes the local-time contract explicit and validates the input before submission. The backend API contract remains unchanged.

## Testing

- Vitest: 249 files, 1,789 tests
- vue-tsc --noEmit
- ESLint for changed files
- Vite production build


lylelllll:

PR #6430, opened 2026-08-31
fix(scheduling): replace repeated window-cost scans with incremental state

cyhhao:
## Summary

- replace repeated scheduling-time aggregation over `usage_logs` with a bounded, durable per-account window-cost state row
- maintain the state from successful `usage_logs` inserts with one statement-level upsert per account in each insert batch
- lazily rebuild missing or invalidated state once per account/window instead of backfilling the full history during deployment
- fail open without falling back to the expensive detail-table scan when the incremental state path is unavailable

## Root cause

Window-cost scheduling currently caches each account total for 30 seconds. Every cache miss executes `COUNT`/`SUM` aggregation against all matching `usage_logs` rows since the current window start. The query is correct, but its work scales with retained detail rows and is repeated indefinitely, so a normal scheduling safeguard becomes sustained database read amplification.

## Design

The migration adds `account_window_cost_state`, which is bounded to one row per account. It deliberately performs no startup backfill.

An `AFTER INSERT` transition-table trigger aggregates each usage-log batch before updating state, so the hot write path performs at most one narrow upsert per eligible Anthropic OAuth/setup-token account. Duplicate usage-log inserts remain excluded because the trigger sees only rows actually inserted by PostgreSQL.

The first read for a missing window acquires state rows in account-ID order, resets provisional totals, and rebuilds all missing accounts in one grouped query. The row locks provide the handoff with concurrent inserts: committed rows are included in the rebuild, while in-flight insert triggers wait and increment the initialized state after commit. Window/config changes and recent source updates or deletes invalidate the row and use the same rebuild path.

Steady-state cache misses are point reads from the state table. Keeping the existing Redis cache preserves current request-path behavior while removing the historical scan from production scheduling. State failures remain fail-open, but no longer reactivate the query that caused the load amplification.

## Validation

- `go test -tags unit ./internal/repository ./internal/service ./migrations`
- `go test -tags integration ./internal/repository -run '^TestAccountWindowCostStateBackfillsThenTracksIncrementally$' -count=1`
- `go vet ./internal/repository ./internal/service ./migrations`

The PostgreSQL integration test covers lazy backfill, incremental inserts, account-setting invalidation, corrected window starts, and source-row deletion.

### DIFF
--- a/backend/internal/repository/usage_log_repo_window_cost.go
+++ b/backend/internal/repository/usage_log_repo_window_cost.go
@@ -1,0 +1,205 @@
+package repository
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/lib/pq"
+)
+
+// GetAccountWindowCostsBatch returns the durable incremental scheduling cost for
+// accounts that share a window start. A state row is rebuilt from usage_logs only
+// on its first read or after the account window/source rows change.
+func (r *usageLogRepository) GetAccountWindowCostsBatch(ctx context.Context, accountIDs []int64, startTime time.Time) (map[int64]float64, error) {
+	result := make(map[int64]float64)
+	ids := normalizePositiveInt64IDs(accountIDs)
+	if len(ids) == 0 {
+		return result, nil
+	}
+	sort.Slice(ids, func(i, j int) bool { return ids[i] < ids[j] })
+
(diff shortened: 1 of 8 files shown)


cyhhao:
@codex review

cyhhao:
@codex review

cyhhao:
@codex review

cyhhao:
@codex review

PR #6438, opened 2026-08-31
fix(codex): preserve known GPT image input capabilities

Brisbanehuang:
## Summary

- Preserve `text` + `image` for known multimodal GPT models when an OpenAI-compatible endpoint omits `input_modalities`.
- Keep explicit provider and synchronized capability metadata authoritative, so explicitly text-only models remain text-only.
- Apply the same fallback to standard OpenAI model lists, native Codex manifests, and locally generated group catalogs.

## Problem

Codex treats `input_modalities` in the model catalog as an authoritative capability contract. In v0.1.184, compatible endpoints without modality metadata receive synthesized text-only descriptors, causing Codex to reject valid image attachments before the request reaches the upstream.

Unknown model IDs continue to fail closed to text-only.

## Validation

- `go vet ./internal/service ./internal/handler`


Brisbanehuang:

PR #6455, opened 2026-09-01
fix: keep release version metadata consistent across build paths

sunshinebird:
## Summary

Release tags currently point to commits whose tracked
`backend/cmd/server/VERSION` still contains the previous release number.
The release workflow fixes this temporarily for official artifacts, but
source builds from GitHub archives have no `.git` directory and therefore
fall back to the stale file.

This can make a Docker/source build of `v0.1.185` report `0.1.184` in the
web UI even though the release artifact itself is correct.

## Changes

- add an `export-subst` version marker so GitHub/git archive source snapshots
  carry the exact release tag without requiring `.git`
- make `resolve-version.sh` accept only exact release versions from archive
  metadata and exact Git tags, while safely falling back for development
  snapshots
- inject `main.Version` directly from `.Version` in both GoReleaser configs
- add shell regression coverage for archive, tagged-checkout, decorated-tag,
  and fallback cases
- document the version resolution behavior in the EN/CN/JA source-build docs

## Why this fixes the issue

The existing tag-aware resolver works when `.git` is available. Docker build
contexts and GitHub source archives intentionally omit `.git`, so they could
still use the previous value from `VERSION`. The archive marker closes that
gap without rewriting historical tags.

## Validation

Passed:

- `./backend/scripts/test-resolve-version.sh`
- `sh -n backend/scripts/resolve-version.sh`
- `sh -n backend/scripts/test-resolve-version.sh`
- YAML parsing for the modified workflow and GoReleaser configs
- `git diff --check`
- real `git archive` verification for an exact release tag

Note: the local `go test ./internal/service/...` run reached an existing,
unrelated failure in
`TestRecordCyberPolicyEvent_RuntimeSnapshotRefreshFailureKeepsStaleScope`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,3 +20,8 @@ backend/migrations/*.sql text eol=lf
 
 # Dockerfile
 Dockerfile text eol=lf
+
+# Preserve the exact release tag in GitHub-generated source archives. The
+# resolver ignores the placeholder in normal checkouts and uses the expanded
+# value when a release archive is built without a .git directory.
+backend/cmd/server/VERSION_TAG export-subst
--- a/.github/workflows/backend-ci.yml
+++ b/.github/workflows/backend-ci.yml
@@ -15,6 +15,7 @@ jobs:
       - name: Check deployment scripts
         run: |
           /bin/bash -n deploy/apple-container.sh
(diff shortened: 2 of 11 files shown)


sunshinebird:

PR #6533, opened 2026-09-03
fix(accounts): preserve custom upstream URL across protocol switches

xtianowner:
## Problem
For CN-provider API-key accounts, changing `api_protocol` in the edit modal can replace a user-entered custom `base_url` with the platform default URL. This is especially disruptive for relay endpoints: saving after a protocol switch silently changes the upstream target.

## Change
- Preserve the current non-empty upstream URL when switching between fixed protocols.
- When entering adaptive mode, carry the current URL into the legacy protocol slot and the adaptive Chat Completions fallback when that slot only contains the generated default.
- When leaving adaptive mode, prefer an explicitly configured non-default protocol endpoint; otherwise keep the current URL instead of silently falling back to the provider default.
- Keep existing preset selection behavior unchanged: an explicit preset click remains an intentional URL change.

## Tests
- Added regression coverage for fixed-to-fixed protocol switches.
- Updated adaptive-switch coverage for custom Anthropic and Responses relays.
- `pnpm test:run` (246 files, 1769 tests)
- `pnpm typecheck`
- `pnpm lint:check`


xtianowner:

PR #6568, opened 2026-09-04
fix: adapt Anthropic reasoning fields to model capabilities

ttxslol:
## Summary

- Adapt OpenAI Responses \
easoning.effort\ to the selected Anthropic model capability.
- Use \	hinking.type=adaptive\ with \output_config.effort\ for models listed in the Claude effort catalog.
- Use bounded manual thinking for model families that do not accept \output_config.effort\, including Claude Haiku 4.5.
- Omit optional reasoning fields for unknown models instead of sending a likely-invalid request.
- Re-apply the reasoning mapping after account/model aliases are resolved, including native Anthropic and Antigravity compatibility paths.

## Problem

\ResponsesToAnthropicRequest\ previously emitted \output_config.effort\ for every model carrying \
easoning.effort\. Anthropic rejects this for models such as Claude Haiku 4.5 with:

\This model does not support the effort parameter.\

The conversion also happened before gateway model mapping, so deciding from the client-facing model alone could select the wrong request shape.

## Validation

- \go test ./internal/pkg/apicompat -count=1\
- \go test ./internal/service -run 'Test(OpenAI|Responses|Anthropic|Antigravity)' -count=1\
- \go test -tags=unit ./...\

The change was also validated against a real Sub2API deployment using Claude Haiku 4.5.

### DIFF
--- a/backend/internal/pkg/apicompat/responses_to_anthropic_reasoning_test.go
+++ b/backend/internal/pkg/apicompat/responses_to_anthropic_reasoning_test.go
@@ -1,0 +1,107 @@
+package apicompat
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestResponsesToAnthropic_ReasoningUsesModelCapabilities(t *testing.T) {
+	maxTokens := 20000
+	tests := []struct {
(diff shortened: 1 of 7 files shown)


ttxslol:

PR #6572, opened 2026-09-04
feat(openai): complete GPT-6 Astra support

lyy0709:
## Summary

Complete GPT-6 Astra support on top of current upstream main, including the already-merged #6620 and #6628 changes. The branch merges upstream commit `0d9e7e1530855c93d4d00fc87406960d4a7fefd5`; it does not resubmit their cache gate, capability-sync implementation, or admin sync UI.

Remaining changes:

- Expose the public `gpt-6` alias, normalize it to `gpt-6-astra`, and add OpenCode configuration plus the complete Astra pricing/fallback coverage from the original PR.
- Give Astra its own 1,050,000-token context window rather than the GPT-5.6 fallback, preserve `max` in both forwarded requests and recorded usage, and correct the model catalog release timestamp.
- Persist and replay only five Codex tool-capability fields: `supports_search_tool`, `apply_patch_tool_type`, `comp_hash`, `tool_mode`, and `use_responses_lite`. Preserve explicit upstream `false` and `null`; use live declarations before snapshots, and snapshots before verified defaults. No prompts or arbitrary metadata are copied.
- Resolve capability metadata through actual model-mapping targets while retaining public alias names. Multiple accounts advertise only shared capabilities. Keep the existing Responses-to-Chat tool-search bridge behavior; unknown native third-party routes do not inherit official Astra capabilities.
- Use Astra fallback tool values checked with `codex-cli 0.153.4 debug models --bundled`. Keep the upstream API-key Responses Lite guard, including mapped aliases and group-generated catalogs; OAuth retains its own protocol behavior.
- Extend #6628's partial snapshot persistence so still-listed or explicitly mapped models keep their previous metadata when the new entry is incomplete. Remove metadata only when an authoritative live list and configured targets both omit it; a 404/405 model-list fallback retains previous entries. Do not overwrite fresh explicit capability values with older ones.
- Recover Messages continuation when HTTP 400 reports `previous_response_id is not available for this user` or `previous_response_id requires an OpenAI API-key account for HTTP requests`: replay the retained history and disable unsupported continuation for that session, without adding retry loops or treating auth failures as capability errors.

Relates to #6622, #6612, #6608, #6611, #6582, #6509 and #6524. Complements the now-merged #6620 and #6628 rather than replacing them.

Official model reference: https://developers.openai.com/api/docs/models/gpt-6-astra

## Verification

Passed on this upstream-compatible branch:

```text
cd backend
go test -tags=unit ./internal/pkg/openai ./internal/service -run 'Astra|GPT6|CodexModel|ConfiguredCodex|UpstreamModel|ModelsDev|CompatModel|PromptCache|Continuation|MaxCapable' -count=1
go test -tags=unit ./internal/handler -run 'CodexModels|GatewayModels|Compact' -count=1
go test ./internal/pkg/apicompat -run 'ToolSearch' -count=1
go vet ./internal/service ./internal/handler
go build ./...

cd frontend
pnpm exec vitest run src/components/account/__tests__/ModelWhitelistSelector.spec.ts src/composables/__tests__/useModelWhitelist.spec.ts src/components/keys/__tests__/UseKeyModal.spec.ts
pnpm run build

git diff upstream/main --check
```

Frontend: 3 files / 41 tests passed. Existing Browserslist freshness and bundle-size warnings remain; build succeeds.

## Scope and limitations

- No Sol price reduction or changes to existing pricing URLs/bundled price JSON.
- No additional `none -> low` mapping implementation. The implementation already merged by upstream in #6626 is left untouched and is not part of this PR's diff.
- No changes to account access, authentication policy, production configuration, or remote-compaction routing.
- No live paid-model calls or production deployment. Tests validate request/response handling and usage persistence with controlled upstream fixtures, not real-account availability or cache-hit rates.
- No database migration. Existing model-capability snapshots remain readable; syncing an account can populate the new optional capability fields.


kopang0730:
Confirmed the model-discovery fix independently on a self-hosted v0.2.0-based deployment.

Before adding `gpt-6-astra` to `openai.DefaultModels`, a direct `/v1/responses` request completed successfully with `response.model = gpt-6-astra`, while `/v1/models` omitted it even when the group's custom model list included it. `GetAvailableModels()` falls back to the default catalog for passthrough/unmapped accounts, and `filterModelsByCustomList()` intersects that catalog with the group's selections.

The `gpt-6-astra` catalog entry in this PR matches the minimal fix we deployed. After that change, the same API key discovered Astra and a streamed Astra request completed with `OK`. Existing Sol requests and GPT-5.4 Mini Chat Completions tool calls with streamed usage also passed.

We added handler regression coverage for:
- unmapped-account default discovery;
- passthrough discovery with a stale explicit mapping;
- a custom group list selecting Astra (including selection order);
- a custom group list excluding Astra;
- a restricted, non-passthrough mapping continuing to exclude Astra.

The first three cases failed before the catalog addition and passed afterward. Keeping `gpt-5.6-sol` first, as this PR does, also preserves the existing admin account-test default contract. These cases may be useful additions to `backend/internal/handler/gateway_models_test.go`.

This validation covers the catalog/discovery fix on our v0.2.0-based branch; the broader alias, reasoning, and billing changes in this PR were outside our verification scope. Sharing the evidence here to keep the overlapping discovery fix in one PR.

PR #6601, opened 2026-09-04
fix(usage): keep token and cost tooltips inside the viewport

viewer12:
## Summary

On the usage records page (user and admin share `UsageTable.vue`), tapping the token / cost `(i)` icon always opens the breakdown popover to the **right** of the trigger. In the mobile card layout that icon sits on the right edge, so the popover is clipped by the viewport and the 费用明细 / Token 明细 cannot be read.

This keeps the existing desktop right-side placement when there is room, and otherwise flips or clamps the teleported popover so the full breakdown stays on screen.

## Changes

- Add `getAnchoredTooltipPosition()` in `frontend/src/utils/floatingPanel.ts`
  - Prefer right, then left, then bottom / top
  - Clamp into the viewport
  - Keep the arrow aligned with the trigger after clamping
- Wire both usage tooltips through that helper and re-measure after render
- Cap popover width at `min(20rem, 100vw - 16px)` so long cost rows wrap instead of overflowing
- Accept click as well as hover so mobile taps open the popover
- Unit tests for the positioner and for 390px / desktop `UsageTable` behavior

## Test plan

- [x] `pnpm exec vitest run src/utils/__tests__/floatingPanel.spec.ts src/components/admin/usage/__tests__/UsageTable.spec.ts`
- [ ] Mobile (~390px): 使用记录 card → tap TOKEN `(i)` → full Token 明细 visible
- [ ] Same card → tap 费用 `(i)` → full 费用明细 visible
- [ ] Desktop table: tooltips still open to the right of the icon when there is space
- [ ] Near the bottom of the viewport: popover flips above instead of overflowing

### DIFF
--- a/frontend/src/components/admin/usage/UsageTable.vue
+++ b/frontend/src/components/admin/usage/UsageTable.vue
@@ -195,8 +195,10 @@
             <!-- Token Detail Tooltip -->
             <div
               class="group relative"
+              data-testid="usage-token-tooltip-trigger"
               @mouseenter="showTokenTooltip($event, row)"
               @mouseleave="hideTokenTooltip"
+              @click.stop="showTokenTooltip($event, row)"
             >
               <div class="flex h-4 w-4 cursor-help items-center justify-center rounded-full bg-gray-100 transition-colors group-hover:bg-blue-100 dark:bg-gray-700 dark:group-hover:bg-blue-900/50">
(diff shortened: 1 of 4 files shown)


viewer12:
Closing this revision to address tooltip lifecycle, short-viewport overflow, and interaction coverage before resubmitting a complete fix.

PR #6603, opened 2026-09-04
fix(usage): keep mobile detail tooltips in viewport

viewer12:
## Summary
- Flip the Token and cost detail tooltips to the left when the right side cannot fit them.
- Cap tooltip width to the available viewport space so the details remain readable on mobile.
- Preserve the existing right-side placement when enough room is available.

## Scope
This change only adjusts horizontal positioning and width for the two usage detail tooltips. Their interaction and vertical positioning are unchanged.

## Test plan
- [x] 390px viewport: Token tooltip opens left and remains inside the viewport.
- [x] 390px viewport: cost tooltip opens left and remains inside the viewport.
- [x] Desktop viewport: existing right-side placement is preserved.
- [x] Full frontend test suite (251 files, 1827 tests).
- [x] Type check.
- [x] Production build.

### DIFF
--- a/frontend/src/components/admin/usage/UsageTable.vue
+++ b/frontend/src/components/admin/usage/UsageTable.vue
@@ -195,6 +195,7 @@
             <!-- Token Detail Tooltip -->
             <div
               class="group relative"
+              data-testid="usage-token-tooltip-trigger"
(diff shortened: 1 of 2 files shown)


viewer12:

PR #6604, opened 2026-09-04
fix: unblock ops dashboard and correct RMB plaza pricing

CatStack-pixe:
## Summary\n- use configured channel pricing as the official-price source for deepseek-018 and glm-018\n- render those groups with the RMB symbol while preserving the per-million-token unit\n- load the ops dashboard through bounded parallel endpoints instead of the slow aggregate snapshot\n- keep optional payment configuration from blocking admin settings readiness\n\n## Verification\n- pnpm run build\n- pnpm run typecheck\n- pnpm run lint:check\n- pnpm exec vitest run src/stores/__tests__/adminSettings.spec.ts\n- backend unit tests require Go 1.27 in CI\n


CatStack-pixe:
Replaced by clean PR #6605 based on upstream main.

PR #6619, opened 2026-09-05
fix: normalize and safely forward Responses Lite requests

lizhaoxinlizhaoxin2-lab:
## Problem
OpenAI Codex Responses Lite requests returned 400 unsupported_value because the gateway forwarded X-OpenAI-Internal-Codex-Responses-Lite: true on ordinary /v1/responses requests without guaranteeing reasoning.context = "all_turns" upstream.

## Fix
- Normalize the final body before the WebSocket-to-HTTP bridge adds the Lite header.
- Stop forwarding the internal Lite header on ordinary HTTP passthrough requests where the Lite payload contract cannot be guaranteed.
- Preserve the Lite normalization path for explicitly handled Responses Lite requests.

## Production validation
Reproduced on Sub2API v0.2.0-goforai-r2-alpha with upstream errors identifying param=reasoning.context, then deployed as sub2api:0.2.0-goforai-r4-lite-header-fix. The production health endpoint is healthy and the new container has no matching Lite errors after startup.

## Testing
- git diff --check passes.
- Production Docker build completed successfully.
- The local Windows workspace does not have the Go toolchain, so the Go test suite was not run locally.

### DIFF
--- a/backend/internal/service/openai_gateway_service.go
+++ b/backend/internal/service/openai_gateway_service.go
@@ -83,7 +83,7 @@ var openaiAllowedHeaders = map[string]bool{
 	"x-codex-turn-state":      true,
 	"x-codex-turn-metadata":   true,
 	"x-codex-window-id":       true,
-	responsesLiteHeaderKey:    true,
+	responsesLiteHeaderKey:    false,
 }
 
 // OpenAI passthrough allowed headers whitelist.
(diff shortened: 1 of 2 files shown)


Wei-Shaw:
recheck

lizhaoxinlizhaoxin2-lab:

PR #6629, opened 2026-09-05
fix(apicompat): preserve function arguments from done events

superdav42:
## Problem

Some OpenAI Responses streams add a function-call item with empty arguments, then provide the complete JSON only in `response.function_call_arguments.done` without argument delta events. The Responses-to-Chat bridge ignored done events, so Chat Completions clients received `arguments: ""` in both streaming and buffered responses.

## Solution

- Track emitted function arguments by Responses output index.
- Emit only the missing suffix from argument-done events, avoiding duplication when deltas were already delivered.
- Treat argument-done payloads as authoritative in the buffered accumulator.
- Fill empty terminal function-call arguments from accumulated stream data.
- Apply the same fallback to custom-tool input-done events.

## Verification

- `go test ./internal/pkg/apicompat`
- `go test -tags=unit ./internal/pkg/apicompat`
- `make -C backend build`
- Live OAuth regression: required integer argument `id` is preserved as `{"id":1}` for streaming and non-streaming Chat Completions.

---
[aidevops.sh](https://aidevops.sh) v3.32.306 plugin for [OpenCode](https://opencode.ai) v1.18.29 with gpt-5.6-sol spent 2d 6h and 5,921,965 tokens on this with the user in an interactive session.

### DIFF
--- a/backend/internal/pkg/apicompat/chatcompletions_responses_test.go
+++ b/backend/internal/pkg/apicompat/chatcompletions_responses_test.go
@@ -1621,6 +1621,108 @@ func TestBufferedResponseAccumulator_ToolCalls(t *testing.T) {
 	assert.Equal(t, `{"city":"NYC"}`, output[0].Arguments)
 }
 
+func TestResponsesEventToChatChunks_FunctionArgumentsDoneWithoutDeltas(t *testing.T) {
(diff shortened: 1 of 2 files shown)


Wei-Shaw:
recheck

superdav42:

PR #6633, opened 2026-09-05
feat(openai): replay encrypted reasoning for Chat Completions clients on the Codex path

### LINKED ISSUE #6632: Chat Completions clients lose reasoning items between tool turns on the Codex OAuth path (quality regression, no error)
## Summary

Chat Completions clients that reach Codex through `/v1/chat/completions` (WorkBuddy in our case, but any agent client speaking Chat Completions) get noticeably worse multi-step results than native Codex clients on the same account pool. Every request returns 200 and nothing is logged, so the regression is invisible on the gateway side.

## Root cause

`ForwardAsChatCompletions` converts the Chat history to Responses input via `apicompat.ChatCompletionsToResponses`. An assistant message with `tool_calls` becomes `function_call` items, but there is no `reasoning` item in front of them: Chat Completions has no field that can carry a Responses reasoning item, so the client cannot return it, and the gateway does not restore it.

Under `store=false` (forced on the OAuth path) the upstream expects the reasoning items that preceded a function call to be replayed with `encrypted_content` when the `function_call_output` comes back. #3588 keeps exactly that for native Codex clients. For Chat clients the item is simply gone, so the model restarts every tool round without its prior reasoning state.

A second, smaller issue: `chatAssistantToResponses` wraps the client's `reasoning_content` in `<thinking>…</thinking>` and emits it as assistant `output_text`, so the model sees its own reasoning summary as text it previously "said".

## Observed impact

Same account pool, `gpt-5.6`, 3-day window:

| client | inbound | cache hit | avg fresh input tokens |
|---|---|---|---|
| Codex Desktop | /v1/responses | 95%+ | 2k–4k |
| WorkBuddy | /v1/chat/completions | 84.6% | ~10k |

Tool-continuation requests were the ones paying: each carried the whole history as fresh input with no reasoning context.

## Proposal

Gateway-side replay, symmetric to the Responses→Chat reasoning cache added in #5729 and to the Anthropic `thinking.signature` ↔ `reasoning.encrypted_content` round-trip:

1. While translating the Responses stream to Chat chunks, bind each completed `reasoning` item (with `encrypted_content`) to the next completed `function_call` and cache it under that `call_id`, tagged with the producing account.
2. When the next Chat request replays `assistant.tool_calls`, look the ids up and insert the cached items as `reasoning` input items right before the matching `function_call`; drop the `<thinking>` echo when real reasoning is restored.
3. Never inject records written by another account (ciphertext is account-bound); retry once without replay if the upstream still answers `invalid_encrypted_content`.

A miss leaves the request unchanged. We have this running in production; after the change tool-continuation requests hit 95.6–99.2% cache and the fresh input dropped to a few hundred tokens. PR follows.

wolfhead:
## Summary

Chat Completions clients that reach Codex through `/v1/chat/completions` lose the Responses `reasoning` items between tool turns. Chat Completions has no field for them, so when such a client replays `assistant.tool_calls` the gateway forwards the `function_call` items without the reasoning that produced them. Under `store=false` the upstream expects those items (with `encrypted_content`) to be replayed in front of the `function_call` (the same contract #3588 preserves for native Codex clients); without them the model restarts every tool round from scratch. Every request still returns 200, so the only symptoms are degraded multi-step agent quality and a lower prompt-cache hit rate.

The bridge also echoed the plaintext `reasoning_content` back to the upstream as a `<thinking>` block inside assistant `output_text`, so the model saw its own reasoning summary as spoken output.

This PR closes the gap on the gateway side, mirroring the existing Responses→Chat reasoning cache (#5729) and the Anthropic `thinking.signature` round-trip:

- **Response direction** (`chatReasoningReplayRecorder`): while the Responses stream is translated to Chat chunks, completed `reasoning` items are bound to the next completed `function_call` / `custom_tool_call` and stored in the shared reasoning cache under the tool `call_id`, tagged with the account that produced them. Later calls of the same parallel batch get a marker record pointing at the batch owner. The buffered (`stream=false`) path is covered too.
- **Request direction** (`apicompat.ChatCompletionsToResponsesWithOptions`): on the Codex OAuth path the converter looks up `tool_calls[].id` and inserts the cached items as `reasoning` input items (with `encrypted_content` and an explicit empty `summary`) right before the matching `function_call`. When real reasoning is restored the `<thinking>` echo is dropped. Records written by another account are never injected (ciphertext is account-bound).
- **Recovery**: if the upstream still rejects the ciphertext (`invalid_encrypted_content`), the request is retried once without replay.

A cache miss leaves the request byte-for-byte as before. Only `account.UsesOpenAICodexProtocol()` accounts are affected; the other six callers of `ChatCompletionsToResponses` keep the nil-options path. No config, schema or API changes; the cache reuses `GatewayCache.Set/GetReasoningContent` with a `cc_tool_call:` key prefix and the existing 7-day TTL.

## Production data

Deployed on our gateway (upstream v0.1.183 + this patch) with a Chat Completions agent client (WorkBuddy) driving Codex `gpt-5.6` on a chatgpt.com OAuth account:

| | before | after |
|---|---|---|
| tool-continuation requests: fresh input tokens | ~10k avg | 250–1300 |
| tool-continuation requests: cache hit | 84.6% | 95.6–99.2% |
| avg latency (tool-heavy task) | 23.2 s | 7.0 s |
| avg TTFT | 3.7 s | 1.0 s |
| `invalid_encrypted_content` retries / account-mismatch skips | – | 0 in ~100 requests, context up to 125k tokens |

Each encrypted reasoning item is ~2.2 KB (max 3.8 KB seen); Redis growth was negligible.

## Logging

Per-request summaries are `Debug` (`reasoning replay recorded` / `reasoning replay lookup` with hit/miss/covered/injected counts). `Info` only when a cached record from another account is skipped, `Warn` on cache read/write failures and on the rejected-ciphertext retry.

## Tests

- `apicompat`: hit inserts items before the call and drops `<thinking>`; partial hit keeps assistant text; miss/blank keeps the legacy shape; blank ids skipped.
- `service`: recorder binding (stream + buffered, parallel batches with sibling markers), account-scoped lookup, round trip, enable gating, error matcher.
- `service` end-to-end through `ForwardAsChatCompletions` with a fake Codex stream: two-turn replay, miss, rejected-cipher retry.
- `go test ./internal/pkg/apicompat/`, `go test -tags unit ./internal/service/`, `go vet`, golangci-lint v2.13 (CI config): clean.

Closes #6632

### DIFF
--- a/backend/internal/pkg/apicompat/chatcompletions_to_responses.go
+++ b/backend/internal/pkg/apicompat/chatcompletions_to_responses.go
@@ -11,12 +11,35 @@ type chatMessageContent struct {
 	Parts []ChatContentPart
 }
 
+// ChatToResponsesOptions carries optional hooks for
+// ChatCompletionsToResponsesWithOptions. All fields are optional; a nil
+// *ChatToResponsesOptions behaves exactly like ChatCompletionsToResponses.
+type ChatToResponsesOptions struct {
+	// ReasoningByToolCallID returns the encrypted reasoning payloads that the
+	// upstream emitted immediately before the function_call with this call_id
+	// in an earlier turn. Chat Completions has no field for carrying Responses
+	// reasoning items, so a Chat client replaying tool history cannot return
+	// them on its own; the gateway caches them under the call_id when the
+	// tool call is streamed to the client and restores them here. Each entry
+	// becomes a `reasoning` input item (with encrypted_content) placed right
+	// before the corresponding function_call item, which is the shape the
+	// Responses API expects for stateless multi-turn reasoning. Return nil on
+	// a miss. A nil hook keeps the original behavior.
+	ReasoningByToolCallID func(callID string) []string
+}
+
 // ChatCompletionsToResponses converts a Chat Completions request into a
 // Responses API request. The upstream always streams, so Stream is forced to
 // true. store is always false and reasoning.encrypted_content is always
 // included so that the response translator has full context.
 func ChatCompletionsToResponses(req *ChatCompletionsRequest) (*ResponsesRequest, error) {
-	input, err := convertChatMessagesToResponsesInput(req.Messages)
+	return ChatCompletionsToResponsesWithOptions(req, nil)
+}
+
+// ChatCompletionsToResponsesWithOptions is ChatCompletionsToResponses with
+// optional hooks (see ChatToResponsesOptions).
+func ChatCompletionsToResponsesWithOptions(req *ChatCompletionsRequest, opts *ChatToResponsesOptions) (*ResponsesRequest, error) {
+	input, err := convertChatMessagesToResponsesInput(req.Messages, opts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


wolfhead:

PR #6575, opened 2026-09-04
feat(antigravity): support Gemini 3.7 and 3.8 Flash models

doublepi123:
## What

Registers **Gemini 3.7 Flash** (GA 2026-08-13) and **Gemini 3.8 Flash** (GA 2026-09-02) with their Antigravity thinking tiers, following the exact pattern already used for Gemini 3.6 Flash.

Closes #6523.

## Why

Neither model is currently routable. Both are absent from `DefaultAntigravityModelMapping`, so requests fall back to the default mapping and the ids are rejected. They are also absent from the pricing surfaces, so even once routed, `GetModelPricing` fails closed with `ErrModelPricingUnavailable`.

## Changes

| File | Change |
|---|---|
| `domain/constants.go` | `DefaultAntigravityModelMapping`: bare id + `-high` / `-low` / `-medium` / `-tiered` for both versions |
| `pkg/antigravity/claude_types.go` | `geminiModels` catalog entries surfaced via `/antigravity/v1/models` |
| `service/account.go` | `ensureAntigravityDefaultPassthroughs` so accounts **with** a custom `model_mapping` still resolve the new ids |
| `service/billing_service.go` | fallback price cards + `getFallbackPricing` substring branches (also matching the `gemini-3-7-flash` / `gemini-3-8-flash` spellings, mirroring 3.6) |
| `service/pricing_service.go` | `normalizeGeminiThinkingTierAlias` generalized from a single `const baseModel` to a base-model list, so each family's tiers share its base price card |

## Pricing

Both models use Google's published rates: **$0.75 input / $3.75 output / $0.075 cached input per MTok**.

These are introductory rates that **expire 2026-12-31 and double to $1.50 / $7.50 / $0.15 on 2027-01-01** ([pricing](https://ai.google.dev/gemini-api/docs/pricing), [3.7 launch post](https://blog.google/innovation-and-ai/models-and-research/gemini-models/introducing-gemini-3-7-flash/)). Each constant carries an inline note so the expiry is visible at the point of change. Happy to drop the notes or handle the rollover differently if you prefer.

Neither model has a long-context (>200K) price tier — the published tables are a single flat rate.

## Notes on the tier suffixes

[...]

PR #6637, opened 2026-09-05
fix(billing): resolve Gemini 3.7/3.8 Flash thinking-tier prices

varrge:
Requests such as `gemini-3.7-flash-medium` fail price lookup even when the pricing catalog contains `gemini-3.7-flash`. The same applies to Gemini 3.8 Flash. `GatewayService.RecordUsage` then persists zero total/actual cost and skips balance deduction despite nonzero token usage, because thinking-tier normalization currently only covers Gemini 3.6 Flash.

This change normalizes the known `-high`, `-low`, `-medium`, and `-tiered` aliases for Gemini 3.7/3.8 Flash to their corresponding base catalog entries. Exact tier-specific catalog prices retain precedence. It uses the existing catalog rates and keeps model routing and default price cards unchanged.

Regression coverage includes catalog lookup, tier-specific pricing precedence, and the complete `RecordUsage` → pricing resolver → billing → usage log/balance deduction path with input, output, cached input, and a group multiplier. For the test fixture, a previously zero-cost request now records a total cost of `$0.02007585` and an actual cost of `$0.0030113775` at a `0.15` multiplier.

Related: #6575 includes the same normalization fix as part of broader model-support work. This PR is a focused billing fix with additional usage-recording and pricing-precedence regression coverage; either approach can resolve the overlapping normalization issue.

Validation:

- `env -u OPENAI_API_KEY go test -tags=unit ./...` — passed.
- `env -u OPENAI_API_KEY go test -tags=integration ./...` — passed, including the PostgreSQL/Redis integration suite.
- `golangci-lint run --timeout=30m ./...` — passed with 0 issues (v2.13.2).
- `gofmt` and `git diff --check` — clean.

Tests ran with Go 1.27.0. `OPENAI_API_KEY` was unset to keep the optional live OpenAI comparison test disabled, matching the default CI environment; the initial run with an inherited key failed that unrelated test with a network EOF.

### DIFF
--- a/backend/internal/service/gateway_record_usage_test.go
+++ b/backend/internal/service/gateway_record_usage_test.go
@@ -194,6 +194,42 @@ func TestGatewayServiceRecordUsage_PreservesRequestedAndUpstreamModels(t *testin
 	require.Equal(t, mappedModel, *usageRepo.lastLog.UpstreamModel)
 }
 
+func TestGatewayServiceRecordUsage_GeminiFlashThinkingTierUsesCatalogPrice(t *testing.T) {
+	for _, baseModel := range []string{"gemini-3.7-flash", "gemini-3.8-flash"} {
+		t.Run(baseModel, func(t *testing.T) {
+			usageRepo := &openAIRecordUsageLogRepoStub{inserted: true}
+			userRepo := &openAIRecordUsageUserRepoStub{}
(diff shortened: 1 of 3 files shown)


varrge:

varrge:

PR #6640, opened 2026-09-05
fix(claude): align billing fingerprint with outbound User-Agent

cyhhao:
## Problem

Claude billing attribution can disagree with the final outbound request in two ways:

1. `syncBillingHeaderVersion` changes the semver but preserves the old three-character fingerprint suffix. That suffix is derived from both the first user message and the CLI version, so keeping it after changing the version is inconsistent with `computeClaudeCodeFingerprint`.
2. OAuth mimicry applies the cached fingerprint first, then forces the built-in User-Agent. Billing is currently synchronized to the cached version, even though it can be overwritten later. When fingerprint unification is disabled or the identity service is absent, the synchronization is skipped altogether despite mimicry still forcing a User-Agent.

Both `/v1/messages` and `/v1/messages/count_tokens` have this ordering issue.

## Fix

- Resolve the effective billing User-Agent using the same OAuth mimicry precedence as the outgoing headers.
- Recompute the recognized message-derived suffix when synchronizing the billing version.
- Keep non-mimic fingerprint behavior and API-key passthrough unchanged.
- Do not change the built-in CLI version or add a dependency on #6481. This is an independent consistency fix on current main.

## Tests

Regression coverage checks both request builders with cached fingerprints, missing identity services, disabled fingerprint unification, and non-mimic OAuth requests. It verifies the final User-Agent, the full `cc_version` value, and the actual HTTP request body. Additional helper coverage checks suffix recomputation and idempotence.

The regression tests fail on the unmodified implementation, including all eight request-builder cases. With the fix applied, the following focused suite passes:

```sh
cd backend
go test -tags=unit ./internal/service -run 'Test(SyncBillingHeaderVersion|BuildOAuthRequest_BillingMatchesWireUserAgent|BuildUpstreamRequest|BuildCountTokensRequest|GatewayService_Anthropic)' -count=1
```

Formatting and `git diff --check` also pass.

### DIFF
--- a/backend/internal/service/gateway_billing_header.go
+++ b/backend/internal/service/gateway_billing_header.go
@@ -5,16 +5,30 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/Wei-Shaw/sub2api/internal/pkg/claude"
 	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 )
 
-// ccVersionInBillingRe matches the semver part of cc_version (X.Y.Z), preserving
-// the trailing message-derived suffix (e.g. ".c02") if present.
+// ccVersionInBillingRe matches the semver part of cc_version (X.Y.Z).
 var ccVersionInBillingRe = regexp.MustCompile(`cc_version=\d+\.\d+\.\d+`)
+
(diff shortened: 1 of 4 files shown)


cyhhao:
@codex review

cyhhao:
@codex review

PR #6661, opened 2026-09-05
fix(grok): release abandoned media slots and preserve video ownership

whhphd:
## Problem

`GET /v1/videos/:request_id` resolves the task's owner, then calls the generic scheduler. If sticky routing escapes to another account, the scheduler can return `Acquired=true`, but `handleGrokMedia` returns 404 for the owner mismatch before registering account cleanup. The user slot is released while the account slot remains until expiry. Repeated polling can exhaust account capacity.

Generation requests have the same release gap when `ensureGrokMediaAccountEligibility` rejects an already acquired selection and the handler retries or exhausts its switch budget.

Confirmed on current `main` / v0.2.1 (`ab99d56e9626e6cd731592dae8553c9758a0efa2`). This PR is based directly on that commit and contains only this fix and regression tests.

## Changes

- Own scheduler-acquired slots immediately, before error, owner, or eligibility checks. Release rejected attempts before selecting again, and keep cancellation/forwarding cleanup once-only.
- Select only the authenticated video task owner, preserving existing group/platform/privacy/schedulability checks. A full owner uses the existing bounded account wait queue; queue saturation or timeout returns 429. Missing or unavailable ownership remains 404. Lookups never forward to another account.
- Preserve the owner binding and its TTL during lookup selection and queue admission. Generation failover, user/API key isolation, and completion billing deduplication remain covered.

Related: #4854 also contains owner-specific video lookup and mismatch cleanup in its current head. It is still open and covers a broader set of changes. This PR addresses the current main branch independently, including the eligibility rejection release gap; it does not introduce the single-slot Grok OAuth policy or durable task persistence from that PR.

## Reproduction

On a checkout of `ab99d56e9`, add only `backend/internal/handler/grok_media_slots_test.go` from this PR and run:

```sh
cd backend
go test -tags=unit ./internal/handler -run 'TestGrokMediaLookupSlotLifecycle/mismatch_acquired|TestGrokMediaEligibilityReleasesBeforeSwitch/exhausted' -count=1
```

Both cases fail on the base commit with unreleased account slots. The same tests pass with this fix. The handler tests also check release counts, repeated polling, WaitPlan admission/rejection/cancellation, eligibility failover, upstream errors, panics, user/API key/group/task isolation, and one-shot completion billing.

## Validation

Passed on this upstream-based branch with Go 1.27.0 and golangci-lint v2.13, matching the repository CI:

```sh
cd backend
go test ./...
go test -tags=unit ./...
go test -tags=integration ./...
go test -race -tags=unit ./internal/handler ./internal/service -run 'TestGrokMedia|TestSelectGrokMedia|TestGrokVideoSticky' -count=1
go build ./cmd/server
golangci-lint run --timeout=30m ./...
```

The integration suite used isolated PostgreSQL/Redis test containers. Lint reported `0 issues`; `gofmt -l` on changed Go files and `git diff --check` were clean.

The equivalent fix is deployed in a downstream production environment. More than 900 video lookups were observed after deployment; the latest account-slot snapshot was 4, with none older than 10 minutes, versus hundreds before deployment. No new owner-mismatch or slot-release-failure events were observed, and no manual Redis slot clearing was performed. These observations support the regression tests but are not an exact measurement of the previous leak count.

No schema, dependency, frontend, configuration, or concurrency-limit changes are included.

### DIFF
--- a/backend/internal/handler/grok_media.go
+++ b/backend/internal/handler/grok_media.go
@@ -198,28 +198,55 @@ func (h *OpenAIGatewayHandler) handleGrokMedia(c *gin.Context, endpoint service.
 	}
 	routingStart := time.Now()
 	requiredCapability := grokMediaRequiredCapability(endpoint)
+	var accountReleaseFunc func()
+	releaseAccount := func() {
+		if accountReleaseFunc != nil {
+			accountReleaseFunc()
+			accountReleaseFunc = nil
+		}
+	}
+	defer releaseAccount()
 
 	for {
+		releaseAccount()
 		if failoverClientGone(c) {
 			return
 		}
-		selection, scheduleDecision, err := h.gatewayService.SelectAccountWithSchedulerForCapability(
-			requestCtx,
-			apiKey.GroupID,
-			"",
-			sessionHash,
-			routingModel,
-			failedAccountIDs,
-			service.OpenAIUpstreamTransportHTTPSSE,
-			requiredCapability,
-			false,
-			false,
-			false,
-			service.PlatformGrok,
-		)
+		var selection *service.AccountSelectionResult
+		var scheduleDecision service.OpenAIAccountScheduleDecision
+		if boundLookupAccountID > 0 {
+			selection, scheduleDecision, err = h.gatewayService.SelectGrokMediaVideoRequestAccount(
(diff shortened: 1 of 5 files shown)


whhphd:

PR #6680, opened 2026-09-05
fix(openai): retain default models in mixed account catalogs

cyhhao:
## Problem

An OpenAI group can contain an ordinary account with no `model_mapping` and a Spark shadow account with an explicit mapping. Once any mapping exists, the group catalog only contains mapped names: ordinary GPT models disappear from `/v1/models` and the generated Codex manifest even though the ordinary account can serve them.

The mapped OpenAI response path also uses the Claude model serializer, omitting the standard OpenAI `object`, `created`, and `owned_by` fields.

## Changes

- Supplement a nonempty mapped catalog with the existing OpenAI default model IDs when an eligible unmapped OpenAI account is present.
- Reuse that rule for ordinary model lists and generated Codex manifests, retaining aliases and account-provided metadata.
- Route OpenAI model-list responses through the existing OpenAI serializer.
- Preserve group custom-list filtering and ordering, cache behavior, ETags, media/automatic-mode filtering, and the all-unmapped discovery fallback. Fully mapped groups do not gain default models.

This changes catalog presentation only, not account mappings, routing, model support checks, or billing. It reuses the built-in catalog; it does not introduce upstream discovery or claim real-time model availability.

## Validation

- Focused handler and service tests passed for standard/Codex model lists, configured manifests, group filtering, caching, and passthrough fallback.
- Expanded model-related checks passed: `go test -tags=unit ./internal/handler ./internal/service -run '(Models|ModelList)' -count=1 -timeout=10m` (from `backend`).
- Regression coverage includes an unmapped OAuth parent with a Spark shadow, unmapped API-key accounts, mapped aliases, all-mapped and cross-platform exclusions, empty custom selections, OpenAI response fields, and Codex ETag/304 behavior.
- `git diff --check` passed.

### DIFF
--- a/backend/internal/handler/gateway_handler.go
+++ b/backend/internal/handler/gateway_handler.go
@@ -1138,7 +1138,7 @@ func (h *GatewayHandler) Models(c *gin.Context) {
 		availableModels := h.compositeAvailableModels(c.Request.Context(), groupID)
 		if apiKey != nil && apiKey.Group != nil && apiKey.Group.CustomModelsListEnabled() {
 			availableModels = filterModelsByCustomList(availableModels, defaultModelIDsForPlatform(service.PlatformComposite), apiKey.Group.ModelsListConfig.Models)
-			writeCustomModelsList(c, service.PlatformComposite, availableModels)
+			writeModelsList(c, service.PlatformComposite, availableModels)
 			return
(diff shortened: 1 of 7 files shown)


cyhhao:
@codex review

PR #6407, opened 2026-08-31
fix(openai): accept delegation bootstrap without call id

wucm667:
## 问题

桌面客户端委派在发起 HTTP Responses 委派的引导请求时可能不携带调用标识，网关会拒绝该合法请求，导致委派流程无法建立。

## 根因

现有校验将引导请求与后续委派调用使用同一套调用标识要求，未区分首次建立委派上下文的请求形态。

## 改动

允许委派引导请求在没有调用标识时进入既有处理流程，并补充覆盖该请求形态及相关边界条件的回归测试。

重复项预检查：未发现来自 `wucm667:fix/issue-6402-delegation-bootstrap` 的已有拉取请求。

## 验证

验证结果：后端单元测试套件通过；后端集成测试套件通过；`internal/handler` 完整测试通过；Go 静态检查 v2.13.0 报告 0 个问题；后端构建通过；前端代码检查和类型检查通过；前端 13 个关键文件共 167 项测试通过；Linux 部署安全、网关环境变量、运行时资源及 Caddy 检查通过。

漏洞扫描的数据库获取因 DNS 超时受阻，因此不声明该项通过。

Fixes #6402

PR #6687, opened 2026-09-06
fix(openai): preserve named standalone Responses inputs

Chenning-Tao:
Codex Desktop sends `create_thread` and `send_message_to_thread` inputs as named `function_call_output` items with no `call_id`. On the OAuth Responses WebSocket path, orphan-output cleanup silently removes these standalone inputs whenever `previous_response_id` is absent. A turn can then complete after seeing only environment/context updates, without receiving the delegation request.

Preserve named standalone function outputs in the shared sanitizer, keeping their native type, namespace, envelope and ordering. Outputs with a nonempty orphan `call_id`, unnamed outputs, and other output types continue to use the existing pairing rules. No user-message conversion or synthetic call ID is introduced.

Related to #6402, #6407 and #6539: the existing HTTP delegation normalization does not run on the WebSocket path. The shared WebSocket compatibility function is used for both initial and subsequent frames.

Validation:
- Regression coverage for both task tools, with and without history and `previous_response_id`, including idempotence.
- Negative coverage confirms that preserving a named standalone input does not preserve unrelated orphan results.
- Focused regression tests, `go test ./internal/service -count=1` and `go vet ./internal/service` passed.
- [CI on this exact patch](https://github.com/Chenning-Tao/sub2api-custom/actions/runs/34004247615) passed unit/integration tests, golangci-lint, frontend checks and shell checks.

All fixtures are synthetic and contain no production request data.

### DIFF
--- a/backend/internal/service/openai_responses_input_compat.go
+++ b/backend/internal/service/openai_responses_input_compat.go
@@ -7,7 +7,8 @@ import (
 const openAIResponsesInputTextMaxChars = 10000000
 
 // sanitizeOpenAIResponsesOrphanToolOutputs removes tool-output items that have
-// no matching call or item reference anywhere in the current input.
+// no matching call or item reference anywhere in the current input. Named
+// function outputs without a call ID are standalone inputs, not orphan results.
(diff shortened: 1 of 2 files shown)


Chenning-Tao:

PR #6688, opened 2026-09-06
fix(openai): honor model_routing for Codex manifests and scheduling

suixincrazy:
## Problem

An OpenAI group that keeps a failover account for a shared public alias serves a Codex model descriptor with no reasoning levels and text-only input. The client's reasoning-effort slider then has nothing to select and cannot be moved.

Reproduction: one OpenAI group, two accounts, both with an explicit `model_mapping` for the same public alias, pointing at different upstream models.

```
account A (primary):  {"my-alias": "my-alias"}
account B (failover): {"my-alias": "some-other-model"}
```

Account B exists because it cannot reach the aliased model natively, so it maps the alias onto the closest model it does have. That is the intended way to keep a standby for an alias, and the Anthropic side is already configured this way in practice.

Served descriptor:

```json
{
  "slug": "my-alias",
  "supported_reasoning_levels": [],
  "input_modalities": ["text"]
}
```

`default_reasoning_level` is absent entirely.

## Cause

Two accounts explicitly mapping one alias to different upstream models makes `codexExplicitModelTargetsConflictForPlatform` true. With no `upstream_model_metadata` snapshot to intersect, `groupCodexModelMetadata` short-circuits to `{reasoningConflict: true, inputModalitiesConflict: true}`.

That override empties the level list and forces `input_modalities` to `["text"]`. Two details make it unrecoverable downstream:

- `DefaultReasoningLevel` is `*string` with `omitempty`, so nil makes the key absent rather than null.
- `mergeMissingCodexModelFields` fills only missing-or-null fields, and an empty array counts as present, so `supported_reasoning_levels: []` is never repaired.

Failing closed is correct when the alias target is genuinely ambiguous. But a group that declares `model_routing` for the alias has already resolved that ambiguity: the operator has stated which accounts serve it, and account `Priority` states which one leads. Treating that as a conflict rejects a deliberate failover configuration.

## Change

**Manifest side.** When a routing rule names the alias, capability declaration is restricted to the routed accounts and the conflict override is skipped. Capabilities then fall back to the same by-name derivation already used for a single account with no metadata snapshot. Behavior is unchanged for groups with no rule, a disabled rule, or a rule naming a different alias.

**Scheduling side.** `model_routing` is stored and looked up platform-agnostically (`admin_group.go` assigns it with no platform check; `Group.GetRoutingAccountIDs` is a plain table lookup), but four gates in `gateway_scheduling.go` only consulted it for Anthropic. An OpenAI group's rules were therefore accepted, persisted, and shown in the admin UI while never affecting selection.

Fixing only the manifest would leave a split brain: the manifest would honor a rule that scheduling ignores. Both sides now share one predicate, and the allowed platform set is defined in exactly one place so adding a platform later is a one-line change.

Selection *within* a routing list is deliberately untouched. The list is a membership set, and the winner is still decided by `Priority` (lower first), then load rate, then `LastUsedAt`.

## Tests

- `TestCodexAliasFailoverMappingHonorsModelRouting` — the routed alias regains `low..max` levels, `medium` default, and image input; three negative cases (no rule, rule disabled, rule for another alias) must still fail closed.
- `TestModelRoutingAppliesToPlatform` — predicate table including composite groups and non-eligible target platforms.
- `TestGatewayService_SelectAccountForModelWithPlatform_RoutedOpenAIGroup` — a routed OpenAI account must win over a higher-priority unrouted one, which fails before this change.

Verified with `go build ./...`, `go vet -tags unit ./internal/service/`, and `go test -tags unit ./internal/service/` (411 tests matching `Codex|ModelRouting|Routed` pass). Five failures in the full package run (`TestContentModerationRuntimeSnapshotRefreshFailureKeepsStaleConfig`, four `TestPluginPackageInstaller*`) reproduce identically on unmodified `main` and are unrelated to this change.

### DIFF
--- a/backend/internal/service/gateway_multiplatform_test.go
+++ b/backend/internal/service/gateway_multiplatform_test.go
@@ -3515,3 +3515,74 @@ func TestGatewayService_ResolveGatewayGroup_DetectsFallbackCycle(t *testing.T) {
 	require.Nil(t, gotID)
 	require.Contains(t, err.Error(), "fallback group cycle")
 }
+
+func TestModelRoutingAppliesToPlatform(t *testing.T) {
+	for _, tc := range []struct {
+		name           string
+		targetPlatform string
+		groupPlatform  string
+		want           bool
+	}{
+		{"anthropic group", PlatformAnthropic, PlatformAnthropic, true},
+		{"openai group", PlatformOpenAI, PlatformOpenAI, true},
+		{"composite group resolved to anthropic", PlatformAnthropic, PlatformComposite, true},
+		{"composite group resolved to openai", PlatformOpenAI, PlatformComposite, true},
+		{"target platform outside the allowed set", PlatformGemini, PlatformGemini, false},
+		{"composite group resolved outside the allowed set", PlatformGemini, PlatformComposite, false},
+		{"group platform does not match target", PlatformOpenAI, PlatformAnthropic, false},
+		{"empty target platform", "", PlatformOpenAI, false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.want, modelRoutingAppliesToPlatform(tc.targetPlatform, tc.groupPlatform))
+		})
+	}
+}
+
+// Model routing used to be read only for requests resolved to Anthropic, so an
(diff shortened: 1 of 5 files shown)


suixincrazy:

PR #6702, opened 2026-09-06
fix(gateway): use JSON escaping for message cache promotion

a1gaoshanlaiyige:
When message cache rewriting promotes string content to a text block, `mustJSONString` uses `fmt.Sprintf("%q", s)`. This produces Go string literals, which can contain escapes that JSON does not support. For example, a message containing DEL (U+007F) becomes `\x7f`, and the upstream rejects the rewritten body with `invalid escaped character in string`. BEL, vertical tab, and some nonprinting Unicode characters trigger the same problem.

Use `encoding/json.Marshal` to serialize the text as a valid JSON string while preserving its content. Add a table-driven regression test covering both message positions selected for promotion, control characters, Unicode, JSON metacharacters, literal backslash sequences, cache metadata, and idempotence. The test validates the complete rewritten body with the strict standard-library JSON decoder.

Validation (from `backend/`):
- Confirmed the new DEL regression fails against unmodified upstream code and passes with this fix.
- `go test -p 3 -tags=unit ./...` — passed.
- `go test -p 2 -tags=integration ./...` — passed.
- `golangci-lint run --timeout=10m --concurrency=3 ./...` (v2.13.0) — 0 issues.
- `git diff --check` — passed.

### DIFF
--- a/backend/internal/service/gateway_messages_cache.go
+++ b/backend/internal/service/gateway_messages_cache.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/Wei-Shaw/sub2api/internal/pkg/claude"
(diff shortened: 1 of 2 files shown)


a1gaoshanlaiyige:

PR #6733, opened 2026-09-07
fix(codex): preserve allowed_tools selection constraints

AndrewZhuCC:
## Problem

`normalizeCodexToolChoice` treats object-form `tool_choice.type` as a declared tool type. However, `allowed_tools` is a selection policy, not a tool declaration. Since no tool is declared with that type, the normalizer replaces the policy with `"auto"`, silently dropping the caller's restriction.

This also affects policies referencing functions declared in `input.additional_tools`. Callers should be able to keep a stable top-level tool list for prompt caching while restricting which tools may be called for the current request.

## Changes

- Preserve `tool_choice: {"type":"allowed_tools", ...}` for upstream validation rather than downgrading it to `"auto"`. Malformed policies are likewise left to upstream validation, not silently weakened.
- Include the policy's nested function references in the existing reserved-name alias traversal, so a function such as `python` is renamed consistently in both its declaration and its allowlist reference.
- Add regressions for `auto`/`required` modes, top-level/`additional_tools` declarations, malformed policies, and reserved-name alias consistency.

The production change is 9 added lines across two existing helpers, plus 69 lines of tests. No new configuration, dependencies, database changes, caching implementation, or parallel-call policy changes.

## Validation

- [x] `cd backend && go test ./internal/service -run 'Test.*(CodexOAuthTransform|CodexToolChoice|CodexToolName)' -count=1 -timeout=120s`
- [x] `git diff --check origin/main...HEAD`
- [x] Built and deployed a Linux/amd64 image; smoke-tested `additional_tools` with an `allowed_tools` policy through the gateway. A real client-side confirmation request completed with the expected allowed decision call while `parallel_tool_calls` remained enabled.

Full repository CI is left to the upstream workflow. No separate issue was opened for this focused fix.

### DIFF
--- a/backend/internal/service/openai_codex_tool_names.go
+++ b/backend/internal/service/openai_codex_tool_names.go
@@ -115,6 +115,9 @@ func collectOpenAIResponsesToolNameFields(reqBody map[string]any) []codexToolNam
 		}
 	}
 	if choice, ok := reqBody["tool_choice"].(map[string]any); ok {
+		if strings.EqualFold(strings.TrimSpace(firstNonEmptyString(choice["type"])), "allowed_tools") {
+			collectTools(choice["tools"])
+		}
 		if strings.EqualFold(strings.TrimSpace(firstNonEmptyString(choice["type"])), "function") {
 			appendName(choice, "name")
 			if function, ok := choice["function"].(map[string]any); ok {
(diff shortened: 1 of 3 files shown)


AndrewZhuCC:

PR #6754, opened 2026-09-07
fix: cancel streaming upstream before closing body

ccemoji:
## Summary

- cancel the detached upstream child context before closing the response body on streaming Chat Completions requests
- preserve the existing detached-context behavior for buffered requests
- add a regression test that blocks the SSE reader until the upstream request context is canceled

## Why

The streaming Responses-to-Chat bridge can return after a terminal event while its keepalive scanner is still blocked in `Read`. The deferred `resp.Body.Close()` then runs without an owned cancellation signal. Cancelling a child context before closing the body makes the cleanup path bounded while keeping downstream client cancellation detached from the upstream request.

This is related to #5910, which documents detached upstream requests and missing lifecycle-stage observability.

## Validation

- `go test -tags=unit ./internal/service -run 'TestForwardAsChatCompletions' -count=1`
- targeted regression test passes
- existing `TestForwardAsChatCompletions_UpstreamRequestIgnoresClientCancel` passes

### DIFF
--- a/backend/internal/service/openai_gateway_chat_completions.go
+++ b/backend/internal/service/openai_gateway_chat_completions.go
@@ -331,6 +331,11 @@ func (s *OpenAIGatewayService) forwardAsChatCompletions(
 
 	// 6. Build upstream request
 	upstreamCtx, releaseUpstreamCtx := detachUpstreamContext(ctx)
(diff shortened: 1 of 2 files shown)


ccemoji:

PR #6821, opened 2026-09-08
fix(antigravity): preserve OAuth plan_type when saving accounts

### LINKED ISSUE #6820: [Bug] Antigravity OAuth drops plan_type before account creation, bypassing paid-tier endpoint selection
### Bug
Normal Antigravity OAuth account creation loses the subscription plan discovered from Google. Confirmed in v0.2.2 and current main (772a038 / v0.2.3).

### Reproduction
1. Authorize an Antigravity account for which loadCodeAssist returns paidTier.id = g1-pro-tier.
2. Complete the normal web account creation flow.
3. The exchange-code JSON response and the credentials submitted when creating the account omit plan_type.
4. Test Gemini 3.1 Pro High.

Expected: preserve the discovered pro plan so paid-account endpoint selection applies immediately.
Actual: accountHasAntigravityPaidTier returns false and resolveAntigravityForwardBaseURL selects the production endpoint. In the observed account, that returned 429 RESOURCE_EXHAUSTED; the daily endpoint succeeded. Correcting only the missing plan_type to pro then made the actual admin panel test pass. This is not a claim that every 429 has this cause.

### Root cause
- AntigravityTokenInfo.PlanType is populated by the OAuth service but tagged json:"-", so both exchange-code and refresh-token validation responses omit it.
- useAntigravityOAuth.buildCredentials independently drops plan_type even if it is present in the response.
- The shared frontend builder is used by OAuth creation, refresh-token import, and reauthorization.
- Background token refresh takes a different path through BuildAccountCredentials, which does preserve the field; it can therefore mask the initial omission after a successful subscription lookup.

### Proposed fix
Serialize plan_type with omitempty, type it in the frontend DTO, and preserve it in the shared credential builder only when returned by the upstream. Keep unknown plans absent. Regression tests should cover pro, ultra, free and unknown plans at both boundaries.

Related endpoint-selection change: #5612. No credentials, customer identifiers or private infrastructure are included in this report.

22zyfeng-collab:
## Problem and fix
Closes #6820.

The Antigravity OAuth service discovers the subscription plan, but its response DTO hides PlanType and the frontend credential builder drops plan_type. A newly authorized paid account consequently misses the paid-tier endpoint selection introduced in #5612 until subscription metadata is repaired or refreshed.

Expose plan_type as an optional JSON field and preserve it in the shared frontend builder used by account creation, refresh-token import and reauthorization. Unknown plans stay omitted; no plan is inferred or forced. The frontend API type now declares the field explicitly. Background refresh and endpoint-selection policies are unchanged.

## Validation
- Added regression tests failed against unmodified main for pro, ultra and free, at both JSON serialization and frontend credential construction boundaries.
- Targeted Go service tests passed: TestAntigravityTokenInfo, TestResolveDefaultTierID, TestResolveAntigravityForwardBaseURL.
- Vitest: useAntigravityOAuth and CreateAccountModal, 39 tests passed.
- Frontend vue-tsc --noEmit passed.
- git diff --check and Gitleaks staged-diff scan passed.

The live symptom was confirmed separately: preserving the Google-reported pro value made the real admin account test succeed. A fresh interactive Google authorization using this patch has not yet been exercised; the regression tests cover the two lossy boundaries without requiring live credentials.

### DIFF
--- a/backend/internal/service/antigravity_oauth_service.go
+++ b/backend/internal/service/antigravity_oauth_service.go
@@ -90,7 +90,7 @@ type AntigravityTokenInfo struct {
 	Email            string `json:"email,omitempty"`
 	ProjectID        string `json:"project_id,omitempty"`
 	ProjectIDMissing bool   `json:"-"`
-	PlanType         string `json:"-"`
+	PlanType         string `json:"plan_type,omitempty"`
 	PrivacyMode      string `json:"-"`
 }
 
--- a/backend/internal/service/antigravity_oauth_service_test.go
+++ b/backend/internal/service/antigravity_oauth_service_test.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"encoding/json"
 	"testing"
 )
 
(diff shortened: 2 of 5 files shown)


22zyfeng-collab:
我已阅读CLA文件，并在此签署CLA文件。

22zyfeng-collab:

PR #6833, opened 2026-09-08
fix(openai): reconcile stale rate limits after quota probe

lylelllll:
## Summary

- Reconcile stale OpenAI OAuth account cooldowns after a successful quota refresh.
- Share the recovery decision between the ordinary Codex response-header probe and the `/wham/usage` path.
- Protect database and in-process scheduler state with compare-and-swap generations.

## Problem

Ordinary OpenAI OAuth accounts refresh quota through `GetUsageBatch -> getOpenAIUsage -> probeOpenAICodexSnapshot`. The initial recovery hook only ran after `QueryUsage()` succeeded, which primarily covered Spark shadow accounts. As a result, an account could show a recovered quota window while the stale account-level rate-limit state continued to exclude it from scheduling.

## Safety

- Clear only when the returned quota data affirmatively shows recovery: required windows are below 100%, or have crossed a valid reset boundary.
- Missing, malformed, future, or still-exhausted windows do not clear the cooldown.
- Database clearing matches the exact observed `rate_limited_at` and `rate_limit_reset_at` values, so a newer 429 wins.
- Runtime scheduler blocks use a generation CAS; transient retry markers are not erased by an older probe.
- Spark shadow quota refreshes do not clear a parent account’s global cooldown.

## Tests

- `go test ./internal/service ./internal/repository -count=1`
- `go test -race ./internal/service -run 'OpenAI|RateLimit|Quota' -count=1`
- `go vet ./internal/service ./internal/repository`
- `go build ./cmd/...`
- `git diff --check`


lylelllll:
Thanks for v0.2.4. I noticed it fixes 429 responses that do not exhaust quota still causing account backoff.

PR #6833 addresses the remaining stale-state case: after a successful OpenAI quota probe confirms that the quota window has recovered, an old persisted/runtime cooldown may still keep the account unschedulable. This PR clears it only on affirmative recovery and uses generation/CAS guards so a newer 429 wins. It also covers the ordinary OAuth probe path, not only /wham/usage.

PR #6850, opened 2026-09-08
fix: anchor channel monitor date_bin buckets to UTC

reallyfeishi:
## Summary

- Anchor every channel-monitor `date_bin` origin to an explicit UTC timestamp.
- Use one shared origin/expression for fixed rollup SQL and ad-hoc bucketed repository queries.
- Add a regression test covering the UTC origin and all fixed-rollup SQL statements.

## Root cause

PostgreSQL interprets `TIMESTAMPTZ '1970-01-01'` in the current session timezone. With an `Asia/Shanghai` session this makes the epoch origin `1969-12-31 16:00:00Z`, so daily and other fixed buckets are persisted at `16:00Z`. The monitor frontend builds its timeline at UTC day boundaries (`00:00Z`) and performs exact bucket-key matching, so valid usage is rendered as empty/unknown cells for the 7d/30d views.

The fix uses `TIMESTAMPTZ '1970-01-01 00:00:00+00'`, making bucket boundaries deterministic regardless of PostgreSQL session timezone.

## Validation

- `git diff --check` passes.
- Added `TestChannelMonitorV2DateBinOriginIsUTC`.
- Go tests were not run locally because the Windows environment does not have `go`/`gofmt` installed; CI should run the repository's Go test and lint workflows.

## Deployment note

Existing rollup rows written with the shifted origin remain unchanged by this code-only fix. A one-time maintenance rebuild/repair of the affected channel-monitor history (up to the configured 90-day retention) is required after deployment so 7d/30d rows are rewritten with UTC-aligned bucket starts. This PR does not perform a destructive data migration automatically.

### DIFF
--- a/backend/internal/repository/channel_monitor_v2_aggregation.go
+++ b/backend/internal/repository/channel_monitor_v2_aggregation.go
@@ -395,11 +395,21 @@ func sameFixedRollupBucket(start, end time.Time, seconds int) bool {
 	return start.Truncate(interval).Equal(end.Add(-time.Nanosecond).Truncate(interval))
 }
 
+// PostgreSQL interprets a TIMESTAMPTZ literal without an explicit offset in
+// the current session timezone. Keep date_bin's origin fixed in UTC so bucket
+// boundaries do not shift when the database session runs in Asia/Shanghai (or
+// any other non-UTC timezone).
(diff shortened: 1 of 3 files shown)


reallyfeishi:

reallyfeishi:
recheck

PR #6869, opened 2026-09-09
fix(openai): accept full heartbeat bootstrap envelopes

### LINKED ISSUE #6868: fix: Codex Desktop full heartbeat envelopes fail HTTP call_id validation
### Problem

Codex Desktop automation heartbeat bootstrap requests fail with:

```text
function_call_output requires call_id on HTTP requests; continuation via previous_response_id is only supported on Responses WebSocket v2
```

The affected setup reports Sub2API v0.2.3. The same compatibility gap is present in the current main branch: `validCodexAutomationHeartbeat` accepts only `automation_id`, while the Desktop envelope also includes `current_time_iso` and `instructions`. This is a follow-up to #6547 and #6553.

### Reproduction

Send an HTTP Responses request with an automation bootstrap input (no `call_id` and no `previous_response_id`) matching this sanitized structure:

```json
{
  "type": "function_call_output",
  "name": "automation_update",
  "namespace": "codex_app",
  "output": "<heartbeat>\n  <automation_id>flutter</automation_id>\n  <current_time_iso>2026-09-09T00:33:34.775Z</current_time_iso>\n  <instructions>Check the task status.</instructions>\n</heartbeat>"
}
```

The heartbeat parser rejects the extra XML children, so bootstrap normalization is skipped and ordinary HTTP tool-output validation rejects the missing `call_id`. The existing ID-only heartbeat fixture does not cover this shape.

### Expected behavior

Recognize the full heartbeat bootstrap envelope and normalize it through the existing user-message path, preserving its original text. Continue accepting the legacy ID-only shape.

### Proposed fix

Allow only `automation_id`, `current_time_iso`, and `instructions`. Require a valid RFC3339 timestamp and non-empty instructions when the extended fields are present. Keep strict rejection of duplicate/unknown fields, attributes, namespaces, nested markup, and malformed XML. Preserve existing call-context gates: ordinary tool outputs must still require `call_id`, and requests carrying real call/response context must not be reclassified.

Add regression tests for the full envelope, lossless/idempotent normalization, malformed variants, and protected call contexts.

Garfielk:
Fixes #6868. Follow-up to #6553 and #6547.

Codex Desktop sends automation bootstrap outputs containing `automation_id`, `current_time_iso`, and `instructions`. The current heartbeat matcher accepts only `automation_id`, so full envelopes are not normalized and HTTP requests fail with `function_call_output requires call_id` before inference.

Accept the full envelope with a valid RFC3339 timestamp and non-empty instructions, while retaining the existing ID-only format. Preserve the original output text and reject duplicate/unknown fields, attributes, nested elements, namespaces, and malformed XML. Existing namespace, call ID, and response-context gates remain unchanged; ordinary tool-output validation is not relaxed.

Regression coverage includes the full observed envelope shape with a sanitized prompt, lossless normalization, idempotency, invalid timestamps, incomplete envelopes, duplicate fields, and protected call contexts.

### Validation

Checked the repository `DEV_GUIDE.md` pre-PR checklist and CI configuration. Local commands ran in the official `golang:1.27` container (resolved to Go 1.27.1 linux/arm64; CI pins Go 1.27.0).

- Passed: `go test ./internal/handler -run TestNormalizeCodex -count=1 -timeout=120s`.
- Passed: `gofmt` and `git diff --check`.
- Attempted: `go test -tags=unit ./... -timeout=180s`. The handler package passed; the service package compiler was killed. Container cgroup counters confirmed an OOM kill, so the full unit suite did not pass.
- Exited successfully: `go test -tags=integration ./... -timeout=180s`. Docker-dependent tests are skipped by the existing harness because the test container has no Docker socket; this is not complete integration validation.
- Attempted: `golangci-lint run ./... --timeout=5m` with v2.13.0. The process was killed (exit 137), so lint is not verified.

Full CI validation remains required. No frontend dependencies, interfaces, or Ent schemas changed.

### DIFF
--- a/backend/internal/handler/openai_automation_bootstrap_test.go
+++ b/backend/internal/handler/openai_automation_bootstrap_test.go
@@ -166,3 +166,44 @@ func TestNormalizeCodexAutomationBootstrapPreservesOrderAndIsIdempotent(t *testi
 	require.False(t, changedAgain)
 	require.Equal(t, got, again)
 }
+
+func TestNormalizeCodexAutomationBootstrapFullHeartbeat(t *testing.T) {
+	const output = `<heartbeat>
+  <automation_id>flutter</automation_id>
+  <current_time_iso>2026-09-09T00:33:34.775Z</current_time_iso>
+  <instructions>
+Read the reference index and report changes. Preserve A &amp; B and &lt;tags&gt;.
+  </instructions>
+</heartbeat>`
+	body := codexAutomationBootstrapBody(t, output, "")
+	got, changed := normalizeCodexAutomationBootstrap(body)
+	require.True(t, changed)
+	require.Equal(t, "user", gjson.GetBytes(got, "input.0.role").String())
+	require.Equal(t, output, gjson.GetBytes(got, "input.0.content.0.text").String())
+	again, changedAgain := normalizeCodexAutomationBootstrap(got)
+	require.False(t, changedAgain)
+	require.Equal(t, got, again)
+
+	for _, field := range []string{"automation_id", "current_time_iso", "instructions"} {
+		duplicate := strings.Replace(output, "</heartbeat>", "<"+field+">duplicate</"+field+"></heartbeat>", 1)
+		require.False(t, validCodexAutomationHeartbeat(duplicate), field)
+	}
+	for _, invalid := range []string{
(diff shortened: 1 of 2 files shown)


Garfielk:

PR #6924, opened 2026-09-09
fix: validate Codex User-Agent before identity parsing

smyyyyy2025:
# fix: validate Codex User-Agent before identity parsing

## Summary

Validate the HTTP field value before pairing a Codex User-Agent with its
originator. Keep raw canonical/override values available until that validation
has run, and fall back consistently when a canonical value is invalid.

## Problem

The existing parser checks the client-name portion, but can accept forbidden
bytes in the remainder of the User-Agent. Trimming a value before validation can
also hide leading or trailing newlines. An invalid canonical UA can otherwise
contribute part of the fallback identity.

## Changes

- Reuse the existing `httpguts.ValidHeaderFieldValue` implementation.
- Validate before normalization inside the shared identity resolver.
- Keep valid platform/terminal suffixes unchanged.
- Add parser and canonical/override resolver regression coverage.

This is defensive validation, not a claim of an exploitable header-injection
vulnerability. HTTP transports may reject these values independently, and other
callers can still normalize input before reaching this layer.

## Scope

No transparent-mode version policy changes, random UA selection, installation
or session ID changes, configuration changes, or new dependencies.

The HTTP and WebSocket paths already share identity handling; this improves that
existing shared implementation rather than introducing that sharing.

## Testing

Independent verification results are recorded in the accompanying `README.md`.

Not verified: full backend suite, race detector, live inference, or performance.
There is no demonstrated causal connection to issue #6911 and this PR does not
claim to fix it.

### DIFF
--- a/backend/internal/pkg/openai/request.go
+++ b/backend/internal/pkg/openai/request.go
@@ -3,6 +3,8 @@ package openai
 import (
 	"regexp"
 	"strings"
+
+	"golang.org/x/net/http/httpguts"
 )
 
 // CodexCLIUserAgentPrefixes matches Codex CLI User-Agent patterns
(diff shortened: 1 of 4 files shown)


smyyyyy2025:

smyyyyy2025:
recheck

Issue #3302, opened 2026-06-16
OpenAI OAuth/Codex image_generation ignores explicit gpt-image-2 size/quality and returns gpt-image-2-codex auto/auto low-resolution output


jzhao20230918:
## Summary

When using OpenAI OAuth/Codex accounts through sub2api, explicit `gpt-image-2` image generation requests with `quality=high` and a custom 4K-compatible size are accepted with HTTP 200, but the upstream lifecycle metadata changes the image tool to:

```json
{
  "model": "gpt-image-2-codex",
  "quality": "auto",
  "size": "auto"
}
```

The actual returned PNG is low-resolution instead of the requested size.

This happens on both:

- `/v1/responses` with an explicit `image_generation` tool
- `/v1/images/generations`, which is bridged through Codex Responses for OAuth accounts

The main concern is silent downgrade: requests that clearly express native image intent can still route through OAuth/Codex and return downgraded images instead of preserving size/quality or failing fast.

## Request Example

`/v1/responses` request shape:

```json
{
  "model": "gpt-5.5",
  "instructions": "Generate the requested image.",
  "input": [
    {
      "role": "user",
      "content": "Create a simple vertical poster."
    }
  ],
  "tools": [
    {
      "type": "image_generation",
      "action": "generate",
      "model": "gpt-image-2",
      "quality": "high",
      "size": "2160x3040",
      "partial_images": 1
    }
  ],
  "tool_choice": {
    "type": "image_generation"
  },
  "stream": true,
  "store": false
}
```

Expected output: a `2160x3040` PNG, or a clear error saying the selected OAuth/Codex account cannot satisfy explicit high/custom-size image generation.

Observed lifecycle metadata:

```json
{
  "type": "image_generation",
  "model": "gpt-image-2-codex",
  "quality": "auto",
  "size": "auto",
  "output_format": "png"
}
```

Observed actual PNG dimensions across repeated tests include:

- `1024x1536`
- `1054x1492`
- `1058x1487`
- `1073x1466`
- `1254x1254`
- `1693x929`

None matched the requested `2160x3040`.

## Images API Also Affected

`/v1/images/generations` request shape:

```json
{
  "model": "gpt-image-2",
  "prompt": "Create a simple vertical poster.",
  "n": 1,
  "quality": "high",
  "size": "2160x3040",
  "response_format": "b64_json"
}
```

Observed response top-level metadata can also be:

```json
{
  "model": "gpt-image-2-codex",
[...]

PR #6939, opened 2026-09-10
fix(openai): preserve OAuth image model routing

qiu121:
## Summary

- Route OpenAI OAuth and Setup Token `/v1/images/generations` and `/v1/images/edits` requests through ChatGPT's dedicated Codex Images endpoints.
- Forward the requested image model and native image parameters without translating them into the Responses `image_generation` bridge.
- Keep the existing OpenAI Images API response shape, including JSON, SSE, multi-image, edit, URL-to-`b64_json`, and downstream streaming compatibility.
- Preserve the upstream response model separately from the model sent by the gateway, and mark mismatches in usage records.

## Motivation

The official `v0.2.4` release fixed the controller-side Image 2.5 model selection, but it did not fully resolve the OAuth/Codex execution path. The previous bridge could accept a request for `gpt-image-2.5-*` while the upstream lifecycle response reported `gpt-image-2-codex` with `auto` image parameters. That makes a successful HTTP response misleading and hides the actual executor from usage records.

The dedicated Codex Images transport sends, for example:

```json
{
  "model": "gpt-image-2.5-flare",
  "prompt": "...",
  "size": "3840x2160",
  "quality": "high",
  "n": 1
}
```

If the upstream still declares another processing model, the client-facing image response remains compatible while the usage record retains both values and sets `upstream_model_mismatch=true`.

## Scope

This PR fixes the dedicated `/v1/images/*` OAuth path covered by the related reports. Mixed `/v1/responses` requests continue to use Responses semantics; their image-tool routing remains a separate compatibility path and is intentionally not rewritten into an Images API response here.

## Validation

- `go test -tags=unit ./...`
- `git diff --check`
- Added regression coverage for `gpt-image-1`, `gpt-image-2`, and `gpt-image-2.5-flare`, including final usage-log persistence.

## Related Issues

Refs #3302
Refs #6905
Refs #6896

### DIFF
--- a/backend/internal/service/openai_gateway_record_usage_test.go
+++ b/backend/internal/service/openai_gateway_record_usage_test.go
@@ -1622,6 +1622,42 @@ func TestOpenAIGatewayServiceRecordUsage_UsesRequestedModelAndUpstreamModelMetad
 	require.Equal(t, 1, userRepo.deductCalls)
 }
 
+func TestOpenAIGatewayServiceRecordUsage_PersistsImageUpstreamModelMismatch(t *testing.T) {
+	usageRepo := &openAIRecordUsageLogRepoStub{inserted: true}
+	userRepo := &openAIRecordUsageUserRepoStub{}
+	subRepo := &openAIRecordUsageSubRepoStub{}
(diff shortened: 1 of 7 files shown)


qiu121:
根据维护者沟通，关闭此 PR，改用中文重新提交，代码内容保持不变。

PR #6972, opened 2026-09-10
fix(antigravity): send thinkingLevel so Gemini 3.x returns thought summaries

WwWangGuan:
## Problem

When a client enables thinking (e.g. Claude Code / any Anthropic-compatible client) and calls an **Antigravity Gemini 3.x** model through `/antigravity/v1/messages`, the upstream **does not return any `thinking` block**, so the client never sees the reasoning.

Claude models on the same account/path work fine, so this is specific to the Antigravity Gemini path.

## Root cause

Antigravity's Gemini 3.x models express reasoning depth through `generationConfig.thinkingConfig.thinkingLevel` (`low` / `medium` / `high`). `buildGenerationConfig` only emits `includeThoughts` + `thinkingBudget`, and a **positive `thinkingBudget` suppresses the thought summary** upstream.

Direct probes against the Antigravity `daily-cloudcode-pa` endpoint with the same OAuth token:

| thinkingConfig | thought summary returned |
| --- | --- |
| `{includeThoughts:true}` | ❌ |
| `{includeThoughts:true, thinkingBudget:2048}` | ❌ |
| `{includeThoughts:true, thinkingLevel:"high"}` | ✅ |
| `{thinkingLevel:"high"}` | ❌ |

So both `includeThoughts` **and** `thinkingLevel` are required, and `thinkingBudget` must not be sent for these models.

## Fix

- `GeminiThinkingConfig` gains a `ThinkingLevel` field.
- For Gemini 3.x models (`gemini-3*`), `buildGenerationConfig` now emits `thinkingLevel` (derived from the `-low` / `-medium` / `-high` variant suffix, defaulting to `high` for `-tiered` and bare ids) and omits `thinkingBudget`.
- Claude and Gemini 2.5 models keep the existing budget-based behavior.

## Testing

- Added `TestBuildGenerationConfig_GeminiThinkingLevel` covering `-tiered` / `-high` / `-medium` / `-low`, `gemini-3.1-pro-low`, and a `gemini-2.5-flash` regression case (still budget-based).
- Manually verified against a live Antigravity account: `gemini-3.8-flash-tiered` via `/antigravity/v1/messages` now streams `thinking_delta` events for non-trivial prompts (simple prompts may legitimately produce no summary — upstream behavior).

Note: the model catalog entries for Gemini 3.7/3.8 are tracked separately in #6575; this change only concerns the thinking configuration for models already routable (3.6 and earlier).

### DIFF
--- a/backend/internal/pkg/antigravity/gemini_types.go
+++ b/backend/internal/pkg/antigravity/gemini_types.go
@@ -77,9 +77,14 @@ type GeminiImageConfig struct {
 }
 
 // GeminiThinkingConfig Gemini thinking 配置
+//
+// ThinkingLevel controls reasoning depth for Gemini 3.x models
+// ("low" / "medium" / "high"). Antigravity only returns thought summaries when
+// it is set; a positive ThinkingBudget suppresses them.
 type GeminiThinkingConfig struct {
-	IncludeThoughts bool `json:"includeThoughts"`
-	ThinkingBudget  int  `json:"thinkingBudget,omitempty"`
+	IncludeThoughts bool   `json:"includeThoughts"`
+	ThinkingBudget  int    `json:"thinkingBudget,omitempty"`
+	ThinkingLevel   string `json:"thinkingLevel,omitempty"`
 }
 
 // GeminiToolDeclaration Gemini 工具声明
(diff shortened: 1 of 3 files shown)


WwWangGuan:
Withdrawing this PR.

PR #6979, opened 2026-09-10
fix(openai): support OAuth same-account retry overrides

liuliping:
## Summary

- Allow OpenAI OAuth accounts to explicitly configure `credentials.pool_mode_retry_count` without enabling API-key pool mode.
- Respect zero as disabling handler-level same-account retries, including the OAuth 429 deadline branch. Explicit positive overrides cap retries within the existing deadline; missing overrides preserve existing behavior.
- Add the existing localized same-account retry control to the OAuth account editor. Support load/save/reload and clearing an override without writing defaults during unrelated edits.
- Preserve the override in scheduler metadata, without retaining OAuth access tokens.

## Motivation

OpenAI OAuth capacity failures can enter same-account retry handling, but `GetPoolModeRetryCount` previously returned the default for non-pool accounts. OAuth accounts also lacked the editor control. An operator could therefore neither configure the behavior through the UI nor make a manually stored zero effective. Deadline-based 429 handling additionally bypassed a zero retry limit.

## Scope

No schema migration. No changes to account types, quota safeguards, concurrency limits, session affinity or cross-account failover. This is not a request-wide attempt budget and does not cap internal transport reconnects. No production data or deployment configuration is included.

## Verification

- Backend unit regressions for OAuth zero/positive overrides, default preservation, deadline retries, failover behavior and scheduler metadata passed.
- Account editor tests: 60 passed, including save/reload, clearing an override, and preserving an unset value.
- Locale completeness: 3 passed.
- Frontend typecheck, targeted ESLint and production build passed.
- `git diff --check` passed.

Unit/component tests use synthetic fixtures. No production deployment was performed.


liuliping:

liuliping:
recheck

PR #7029, opened 2026-09-11
fix(responses): apply compatible-upstream reasoning summary contract

HTExplicit:
## Summary
Replace the PNNL host/model whitelist hotfix with a destination reasoning.summary contract.

- Official OpenAI API Key `/v1/responses` keeps the client summary.
- Other OpenAI API Key Responses hosts default to `auto`.
- Account Extra `openai_reasoning_summary_mode` can be `passthrough`, `auto`, or `omit`.
- GPT-5.5 and Luna still return wrapped `invalid_request_error` without `error.param` or an upstream body, so this change does not add model-specific field strips.

## Test plan
- `go test ./internal/pkg/openai_compat ./internal/service -count=1 -run '^(TestResolveReasoningSummaryMode|TestParseReasoningSummaryMode|TestNormalizeOpenAICompatibleResponsesReasoningSummary|TestNormalizeOpenAICompatibleResponsesReasoningSummaryModels|TestNormalizeOpenAICompatibleResponsesReasoningSummaryOmit|TestOpenAICompatibleResponsesReasoningSummaryHTTPBuilders)$'`


HTExplicit:
Opened against the wrong repository.

PR #7038, opened 2026-09-11
fix(openai): remove Responses Lite markers after OAuth mapping to GPT-5.5

zenggang:
## Background

OpenAI Codex clients may send the X-OpenAI-Internal-Codex-Responses-Lite marker when requesting GPT-5.6 or GPT-6 models. Some OAuth accounts map those requested models to GPT-5.5 through model_mapping, but the marker remains.

## Problem

OpenAI rejects the incompatible combination with an error similar to: This model is not supported when using X-OpenAI-Internal-Codex-Responses-Lite. This can affect Responses, passthrough, and WS-to-HTTP bridge requests after model mapping.

## Solution

Apply a final outbound compatibility transform after account and model resolution. For resolved OAuth upstream model gpt-5.5, remove the Lite header and corresponding WS client metadata. Preserve ingress body and headers for retries or failover. Leave native GPT-5.6/GPT-6, API Key, and non-OpenAI accounts unchanged, preserving tools, namespaces, reasoning context, history, and tool results.

## Tests

Covers mapped and native model behavior, account-specific routing, HTTP, passthrough, and WS-to-HTTP builder paths. Full Go test suite passes.

### DIFF
--- a/backend/internal/service/openai_gateway_forward.go
+++ b/backend/internal/service/openai_gateway_forward.go
@@ -1517,6 +1517,9 @@ func (s *OpenAIGatewayService) buildUpstreamRequest(ctx context.Context, c *gin.
 	setOpenAICodexRoutingHintFromBody(req.Header, account, body)
 	logOpenAIRoutingDiagnosticsFromBody(ctx, account, "http", req.Header, body, "not_applicable")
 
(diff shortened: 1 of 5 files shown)


zenggang:
recheck

zenggang:

PR #7077, opened 2026-09-12
fix(antigravity): handle prefixItems and ensure array schema has valid items fallback

ggllzz04:
### Summary

Fixes the 400 Bad Request error when using Claude Code (2.1+) or other tools with JSON Schema Draft 2020-12 `prefixItems` (tuple schemas, such as `query.where` in Claude Code's Artifact tool) against Gemini / Antigravity backend:
`* GenerateContentRequest.tools[0].function_declarations[...].parameters.properties[query].properties[where].items.items: missing field.`

### Problem
- In JSON Schema Draft 2020-12, array tuples use `prefixItems` rather than array-form `items`. Because `prefixItems` was omitted in the allowlist and not handled, the nested schema ended up with `type: array` but no `items`.
- Google Gemini's Protobuf specification strictly requires any `type: "array"` schema to have a valid `items` schema object. If `items` is missing or empty, Google API returns a `400 INVALID_ARGUMENT (missing field: ...items.items)`.
- Additionally, the previous recursive step used `else if items` after checking `properties`, which prevented `items` from being recursively cleaned if a schema contained both.

### Changes
1. **Handle `prefixItems`**: Extracts the best schema from `prefixItems` and converts it into a valid `items` schema object when `items` is absent or boolean.
2. **Array `items` fallback**: Automatically injects `items: {"type": "string"}` if an array schema has no `items` or an empty `items` object.
3. **Separate traversal**: Decoupled `properties` and `items` recursion to ensure full coverage.
4. **Unit Tests**: Added test cases in `schema_cleaner_test.go` verifying `prefixItems` conversion, nested empty arrays, and normal array preservation.

### DIFF
--- a/backend/internal/pkg/antigravity/schema_cleaner.go
+++ b/backend/internal/pkg/antigravity/schema_cleaner.go
@@ -120,16 +120,21 @@ func cleanJSONSchemaRecursive(value any) any {
 	mergeAllOf(schemaMap)
 
 	// 0.5 [FIX] 处理 Draft 2020-12 的 prefixItems (元组定义)
-	if prefixItems, ok := schemaMap["prefixItems"].([]any); ok && len(prefixItems) > 0 {
-		itemsVal, hasItems := schemaMap["items"]
-		_, itemsIsBool := itemsVal.(bool)
-		if !hasItems || itemsVal == nil || itemsIsBool {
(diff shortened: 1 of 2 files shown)


ggllzz04:

PR #7079, opened 2026-09-12
fix(build): restore the embedded dist placeholder

Dev-next-gen:
While building this repo with `-tags embed` I hit a hard stop on a clean checkout:

```
$ cd backend && go build -tags embed ./...
internal/web/embed_on.go:29:12: pattern all:dist: no matching files found
```

Digging into it, the placeholder that is supposed to prevent this was never committed. `.gitignore` has carried these lines since aa89777dd (`ci(backend): 调整 embed server`):

```
# Keep a placeholder file so `//go:embed all:dist` always has a match in CI/lint,
# while still ignoring generated frontend build outputs.
backend/internal/web/dist/
!backend/internal/web/dist/
backend/internal/web/dist/*
!backend/internal/web/dist/.keep
```

but `git ls-files backend/internal/web/` only returns the five `.go` files — the negation rule points at a file that does not exist, so the exemption never had anything to exempt.

The consequence is limited to tooling, but it is total: `embed` is the tag both `Dockerfile` and `deploy/Dockerfile` pass to `go build`, and it is what `.goreleaser.yaml` puts in `builds[].flags`, so `internal/web/embed_on.go` and the 911 lines of `embed_test.go` next to it cannot be compiled, vetted or linted from a plain checkout. `go vet -tags embed ./...`, `golangci-lint --build-tags embed` and `go test -tags=embed ./internal/web/...` all die on the same `//go:embed` line before any analysis runs. Release builds are fine because the frontend stage populates `internal/web/dist` first; nobody else is.

This adds the empty placeholder the ignore rule already describes.

## Proof

Before, on `main`:

```
$ go build -tags embed ./...
internal/web/embed_on.go:29:12: pattern all:dist: no matching files found
exit=1
$ go vet -tags embed ./internal/web/...
internal/web/embed_on.go:29:12: pattern all:dist: no matching files found
exit=1
```

After, with the placeholder:

```
$ go build -tags embed ./...
exit=0
$ go vet -tags embed ./...          # covers embed_on.go and embed_test.go
exit=0
$ GOOS=darwin  GOARCH=arm64 go build -tags embed ./cmd/server   # darwin/arm64 OK
$ GOOS=windows GOARCH=amd64 go build -tags embed ./cmd/server   # windows/amd64 OK
$ GOOS=linux   GOARCH=arm64 go build -tags embed ./cmd/server   # linux/arm64 OK
```

`go test -tags=unit ./...` stays green (Go 1.27.0, linux/amd64).

## Why this is safe

I checked the three places the extra file could leak into:

- Docker: `.dockerignore` has `dist/`, so `backend/internal/web/dist` never enters the build context, and the image is unchanged. The real assets still arrive via `COPY --from=frontend-builder`.
- goreleaser / release archives: the frontend build writes into the same directory, so the placeholder just sits alongside `index.html` as a zero-byte file.
- Runtime: `HasEmbeddedFrontend()` gates on `dist/index.html`, not on the directory, and both `cmd/server/main.go` and `internal/server/router.go` branch on it. A binary built with `-tags embed` but no frontend therefore behaves exactly as it does today — it just fails at startup detection instead of at compile time.

Nothing else in the tree changes; the diff is one empty file.

AI tools used


Dev-next-gen:

PR #7082, opened 2026-09-12
fix(antigravity): isolate OAuth token cache key by account ID to prevent 429 collision

ggllzz04:
### Summary
Isolate Antigravity OAuth Access Token cache keys by `account.ID` (aligned with OpenAI, Claude, and Grok) instead of `project_id`.

### Problem
For personal Antigravity (Google Gemini) accounts, the upstream OAuth flow returns a common, fixed project ID: `"aicode-consumers"`.

Previously, `AntigravityTokenCacheKey` used `project_id` when present:
```go
func AntigravityTokenCacheKey(account *Account) string {
    projectID := strings.TrimSpace(account.GetCredential("project_id"))
    if projectID != "" {
        return "ag:" + projectID
    }
    return "ag:account:" + strconv.FormatInt(account.ID, 10)
}
```
This causes all personal accounts to share the exact same Redis cache key: `oauth:token:ag:aicode-consumers`.

**Consequence:**
When Account A exhausts its 5-hour quota or triggers an upstream 429 rate limit, the Sub2API scheduler flags Account A and switches to Account B. However, Account B reads Account A's cached token from Redis and sends it upstream. Google immediately rejects it with another 429, causing the scheduler to cascade-mark all healthy Antigravity accounts in the pool as rate-limited, resulting in a false-positive `503 No available accounts` outage.

### Solution
- Change `AntigravityTokenCacheKey` to always use `"ag:account:" + strconv.FormatInt(account.ID, 10)`, consistent with OpenAI and Claude.
- Update unit test cases in `token_cache_key_test.go`.

### Verification
- Unit tests (`TestAntigravityTokenCacheKey`) pass.
- Verified in live environment with multiple personal Antigravity accounts: independent Redis cache keys (`oauth:token:ag:account:<id>`) are properly generated and account failover operates without token collision.

### DIFF
--- a/backend/internal/service/antigravity_token_provider.go
+++ b/backend/internal/service/antigravity_token_provider.go
@@ -231,9 +231,5 @@ func (p *AntigravityTokenProvider) markBackfillAttempted(accountID int64) {
 }
 
 func AntigravityTokenCacheKey(account *Account) string {
-	projectID := strings.TrimSpace(account.GetCredential("project_id"))
-	if projectID != "" {
-		return "ag:" + projectID
-	}
 	return "ag:account:" + strconv.FormatInt(account.ID, 10)
 }
--- a/backend/internal/service/token_cache_invalidator.go
+++ b/backend/internal/service/token_cache_invalidator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"strconv"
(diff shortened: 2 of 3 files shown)


wx2cyj:
我也是  很烦 而且老是429  可以修复吗

PR #7098, opened 2026-09-13
fix(antigravity): emit visible retryable text on MALFORMED_FUNCTION_CALL instead of empty 200

### LINKED ISSUE #7081: [Bug] Antigravity: MALFORMED_FUNCTION_CALL falls through to stop_reason=end_turn and returns HTTP 200 with empty content
## Summary

When Gemini returns `finishReason: "MALFORMED_FUNCTION_CALL"`, the Antigravity compat stream transformer does not handle that reason and falls through to the default `stop_reason: "end_turn"` with **zero content blocks**. Clients therefore receive a **successful but completely empty** completion, instead of an error.

Observed on `gemini-3.8-flash` via an Antigravity OAuth account, in large contexts.

## Environment

- Image: `weishaw/sub2api:latest`
- Version: `0.2.4`
- Revision: `5de5e2bed035d43591a2e10e51f420ef6a84eb98`
- Digest: `sha256:ccf47a1c62e355f51f896e489f8253e119fe4101b103cd701ba458cc6c6f0f77` (verified identical to the registry's current `latest`)
- Model: `gemini-3.8-flash` (maps to `gemini-3.8-flash-tiered`)
- Ingress: `POST /v1/chat/completions`, `stream: true`

## Evidence

The gateway logs the upstream signal:

```
INFO stdlog service/antigravity_gateway_compat_stream.go:136
  [Antigravity] MALFORMED_FUNCTION_CALL detected in stream for model gemini-3.8-flash

INFO stdlog service/antigravity_gateway_compat_stream.go:136
  [Antigravity] Malformed content: {"role":"model","parts":[{"thoughtSignature":"..."}]}
```

Note the content: a **thought signature only** — no `text`, no valid `functionCall`.

The HTTP response is still **200**:

```
INFO middleware/logger.go:82 http request completed
  {"status_code": 200, "model": "gemini-3.8-flash", "platform": "antigravity", "account_id": 3}
```

…but the client receives no content at all.

## Rate and correlation

Over a 24h window on this instance:

| Metric | Value |
|---|---|
| Total Antigravity completions | 1310 |
| `MALFORMED_FUNCTION_CALL` events | **41** |
| …attributed to `gemini-3.8-flash` | **41 / 41 (100%)** |
| …attributed to any other model | **0** |
| Request size preceding every event | **>= ~930 KB** (traced range 0.93–2.77 MB) |

Every occurrence was preceded by a large request body; none occurred on small requests. Retries of the same turn reproduce it, producing bursts (e.g. 11 events inside ~90s).

## Where it looks wrong

`backend/internal/pkg/antigravity/stream_transformer.go` detects the reason purely for logging at line ~134:

```go
if finishReason == "MALFORMED_FUNCTION_CALL" {
    log.Printf("[Antigravity] MALFORMED_FUNCTION_CALL detected in stream for model %s", p.originalModel)
    if geminiResp.Candidates[0].Content != nil {
        if b, err := json.Marshal(geminiResp.Candidates[0].Content); err == nil {
            log.Printf("[Antigravity] Malformed content: %s", string(b))
        }
    }
}
```

…but `emitFinish()` (~line 513) then maps it as a normal completion:

```go
stopReason := "end_turn"
if p.usedTool {
    stopReason = "tool_use"
} else if finishReason == "MAX_TOKENS" {
    stopReason = "max_tokens"
}
```

`MALFORMED_FUNCTION_CALL` matches none of these, and `p.usedTool` is false (the call never parsed), so it becomes `end_turn`. Because the malformed part carried no text, the message ends with **no content blocks** and `finish_reason: "stop"`.

## Why this is worth fixing

Downstream, a contentless `stop` is indistinguishable from "the model chose to say nothing". An agent harness will treat it as a completed turn:

- In our case the client raised `EMPTY_RESPONSE` ("model returned a completed response with no content") and retried — hitting the same deterministic malformed call each time, so the turn could never complete.
- It also bills/records a normal successful request while the turn is effectively dead.

## Suggested behaviour

Either:

1. map `MALFORMED_FUNCTION_CALL` to a distinct, non-`end_turn` outcome and surface it to the client (an error event/finish reason), or
2. if content is truly empty, fail the turn rather than emitting a contentless success.

## Reference

Related to #7080 (a separate mixed-tool 400 on the same Antigravity compat path).

MHJoy99:
Upstream Gemini returns finishReason `MALFORMED_FUNCTION_CALL` with thoughtSignature-only content on large contexts. It fell through to `stop_reason: end_turn` with zero content blocks, so agents got HTTP 200 with empty content and retried the same turn indefinitely.

Change: emit a visible retryable text block in `emitFinish` (`stream_transformer.go`), keeping the valid `end_turn` stop reason (`error` is not a valid Anthropic stop_reason).

Fixes #7081


MHJoy99:

MHJoy99:
recheck

PR #7100, opened 2026-09-13
fix(antigravity): degrade mixed search+function tools instead of hard-400

### LINKED ISSUE #7080: [Bug] v0.2.4 Antigravity: Claude models still 400 with tool_config.include_server_side_tool_invocations on mixed built-in + function tools
## Summary

On **v0.2.4** (current `latest`), Antigravity requests using **Claude models** are rejected with `400` whenever the request mixes **function tools with a built-in tool** (`web_search`). The same request shape works on Gemini models, and works on Claude models if the built-in tool is removed.

This looks like a **recurrence of #5843** (closed by #5860 "preserve mixed built-in chat tools"), but narrowed to the Claude path.

## Environment

- Image: `weishaw/sub2api:latest`
- Version: `0.2.4`
- Revision: `5de5e2bed035d43591a2e10e51f420ef6a84eb98`
- Digest: `sha256:ccf47a1c62e355f51f896e489f8253e119fe4101b103cd701ba458cc6c6f0f77` (verified identical to the registry's current `latest`)
- Account: Antigravity OAuth (`platform: antigravity`)
- Ingress: `POST /v1/chat/completions`, `stream: true|false`

## Reproduction

Two requests that differ **only** by the presence of a `web_search` tool.

**A — function tool only → 200 OK**

```bash
curl -s -o /dev/null -w "%{http_code}\n" -X POST "$BASE/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{
    "model": "claude-sonnet-4-6",
    "messages": [{"role":"user","content":"say OK"}],
    "max_tokens": 32, "stream": false,
    "tools": [{"type":"function","function":{"name":"bash","description":"run a command",
      "parameters":{"type":"object","properties":{"command":{"type":"string"}},"required":["command"]}}}]
  }'
# => 200
```

**B — same request plus a built-in tool → 400**

```bash
curl -s -X POST "$BASE/v1/chat/completions" \
  -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{
    "model": "claude-sonnet-4-6",
    "messages": [{"role":"user","content":"say OK"}],
    "max_tokens": 32, "stream": false,
    "tools": [
      {"type":"function","function":{"name":"bash","description":"run a command",
        "parameters":{"type":"object","properties":{"command":{"type":"string"}},"required":["command"]}}},
      {"type":"web_search","name":"web_search"}
    ]
  }'
# => 400
```

Response body returned to the client:

```json
{"error":{"code":null,"message":"Upstream request failed","param":null,"type":"upstream_error"}}
```

Upstream error recorded by the gateway:

```
ERROR handler/gateway_handler_chat_completions.go:321 gateway.cc.forward_failed
  {"model": "claude-sonnet-4-6", "stream": true, "account_id": 2,
   "error": "upstream error: 400 Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling."}
```

## Model scope

| Model | function tools only | + `web_search` |
|---|---|---|
| `claude-sonnet-4-6` | 200 | **400** |
| `gemini-3.8-flash` | 200 | 200 |

So the failure is specific to the **Claude** branch of the Antigravity compat path.

## Where it looks wrong

In `backend/internal/service/antigravity_gateway_compat.go`, `buildAntigravityCompatGeminiBody` injects the flag **only on the `gemini-*` branch**:

```go
if strings.HasPrefix(strings.ToLower(mappedModel), "gemini-") {
    body, err := convertClaudeMessagesToGeminiGenerateContent(claudeBody)
    ...
    body, err = enableMixedGeminiToolInvocations(body)   // <-- only here
    ...
}

// Claude models fall straight through to:
options := s.getClaudeTransformOptions(ctx)
options.EnableIdentityPatch = true
return antigravity.TransformClaudeToGeminiWithOptions(claudeRequest, projectID, mappedModel, options)
```

The Claude branch relies instead on `hasMixedToolInvocations()` in `backend/internal/pkg/antigravity/request_transformer.go:162`, which sets the **camelCase Gemini** field:

```go
if hasMixedToolInvocations(tools) {
    enabled := true
    innerRequest.ToolConfig.IncludeServerSideToolInvocations = &enabled
}
```

Note the upstream message names **`tool_config.include_server_side_tool_invocations`** — snake_case, i.e. the **Anthropic-shaped** field. So for Claude models the upstream appears to expect the flag in the Anthropic `tool_config` envelope, whereas the code above writes it into the Gemini `toolConfig`. That would explain why the Gemini branch passes (its raw-body injection lands where the upstream reads it) while Claude models do not.

I have not captured the exact outbound body for the Claude path, so treat that last paragraph as the leading hypothesis rather than a confirmed root cause — but the branch asymmetry above is the obvious place to look.

## Impact

Any client that always advertises a `web_search`-style tool (very common for agent harnesses) cannot use **any Claude model through an Antigravity account** — every request 400s. Gemini models on the same account are unaffected.

## Workaround

Remove the built-in/`web_search` tool from the request; function tools alone work.

## References

- #5843 — same 400, closed by #5860
- #5709 — referenced in the source comment as the original report
- #5860 `fix(antigravity): preserve mixed built-in chat tools`

MHJoy99:
Claude models fail 100% with HTTP 400 on mixed built-in search + function tools. I verified live that v1internal rejects the mix on every tested model (2.5-flash, 3.6-high, 3.8-tiered) even with `include_server_side_tool_invocations` set (also matches LiteLLM, which drops search tools on conflict).

Change: drop the googleSearch builtin and keep function declarations (with a server log), so turns succeed with tools instead of hard-400ing. Pure-search requests still use the cheap 2.5-flash fallback. Live-verified: mixed went from 100% 400 to 200 on Claude and 3.8-tiered; pure search/function paths unregressed. Includes unit tests.

Fixes #7080


MHJoy99:

MHJoy99:
recheck

MHJoy99:
recheck

MHJoy99:
recheck

PR #7106, opened 2026-09-13
feat(streaming): first-token watchdog and identity encoding on streams

### LINKED ISSUE #3603: Streaming `/responses` withholds `response.created` + the entire reasoning phase until output begins (pre-output hold) — reasoning clients see a frozen stream via OpenAI-compatible / apikey upstreams
> **⚠️ CORRECTION (see the comment below): the root cause is SSE response *compression* buffering (gzip/zstd), not the pre-output failover hold described in this body. The symptom + repro are valid; the mechanism analysis below is superseded.**

---

## Summary

For a streaming `/v1/responses` request to a **reasoning** model through an OpenAI account of type **`apikey`** (OpenAI-compatible upstream, e.g. another sub2api used as upstream), sub2api sends the client **no real SSE events until the model's output begins** — `response.created` and the whole reasoning phase are buffered and flushed only when the first output item appears. On a reasoning-heavy turn that means the client stares at a dead stream (only keepalive comments) for the full reasoning duration (tens of seconds to minutes).

Hitting the *same* upstream **directly** delivers `response.created` almost immediately (decoupled from output), so the client shows "thinking" right away. The delay is introduced by sub2api's gateway, not the model or the network.

## Environment

- sub2api inspected at **v0.1.139** (behavior appears unchanged through 0.1.141).
- Front account: OpenAI, **type `apikey`**, `base_url` pointing at an OpenAI-compatible `/responses` upstream that serves a reasoning model (reproduced with another sub2api as the upstream; any upstream that does *not* emit early reasoning-summary events should trigger it).
- Client: Codex CLI / raw `curl -N` on `/v1/responses`, `stream: true`, `reasoning.effort: high`.

## Reproduction

1. Configure a front sub2api with an `apikey` OpenAI account whose `base_url` is an OpenAI-compatible `/responses` endpoint serving a reasoning model.
2. Send a reasoning-heavy streaming request:
   ```bash
   curl -sN https://<front>/v1/responses \
     -H "Authorization: Bearer <key>" -H "Content-Type: application/json" \
     -d '{"model":"<reasoning-model>","stream":true,"reasoning":{"effort":"high"},
          "input":[{"role":"user","content":"<a prompt that induces long reasoning>"}]}' \
   | while IFS= read -r l; do printf '%s  %s\n' "$(date +%s.%N)" "$l"; done
   ```
3. Timestamp the arrival of each SSE event.

**Observed timeline (representative):**

| | `response.created` reaches client | first `response.output_text.delta` |
|---|---|---|
| **direct to the upstream** | ~1.5s | ~28s |
| **via the sub2api front (apikey)** | ~28–42s (bunched with output) | ~28–42s |

Toggling `openai_passthrough` on the front account does **not** change this — the hold is present in both the managed and passthrough streaming paths.

## Expected

`response.created` (and streamed reasoning, if the upstream emits it) should reach the client promptly, so a reasoning turn shows immediate activity instead of a frozen stream — matching the direct-to-upstream behavior.

## Root cause (analysis)

The OpenAI `/responses` streaming handlers hold "preamble" events in a **pre-output failover buffer** until the first *client-output* event, so the gateway can still failover to another account before committing output to the client. Relevant pieces (function names as of v0.1.139; line numbers may differ):

- `openAIStreamEventIsPreamble` whitelists only `response.created` / `response.in_progress` as preamble.
- `openAIStreamDataStartsClientOutput` = "not preamble"; the pending queue is flushed only once this is true.
- Both `handleStreamingResponse` and `handleStreamingResponsePassthrough` implement this hold.

On an **OAuth → ChatGPT** upstream, an early non-preamble event (a reasoning-summary `output_item.added`) arrives within ~2s and releases the hold, so preamble is flushed early. On an **`apikey` / OpenAI-compatible** upstream that doesn't surface that early event, the first non-preamble event is the actual output — so the hold releases only when output begins, withholding `response.created` and the reasoning phase for the entire think time.

> Open sub-question for maintainers who know the code best: exactly why the early release event doesn't reach the client on the chained/OpenAI-compatible path (upstream not forwarding reasoning-summary items vs. the front not counting them). Happy to gather more traces.

## Why this is distinct from the keepalive issues

This is **not** the SSE-keepalive problem (#2121, #3171). Those keep the socket alive to avoid CDN 524s; keepalives are already emitted here, yet the client still sees no *real* events until output. The ask is to **forward the preamble/reasoning early**, not to keep the connection open.

## Design consideration

The hold exists on purpose (pre-output failover). A blanket early-flush would trade that away. Suggested directions:

- Release the pre-output hold on `response.created` (or after a short grace/keepalive window) **when the account has no failover candidate** (single upstream in the group), where the hold is pure cost.
- Or a config flag (e.g. `openai_flush_preamble_before_output`) to opt in.
- Keep failover intact until the first *content* event where multiple candidates exist.

## Related

- #2121, #3171 — keepalive during pre-output wait (same symptom family, different fix).
- #2349, #2402 — "prime / flush stream preamble before upstream first byte" (images; same early-flush idea, currently open).
- #3172, #2976 — downstream keepalive PRs.

### LINKED ISSUE #5290: No first-token timeout: streaming request can hang for minutes when upstream stalls
## Summary

Sub2API has no first-token timeout. When an upstream accepts a streaming `/v1/responses` request but then stalls (no first token, no error), the request hangs indefinitely until the *client* gives up. Real-world impact: requests that eventually time out after **5-13 minutes** with zero output, wasting user time and hiding upstream problems.

## Environment

- Sub2API version: **0.1.171** (also present in earlier versions)
- Endpoint: `POST /v1/responses`, `stream: true`

## Evidence

From `usage_logs.first_token_ms` (3 days, ~2400 streaming requests):

| metric | value |
|---|---|
| p50 | 1.34s |
| avg | 3.58s |
| p99 | 14.8s |
| max | **819s (13.6 min)** |

The extreme tail (28 requests > 10s, max 819s) contains requests where the upstream simply never produced a first token; Sub2API kept waiting and only terminated when the client cancelled (`context canceled` → `openai.ws_bind_response_account_failed`).

## Expected behavior

A configurable **first-token timeout** per request/account/channel:
1. If no first token arrives within `N` seconds (e.g. 30-60s, configurable), abort the upstream call.
2. Optionally **failover** to another eligible account in the pool (when available) instead of failing the request.
3. Return a clear SSE `response.failed` / `event: error` to the client so it does not hang.

## Why it matters

Without it, a single slow/stuck upstream (e.g. a busy gateway endpoint during peak hours) turns into multi-minute hangs for clients, which also drives the repeated-client-retry pattern seen in logs. A first-token timeout makes the gateway degrade gracefully: fast failure or transparent failover, instead of silent stalls.

## Suggested scope

- Account-level setting `first_token_timeout_ms` (or global default), applied during streaming forward before the first upstream byte.
- On timeout: failover to next account if pool > 1; otherwise return SSE error and close.

MHJoy99:
Two streaming resilience fixes:
- stream_first_token_timeout (default 60s, 0 disables): abort streams whose upstream never produces a first token, emitting a retryable first_token_timeout error so callers fail over instead of hanging for minutes.
- Accept-Encoding: identity on streaming upstreams so gzip/zstd intermediaries stop buffering SSE reasoning deltas.

Code + config review verified; the watchdog path needs a genuinely stalled upstream to trigger live.

Fixes #5290, fixes #3603


MHJoy99:

MHJoy99:
recheck

PR #7139, opened 2026-09-14
fix: keep API-key accounts available for unknown models

### LINKED ISSUE #7123: 401 with an "unknown model" body permanently disables an apikey account (SetError, schedulable=false) — model names are caller-controlled, so one typo takes the whole group offline and every other user sees a misleading 503; case 400 already content-discriminates, case 401 does not
### Summary

For an `type=apikey` account, **any** upstream `401` permanently disables the account (`status=error`, `schedulable=false`). Some upstreams return `401` for an **unknown model name**, which is a caller error, not a credential problem. Because the model name is supplied by the client, a single typo from one caller takes the account offline for **every** user of that group, on the first bad request, with no automatic recovery.

The stored `error_message` ends up self-contradicting, which is the clearest symptom:

```
status=error, schedulable=false
error_message = "Authentication failed (401): Model definitely-not-a-real-model-xyz is not supported"
```

An "authentication failure" that names a model is not an authentication failure.

### Reproduce

1. Configure an `platform=openai`, `type=apikey` account with `credentials.base_url` pointing at an OpenAI-compatible upstream that returns `401` for unknown models.
2. Send a request with a model id that does not exist upstream.
3. Observe the account is now `status=error`, `schedulable=false`.
4. Every subsequent request to that group returns `503 Service temporarily unavailable` — which points at an outage rather than at the caller's typo, so the real cause is not discoverable from the error the other users see.

Verified against **v0.2.4**. A direct call to the same upstream returned `200` in the same minute, so the upstream was healthy and the state change was entirely local.

### Where

`backend/internal/service/ratelimit_service.go`, the `case 401:` arm.

It already special-cases two OpenAI shapes before falling through:

- `token_invalidated` / `token_revoked` → `handleAuthError` (correct — genuinely revoked)
- `{"detail":"Unauthorized"}` → `handleAuthError` (correct)
- `AccountTypeOAuth` → temporarily unschedulable, so a token refresh can recover it (correct, and notably *recoverable*)
- **everything else, including every `apikey` account → generic `SetError`, permanent**

So an `apikey` account is the only kind with no recoverable path at 401, and it is also the kind most likely to be pointed at a third-party OpenAI-compatible upstream with non-standard `401` semantics.

### Suggested fix

The `case 400:` arm in the same function already does exactly the right thing — it inspects the upstream message before deciding, discriminating `"organization has been disabled"`, `"credit balance"` and `"identity verification is required"`. The same treatment at `401` would be consistent with the existing design:

1. Before `handleAuthError`, check whether the upstream body indicates a **model** problem (e.g. `not supported`, `unknown model`, `model_not_found`). If so, return the error to the caller and leave account state untouched — it is a 4xx for the *request*, not for the *credential*.
2. Optionally, require N consecutive auth failures before permanently disabling an `apikey` account, mirroring the cooldown that OAuth accounts already get.

Either alone fixes the reported blast radius; (1) is the smaller change.

### Why it matters

The severity is not the disable itself — it is that the trigger is **caller-controlled and unvalidated**, the blast radius is **the whole group rather than the caller**, and the symptom presented to everyone else (`503`) actively misdirects the investigation.

Happy to open a PR if the maintainers prefer a particular shape for the discrimination.

HP-network:
## Summary

A model name rejected with HTTP 401 can currently disable an API-key account for the whole group. This happens with compatible upstreams that use 401 for unknown models.

This change recognizes an explicit unknown-model response and applies the existing per-model cooldown instead of marking the account as failed. Authentication failures such as invalid credentials continue to use the existing account-level handling.

Fixes #7123

## Verification

- `go test -tags=unit ./internal/service -run 'TestRateLimitService_HandleUpstreamError_(NonOAuth401|OAuth401|APIKeyModel401)|TestIs(OpenAICompatibleModelNotFound400|UpstreamModelNotFoundError)' -count=1`
- `git diff --check`


HP-network:

PR #7200, opened 2026-09-15
fix(gemini): preserve billing after client disconnects

zxczrx123:


zxczrx123:

PR #7223, opened 2026-09-16
feat(accounts): add Pelican SVG test and passive hourly performance metrics

NIKOPACK:
## Summary

Adds two account-management features in separate commits:

1. An optional Pelican SVG generation test in the existing account test dialog.
2. Passive account performance statistics for the rolling last hour, displayed immediately after the usage-window column.

### Pelican SVG test

- Adds a `pelican` mode for OpenAI accounts using Responses or Chat Completions, preserving the existing default and compact modes.
- Uses the fixed prompt `Generate an SVG of a pelican riding a bicycle` with medium reasoning effort.
- Extracts the generated SVG and reuses the existing image SSE event and preview. Missing or incomplete SVG output is reported as a test failure.
- Provides a visual generation exercise, not an automated intelligence score.

### Passive rolling-hour performance

- Adds the admin-only `POST /api/v1/admin/accounts/performance/batch` endpoint, accepting up to 1,000 account IDs.
- Aggregates existing usage logs in one query over `[window_end - 1 hour, window_end)`, using the existing account/time index.
- Caches snapshots for 30 seconds with request coalescing; the response includes the actual snapshot window.
- Adds a hideable account-list column with independent metric sample counts, empty states, and stale/error indicators.
- Refreshes every 30 seconds independently of account-list auto-refresh. Polling pauses when the page or column is hidden; requests are cancelled on scope changes and unmount.
- Includes Chinese and English labels and metric explanations.

| Metric | Calculation |
| --- | --- |
| TTFT | Mean valid recorded first-token time for eligible streaming text requests |
| Output TPS | Total output tokens divided by total elapsed time after the first token, using records with a positive elapsed interval |
| Cache rate | Cache-read tokens divided by input + cache-creation + cache-read tokens, including eligible nonstreaming text requests |

Known image, video, audio, live, cyber-blocked, and compaction records are excluded from text metrics. Request counts and last-request timestamps still reflect all records in the window. Missing samples or invalid denominators produce `null`, not a misleading zero.

These metrics describe recorded usage: TTFT is the gateway's recorded first-response timing, TPS is an estimate, and cache rate is the recorded/billing token share rather than a request-level upstream cache-hit guarantee. Sample counts do not assert successful completion.

### Scope

The passive feature sends no active probes or upstream requests. It introduces no schema migrations, account-model changes, or changes to usage recording, forwarding, billing, scheduling, or quota behavior. It remains separate from the explicitly invoked Pelican test.

## Validation

[...]

PR #7230, opened 2026-09-16
feat(accounts): add Pelican SVG test and passive hourly metrics

NIKOPACK:
## Summary

Supersedes #7223 with corrected GitHub-linked commit attribution and fixes from a follow-up review. The original PR history is not rewritten.

This adds two account-management features:

1. An optional Pelican SVG generation test in the existing account test dialog.
2. Passive rolling-hour account performance metrics, displayed immediately after the usage-window column.

## Pelican SVG test

- Supports OpenAI Responses and Chat Completions accounts with the fixed prompt `Generate an SVG of a pelican riding a bicycle` and medium reasoning effort.
- Reuses the existing image SSE event and preview; normal and compact tests remain separate.
- Recovers complete text from done/terminal Responses events without duplicating streamed deltas.
- Validates balanced XML SVG output, including nested SVGs, rather than truncating at the first closing tag.
- Handles final SSE lines without a trailing newline and rejects malformed, failed, or incomplete responses.
- Bounds SVG size to 2 MiB, accumulated text to 4 MiB, and text parts to 1,024.
- Reuses the existing OpenCode session-header helper, preserving inbound and configured header precedence.

This is a visual generation exercise, not an automated intelligence score.

## Passive rolling-hour performance

- Adds admin-only `POST /api/v1/admin/accounts/performance/batch`, accepting up to 1,000 normalized account IDs.
- Aggregates existing usage logs in one query over `[window_end - 1 hour, window_end)`, using the existing account/time index.
- Uses a 30-second snapshot cache with singleflight, a 128-snapshot capacity, and active expiration even when old account combinations are never requested again.
- Keeps the existing unbounded cache constructor behavior for other endpoints; only account performance opts into bounded caching.
- Adds a hideable account-list column with independent metric sample counts, empty states, stale/error indicators, and Chinese/English explanations.
- Polls every 30 seconds independently of account-list auto-refresh, pausing when the page or column is hidden and cancelling obsolete requests.

| Metric | Calculation |
| --- | --- |
| TTFT | Mean valid recorded first-token time for eligible streaming text requests |
| Output TPS | Total output tokens divided by total elapsed time after the first token, using positive elapsed intervals |
| Cache rate | Cache-read tokens divided by input + cache-creation + cache-read tokens, including eligible nonstreaming text requests |

Known image, video, audio, live, cyber-blocked, and compaction records are excluded from text metrics. Request counts and last-request timestamps still include all records in the window. Missing samples or invalid denominators produce `null`.

These are logged measurements: TTFT reflects recorded first-response timing, TPS is an estimate, and cache rate is a recorded/billing token share, not an upstream request-level hit guarantee. Sample counts do not assert successful completion.

## Scope and regression coverage

The passive feature sends no active probes or upstream requests. No schema migrations or changes to usage recording, forwarding, billing, scheduling, or quota behavior are introduced.

- Targeted Go tests passed across service, repository, admin handler, routes, and API compatibility packages.
- Race-enabled cache tests passed, including the exact production capacity of 128, expiry without reads, refresh preservation, and concurrent writes. Bounded-cache regressions also passed 20 repeated runs.
- `make test-frontend-critical` now includes the account feature suites: all 289 tests across 27 files passed.
- Changed frontend ESLint and typecheck passed.
- Repository integration tests compile with the `integration` tag. The Docker-backed integration suite was not executed locally because Docker is unavailable.

The replacement branch was created from official `main`, and official `main` was pulled again immediately before committing.


NIKOPACK:

PR #7231, opened 2026-09-16
fix(auth): support camelCase OIDC token responses

steven-ld:
Adds camelCase fallbacks for generic OIDC token responses while retaining RFC 6749 snake_case fields as the preferred source.

This supports providers that return `accessToken`, `refreshToken`, `tokenType`, and `expiresIn` after an authorization-code exchange.

Validation: `go test ./internal/handler -run 'TestOIDCParseTokenResponseSupportsCamelCaseJSON|TestOIDCParseAndValidateIDToken|TestOIDCParseUserInfoIncludesSuggestedProfile' -count=1`


steven-ld:

PR #7273, opened 2026-09-17
feat(subscription): 管理端调整重置周期，并支持分组批量重置配额

zcychar:
## Summary

Administrators currently have **Reset Quota**, which zeros usage and restarts windows from now. There is no way to move a subscription onto another reset cycle while keeping used amounts, and group management has no quota action.

This PR adds two separate admin operations:

1. **Set reset cycle** — change daily/weekly/monthly window starts and leave used quota unchanged.
2. **Reset group quota** — on subscription groups in Group Management, apply the same behavior as the subscription-page Reset Quota action to every active subscription: zero selected usage and restart those windows (daily at midnight, weekly/monthly from now).

## Why

Shared subscription groups often need the same weekly/monthly reset clock. Reset Quota cannot do that because it also clears usage. After an expired subscription is re-assigned, operators have no UI to realign windows without wiping usage.

## API

- `POST /api/v1/admin/subscriptions/:id/quota-windows`
- `POST /api/v1/admin/groups/:id/subscriptions/quota-windows`
- `POST /api/v1/admin/groups/:id/subscriptions/reset-quota`
- `POST /api/v1/admin/subscriptions/bulk-action` action `set_quota_windows`

Group endpoints only accept `subscription` billing groups. They iterate **active, unexpired** subscriptions and return partial-success counts.

`quota-windows` never writes usage columns. `reset-quota` reuses `AdminResetQuota`.

## UI

- Subscriptions: per-row **Set Reset Cycle**, plus a bulk action.
- Groups: subscription groups get **Align Reset Cycle** and **Reset Group Quota**. Standard billing groups do not.

The two group buttons are separate so aligning a cycle cannot be mistaken for clearing usage.

## Tests

- Backend unit tests for keep-usage window updates, daily midnight snap, expired-row skip, and group reset matching per-subscription reset-quota.
- Frontend API, GroupsView visibility/action, bulk dialog, and i18n completeness tests.

## Notes

- No schema/migration.
- Does not change expired assign/renew behavior.
- Related but distinct from #2660 / #5539 (those reset usage). This PR puts group reset-quota on **Group Management**, not the subscription list filter.


zcychar:

PR #7276, opened 2026-09-17
fix(openai): keep transport heartbeats outside semantic output

cyhhao:
## Summary

- Treat the exact Responses `keepalive` event as transport metadata, not semantic model output.
- Preserve pre-output retry/failover, first-output timeouts, and semantic first-token timing after heartbeat-only activity.
- Forward WebSocket HTTP bridge heartbeats immediately without releasing staged lifecycle metadata or committing the attempt.
- After a known client disconnect, drain the current bridge attempt for terminal usage instead of starting failover.

## Problem

A Responses stream can send lifecycle metadata and a heartbeat before any model output:

```text
response.created
keepalive
error: server_is_overloaded
response.failed
```

The shared classifier currently treats `keepalive` as output. That closes the safe pre-output retry boundary, records a first-token time without model content, and can disarm the native first-output timeout.

The WebSocket HTTP bridge needs an additional distinction: merely classifying the heartbeat as metadata would buffer it and leave the downstream connection idle. Heartbeats must reach the client immediately, without flushing attempt-local metadata or setting/resetting semantic commitment. If a heartbeat write reveals a disconnect, later overload, read-error, or EOF handling must not retry a request for that disconnected client; a failed terminal can still carry usage that needs to be retained.

## Changes and boundaries

- Extend the existing metadata predicate with only `keepalive`.
- Reuse the bridge's existing writer, semantic-output state, and disconnect state; do not add a new heartbeat loop or retry mechanism.
- Keep real content and unknown events conservative: they still commit output.
- Keep replay protection after content, even when later heartbeats arrive.
- Keep existing HTTP SSE comment keepalives, retry budgets, configuration, and dependencies unchanged.

Related: #5497 addresses metadata preambles and SSE comment accounting. This change specifically handles JSON `keepalive` events and the HTTP-to-WebSocket bridge boundaries, and applies directly to current `main`.

## Validation

- The new/extended regressions reproduced the heartbeat classification, retry, timeout, timing, and bridge-boundary failures against unmodified upstream code.
- The related streaming, failover, TTFT, and WebSocket bridge regression suite passes, including actual WebSocket frame forwarding.
- Complete `internal/service` and `internal/handler` unit-test packages pass.
- Independent read-only review of the complete patch and consuming paths: no actionable findings.
- Synthetic coverage includes immediate heartbeat delivery while upstream is idle, no staged metadata leakage, post-content replay protection, downstream write failures, and disconnect followed by failed-terminal usage, read failure, EOF, or DONE without a terminal.

Commands run from `backend/`:

```sh
go test ./internal/service -run 'Test(OpenAI(Stream|Streaming|ResponsesTTFT|VisibleOutput|NativeMetadata)|ProxyOpenAIWSHTTPBridgeTurn|OpenAIWSHTTPBridgeRelaysSSEFrames)' -count=1 -timeout 5m
go test -tags=unit ./internal/service ./internal/handler -count=1 -timeout 10m
```


cyhhao:
@codex review

PR #7312, opened 2026-09-18
fix(accounts): keep Codex model aliases through edit and test picker

baryon:
## Summary

Codex users who map GPT model names onto a third-party OpenAI-compatible account (for example DeepSeek) currently lose that mapping in two admin surfaces:

- The **edit account** dialog saves whitelist and alias mappings together (`combined`). The UI shows exclusive tabs, so after reload the aliases hide behind the whitelist tab.
- The **account test picker** lists the raw upstream catalog (`deepseek-chat`, …) instead of the mapped public names (`gpt-4o`, …), so the aliases cannot be selected for a test.

This PR matches create / bulk-edit: save only the active tab, and project `model_mapping` onto a local copy of the catalog before filling the test picker (same helper as gateway `/v1/models`).

The account shape we hit in production is the one described in #7285: `platform=openai`, `base_url=https://api.deepseek.com`, GPT aliases mapped onto DeepSeek models.

## References

Issues and PRs this change is based on:

- https://github.com/Wei-Shaw/sub2api/issues/7285 — `[Bug] DeepSeek 的 /responses 接受但静默忽略 input[].additional_tools…`  
  User story for Codex GPT→DeepSeek `model_mapping` on an openai-platform API-key account. That setup is how we found the edit-tab merge and the test picker showing upstream IDs.
- https://github.com/Wei-Shaw/sub2api/issues/6894 — `模型测试不受模型限制管控`  
  Test picker should only show models from the account restriction, not the raw upstream list.
- https://github.com/Wei-Shaw/sub2api/issues/7013 — `测试模型不是按照拉去列表出的`  
  Same picker: listed models do not match the configured mapping/list.
- https://github.com/Wei-Shaw/sub2api/issues/6796 — `为什么账号管理测试链接选择测试模型的时候无法显示模型名称？`  
  Test picker missing usable model names (related display of mapped/public IDs).
- https://github.com/Wei-Shaw/sub2api/pull/7313 — sibling PR for the DeepSeek thinking-mode `reasoning_content` 400 on this same mapped-account path.

Relates to: #7285 #6894 #7013 #6796 #7313

## Test plan

- [ ] Edit an OpenAI API-key account that already has whitelist + alias mapping. Stay on the mapping tab, save, reopen: mapping tab is active and aliases are intact.
- [ ] Same account, whitelist tab: rewrite the whitelist, save, reopen: only whitelist entries remain.
- [ ] Open the test dialog: picker shows mapped GPT names, not upstream DeepSeek IDs.
- [ ] `go test ./internal/service/ -run 'TestFetchOpenAIAccountModelsAPIKeyAppliesAliasMapping|TestFetchOpenAIAccountModelsAPIKeyKeepsWhitelistEntries'`
- [ ] `pnpm exec vitest run src/components/account/__tests__/EditAccountModal.spec.ts -t "saves only|opens the mapping tab"`


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

Wei-Shaw:
https://github.com/Wei-Shaw/sub2api/pull/7491

PR #7315, opened 2026-09-18
feat(openai): add one-hour Codex tickets with live harvest proxy settings

Tinghecui:
## Summary

Add opt-in, per-account/model `x-codex-turn-state` tickets and a live harvest-proxy setting under **System settings → Gateway → Codex settings → 292 harvest proxy**. This extends the existing turn-state relay/guard from #5668.

When enabled, the gateway probes active OpenAI OAuth accounts through a dedicated HTTP(S)/SOCKS harvest proxy. It accepts HTTP 200 responses containing a 292-character `gAAAAA…` header, stores tickets for 3600 seconds by default, and refreshes them 600 seconds before expiry. Harvest connections do not reuse keep-alive or HTTP/2 connections; the proxy provider is responsible for rotating egress IPs.

Production Responses requests (native, passthrough, Anthropic compatibility and WebSocket) use the selected account's normal proxy and inject its valid ticket. Configured models are blocked without a ticket only when `fail_closed` is enabled. The feature remains disabled by default.

## Admin configuration

- The administrator can replace the harvest proxy URL without a redeploy. The saved setting overrides YAML/environment configuration. The runtime cache lasts up to five seconds and subsequent probes use the new URL; existing valid tickets remain usable until refresh.
- GET and PUT responses mask proxy passwords. Omitted, empty or unchanged masked values preserve the stored URL. Invalid URLs are rejected without echoing credentials.
- Proxy examples use escaped `@` in both locales to avoid vue-i18n rendering errors.
- Account status uses the configured models and fail-closed policy; a disabled feature does not display misleading paused models.

## Ticket state protection

Ticket data is server-managed and omitted from editable account `extra` responses. Account updates preserve the latest database tickets under the existing row lock, including tickets captured after an edit started. Concurrent model probes own their account metadata maps. The unused burst-harvest implementation and its ineffective configuration options have been removed; tests exercise the continuous probe path.

## Configuration

```yaml
gateway:
  openai_codex_ticket:
    enabled: false
    target_length: 292
    ttl_seconds: 3600
    refresh_before_seconds: 600
    harvest_proxy_url: "" # fallback when no proxy is saved in admin settings
    harvest_probe_interval_seconds: 6
    harvest_attempt_timeout_seconds: 25
    fail_closed: true
    models:
      - gpt-6-astra
      - gpt-5.6-sol
```

Environment fallback: `GATEWAY_OPENAI_CODEX_TICKET_ENABLED` and `GATEWAY_OPENAI_CODEX_TICKET_HARVEST_PROXY_URL`.

## Validation

- Ticket service tests under `go test -race`, including simultaneous model probes.
- Admin settings save/read, hot reload, password masking and URL validation tests.
- Account edit/persistence protection and configured-status regression tests.
- HTTP harvest transport regression; configuration environment reachability and API contracts.
- Server package compilation.
- SettingsView and both locale suites: 93 tests passed.
- `pnpm run build` (Vue/TypeScript checking and production build).

No production credentials or provider-specific session rewriting are included.

## Lifecycle and account handling

- Synthetic harvest requests use the dedicated no-reuse HTTP transport directly, avoiding both plugin routing and the plugin initialization race. They close the stream after reading headers.
- Shutdown cancels account enumeration, settings reads, upstream probes and persistence; repeated start/stop calls are safe.
- Account responses and account backup exports strip ticket blobs and the retired account-level harvest-proxy secret. Ordinary backup credentials remain unchanged.
- Credential shadows retain the existing forwarding policy and are consistently excluded from harvest, ticket gating, injection and ticket status.
- Setup-token accounts display ticket status without exposing OAuth-only quota-reset controls.
- The live toggle is read through the settings cache rather than mutating shared startup configuration; account status reads the effective live policy after a restart.

Validation additionally covers plugin wiring under `-race`, cancellation at each blocking stage, export redaction, shadow policy and setup-token rendering. The focused frontend suites pass all 93 tests; production frontend build, server compilation, regenerated Wire output and API contracts pass.


Tinghecui:

lovinrain:
please merge this much-needed PR

PR #7391, opened 2026-09-20
fix(codex): align Astra catalog default context with Codex

jiajiaxd:
## Problem

Locally generated Codex manifests advertised GPT-6 Astra as `context_window=1050000` and `max_context_window=1050000`, using the API model capacity as the Codex default. Codex 0.155.0 ships a 272000 default and an 872000 configurable maximum. Generic OpenAI API and models.dev snapshots could also reintroduce the 1.05M default after synchronization.

## Changes

- Generate Astra (and the `gpt-6` alias) with 272000/872000 when no authoritative Codex metadata is available.
- Keep generic API/registry capacity metadata from inflating Astra catalog defaults, without changing the stored API metadata or lowering explicitly smaller provider limits.
- Preserve explicit native upstream manifest values; if a native Astra manifest supplies only a default window, use it as the missing maximum.
- Extend coverage for native OAuth manifests, registry/API snapshots, custom providers, legacy limits, and converted API-key model lists.

Related to #7042, which covered the opposite maximum-window loss.

## Validation

`go test ./internal/service ./internal/handler ./internal/server/routes -count=1` (Go 1.27 container): all passed.


jiajiaxd:

goransh-walia:
This looks good to me at a glance — thanks for putting it together.

PR #7393, opened 2026-09-20
fix(antigravity): preserve string const constraints in tool schemas

Achordchan:
## Summary

Antigravity request conversion drops string `const` constraints when a schema has an explicit type, and leaves an unsupported `const` keyword untouched when the type is omitted. For example, `{ "type": "string", "const": "browser" }` currently becomes `{ "type": "string" }`.

Convert string constants to singleton enums before schema filtering, infer `string` only when no type is provided, and intersect with existing enum values so the conversion does not widen the allowed values. Add nine table-driven cases through `buildTools`, covering nested properties, array items, empty strings, conflicting enums, metadata removal, and a property named `const`.

Related to #7392. This addresses a reproduced request-side constraint loss; it does not establish the cause of the reported client session-restoration error. Non-string constants and response rewriting are outside this change. It leaves the array traversal changes proposed in #7077 untouched.

## Validation

- Before the fix, five regression cases failed on the current main branch behavior.
- `go test ./internal/pkg/antigravity -count=1` passed.
- `go test -race ./internal/pkg/antigravity -count=1` passed, including all nine new cases.
- `go vet ./internal/pkg/antigravity` and `git diff --check` passed.
- Tested locally with Go 1.27.1. Full backend and database integration suites, and live Antigravity requests, were not run.


Achordchan:

PR #7395, opened 2026-09-20
fix(frontend): ensure Codex base URLs include /v1

zhuzhihaocool-bit:
# fix(frontend): ensure Codex base URLs include `/v1`

## Summary

- Normalize the `openai` group's Codex and Codex WebSocket `base_url` to include exactly one `/v1`.
- Keep the same group's Claude Code `ANTHROPIC_BASE_URL` on the bare origin because Claude Code appends `/v1/messages` itself.
- Normalize the Codex endpoint generated by the CC Switch import link.
- Add regression coverage for base URLs with and without a trailing slash or `/v1`.

## Root cause

When `api_base_url` is empty, the front end falls back to `window.location.origin`.
The `openai` branch passed that unnormalized URL to the Codex generator, producing:

```toml
[model_providers.OpenAI]
base_url = "https://api.example.com"
```

Codex appends `/responses` directly and does not add `/v1`, so requests go to
`https://api.example.com/responses` instead of `https://api.example.com/v1/responses`.

The same file already computes `baseRoot` and `apiBase`; other Codex-capable branches use
`apiBase`, but the `openai` branch did not. The CC Switch utility likewise already has
`withV1Endpoint()`, but the `openai` branch did not call it.

Configuring `api_base_url` with `/v1` is not a safe workaround because the same value also
feeds the Claude Code tab, whose base URL must remain on the bare origin.

## Verification

- `vitest run src/components/keys/__tests__/UseKeyModal.spec.ts src/utils/__tests__/ccswitchImport.spec.ts`
  - 36 tests passed
- `vue-tsc --noEmit`
- `eslint` on all four changed files
- `git diff --check`

Base revision: `1a9d49e16f7a22c432b428fce4af8d731f1fa364`


zhuzhihaocool-bit:

PR #7403, opened 2026-09-20
fix(accounts): preserve model edits during refresh

jeesoul-001:
## Problem

When an administrator syncs upstream models while editing an account, the account list auto-refresh can replace `props.account` with a new object for the same account ID. `EditAccountModal` treated every object-reference change as a new edit session and rehydrated the whole form, so the newly selected model was replaced by the stale `credentials.model_mapping` before the user clicked Update.

This affects OpenAI-compatible API-key accounts and can make a successful model sync appear to have no effect. The related model-mapping tab and test-picker behavior is tracked separately in #7312.

## Changes

- Keep the in-progress edit state when a same-ID account object is refreshed; rehydrate only when opening the dialog or switching to another account.
- Add a regression test proving that a refreshed same-ID account does not overwrite an unsaved model selection.

## Validation

- `pnpm exec vitest run src/components/account/__tests__/EditAccountModal.spec.ts src/components/account/__tests__/ModelWhitelistSelector.spec.ts` (67 passed)
- `pnpm run typecheck`
- `pnpm run lint:check`
- `pnpm run build`

Related: #7312


jeesoul-001:

jeesoul-001:
recheck

PR #7411, opened 2026-09-20
fix(antigravity): neutralize Claude Agent SDK identity in system prompts

### LINKED ISSUE #7410: Antigravity 429: stripping the attribution line is not enough — Claude Agent SDK identity in system prompt also triggers RESOURCE_EXHAUSTED
## Summary

Stripping the `x-anthropic-billing-header:` line (added in be4a499, documented in `docs/ANTIGRAVITY_ATTRIBUTION_429.md`) removes one trigger of the Antigravity `429 RESOURCE_EXHAUSTED` class, but not all of them.

Claude Code's Agent SDK prompts carry a second, separate system block:

```text
You are a Claude agent, built on Anthropic's Claude Agent SDK.
```

Requests that keep this block are still rejected by the Google upstream, even with the attribution block already removed.

## Environment

- **Sub2API 0.2.7** (commit aea725f), which already contains be4a499, so the attribution line is stripped by the gateway before the request leaves. Antigravity OAuth accounts (Pro), mapped model `gemini-3.8-flash-high`
- Claude Code 2.1.278, `claude-vscode` entrypoint, Anthropic Messages protocol (`/v1/messages`, streaming)
- The failing call is Claude Code's **session title generation**, not the main conversation: ~4 KB body, 3201 characters of system text, 1 user message, **0 tools**

Its system field is three text blocks:

1. attribution metadata only — dropped by the current fix
2. `You are a Claude agent, built on Anthropic's Claude Agent SDK.`
3. the actual task instructions, no vendor names

Claude Code's *main* system prompt starts with `You are an interactive CLI tool`, so `filterOpenCodePrompt` already removes the vendor identity there. That is why only the SDK-driven calls fail.

## Measurements

Every request below ran against 0.2.7, i.e. `stripClaudeAttribution` was already in effect. They were sent through a group bound to a single dedicated account, and every request waited for the per-model cooldown to expire first — otherwise a previous 429 turns the next result into `503 no available accounts` and the comparison is meaningless.

| Variant | Result |
| --- | --- |
| original request replayed | 429 (8/8) |
| same request, generic system prompt | 200 (8/8) |
| attribution block also removed client-side, on top of the gateway already stripping it | **429 (3/3)** |
| attribution removed **+ identity sentence neutralized** | **200 (3/3)** |
| only `Claude` replaced in system text | 200 (2/2) |
| only `Anthropic` removed from system text | 200 (2/2) |
| `output_config` removed / `max_tokens` lowered / non-streaming | 429 — no effect |

Request rate, concurrency and body size were ruled out separately: 429s occurred with zero successful requests in the preceding 60 seconds, 10 parallel minimal requests all returned 200, and larger requests from a Gemini CLI client on the same accounts never hit this.

## Suggested fix

After the attribution handling, neutralize a vendor identity sentence when it appears **at the start** of a system block, so that user instructions that merely mention Claude/Anthropic and model ids such as `claude-sonnet-4-6` are untouched.

PR: #7411

As with the existing note in the doc, these comparisons only support "this text triggered rejection in these requests"; they do not establish Google's internal policy.

HuntercodeT:
Fixes #7410.

## What

`stripClaudeAttribution` (be4a499) removes the `x-anthropic-billing-header:` line from Antigravity system text. That covers one trigger of the 429 `RESOURCE_EXHAUSTED` class, but Claude Code's Agent SDK prompts put the vendor identity in a **separate** system block:

```text
You are a Claude agent, built on Anthropic's Claude Agent SDK.
```

Those requests are still rejected. This PR neutralizes such an identity sentence when it opens a system block, rewriting it to `You are an AI agent.` and keeping the rest of the block.

## Why anchored at the start

Only a leading sentence is treated as a client identity declaration. User instructions that mention Claude or Anthropic mid-prompt, and model ids such as `claude-sonnet-4-6`, must not be rewritten — a rewritten model id would turn into a request for a model that does not exist. Tests cover both cases.

## Evidence

Measured on **Sub2API 0.2.7**, which already ships `stripClaudeAttribution`, so every row below already has the attribution line removed by the gateway. Requests went through a group bound to a single dedicated account, waiting for the per-model cooldown to expire before each one (otherwise a previous 429 turns later results into `503 no available accounts`):

| Variant | Result |
| --- | --- |
| original request replayed | 429 (8/8) |
| same request, generic system prompt | 200 (8/8) |
| attribution block also removed client-side | 429 (3/3) |
| attribution removed + identity neutralized (this PR) | 200 (3/3) |
| only `Claude` replaced / only `Anthropic` removed | 200 (2/2 each) |
| `output_config`, `max_tokens`, streaming toggled | no effect |

Rate, concurrency and body size were ruled out: 429s happened with zero traffic in the preceding 60 seconds, 10 parallel minimal requests all returned 200, and larger requests from a Gemini CLI client on the same accounts never reproduced it.

## Scope

- Antigravity transformer only; the native Anthropic path is untouched, since OAuth there may require the attribution.
- No change to user messages, tool definitions, model mapping, auth or quota handling.
- `docs/ANTIGRAVITY_ATTRIBUTION_429.md` gets a follow-up section with the comparison table.

## Tests

`go test -tags=unit ./internal/pkg/antigravity/` — new cases for both openers, leading whitespace, a mid-text mention that must stay, an opener that is not at the start, and the attribution + identity combination.


HuntercodeT:
The previous `test` run failed on `TestOpenAIGatewayService_ProxyResponsesWebSocketFromClient_SameCodexThreadStillPreempts`:

```
failed to close WebSocket: received close frame: status = StatusTryAgainLater
and reason = "session preempted by a newer connection"
```

That looks unrelated to this change, which only touches `internal/pkg/antigravity`:

- the first CI run on this same branch passed; the only difference since then was a doc paragraph and a commit message reword
- the test passes locally with `-count=5` on this branch
- nothing here touches the OpenAI WebSocket forwarder

I do not have permission to re-run the job, so I pushed a test that exercises the rewrite end-to-end through `TransformClaudeToGeminiWithOptions` (real block layout: attribution block, identity sentence, task instructions) and asserts that user content mentioning Claude Code and `claude-sonnet-4-6` is forwarded unchanged. That also retriggers CI.

PR #7413, opened 2026-09-20
ci: parallelize release builds across platform runners

ranxi2001:
The Release workflow currently compiles five Go targets on one runner. This moves each target to its own matrix runner, then verifies and publishes their archives without compiling them again on the publishing runner.

The frontend is built once and shared. The matrix comes from the existing GoReleaser target configuration; simple mode selects only Linux amd64. Each archive carries a source SHA, version, target and SHA256, and publication fails if any expected artifact is missing, inconsistent or corrupt. Per-target Go caches use a stable restore/save key and can be refreshed on later commits.

GoReleaser OSS still supplies archive naming/content, checksums and release templates. DockerHub/GHCR tags and simple-release behavior are retained. Runtime image instructions still use Buildx/QEMU. The new `dry_run` input builds archives and OCI images but skips registry pushes, Release publication, notifications and VERSION synchronization. Manual releases use one resolved application commit; helper scripts are passed separately from the workflow revision so older tags do not need the new scripts.

Validation:
- 10 helper tests, Bash syntax checks and actionlint passed.
- A GoReleaser OSS publication-only smoke test generated five checksums without rebuilding binaries or publishing.
- Full [cold-cache](https://github.com/ranxi2001/sub2api/actions/runs/35500040919), [restored-cache](https://github.com/ranxi2001/sub2api/actions/runs/35500390567) and [simple-mode](https://github.com/ranxi2001/sub2api/actions/runs/35500481950) dry runs passed in the owner fork. All five full-run archives matched the final checksum file and source metadata; the Linux binary reported the expected version and commit.
- The [final release workflow full dry run](https://github.com/ranxi2001/sub2api/actions/runs/35500851974) passed in 5m12s. The latest helper-test CI job also passes; application CI jobs may still be running.

Measured on the fork's unchanged 2.7.7 application sources: the previous publish workflow took 11m15s from first job start through publication, versus 5m41s for the cold matrix dry run and 5m12s–5m19s in the restored-cache runs. Pure Go compilation was 6m56s across the old runner's five targets; individual cold matrix builds took 1m21s–1m55s, and restored-cache builds about 3s each. These are single-run observations, not an upstream performance guarantee: dry runs export OCI images and do not measure registry/Release upload time. Parallel runners also trade more concurrent resources for lower elapsed time.

No application code or release tags are changed by this PR.


ranxi2001:
@Wei-Shaw ci的release编译速度严重影响了版本发布速度，我用矩阵并行构建编译降低了50%的构建耗时，please take a look!

PR #7415, opened 2026-09-20
fix: passthrough upstream 413 instead of converting to 502

paicha:
## Problem

When an upstream provider (e.g. Anthropic) returns HTTP 413 (Request Entity Too Large), the gateway converts it to a generic **502 Bad Gateway** with message "Upstream request failed". This swallows the original error context and prevents downstream clients from handling the error appropriately.

For example, Claude Code can display a helpful "Request too large, run /compact" message when it receives a 413, but a generic 502 gives no actionable information.

### Root cause

In `handleErrorResponse`, the `switch resp.StatusCode` has no `case 413` — it falls through to `default`, which returns 502. This affects all three gateway paths (Anthropic, OpenAI, Gemini).

## Fix

Add explicit `case 413` handling in all three gateway error response functions:

- **Anthropic** (`gateway_upstream_response.go`): passthrough the raw upstream JSON body with 413 status (same pattern as existing `case 400`)
- **OpenAI** (`openai_gateway_upstream_errors.go`): return 413 with `invalid_request_error` type and the upstream message
- **Gemini** (`gemini_messages_compat_service.go`): return 413 with `invalid_request_error` type and the upstream message

## Testing

- `go build ./...` passes
- The fix follows the same passthrough pattern already established for `case 400` in the Anthropic handler


paicha:

PR #7428, opened 2026-09-20
fix(openai): safely retry TLS handshake failures before HTTP is sent

blackdm666:
## Summary

Retry a transient TLS-handshake reset/EOF **once, on the same configured proxy**, only when positive `httptrace` evidence proves that HTTP has not received a connection.

This is an independent repository-layer fix based on upstream `main` at `7c700729c23187d31ed320f6b19c790e2f194826`. It does not depend on runtime proxy fallback and contains no fork-specific ticket collection or account eligibility changes.

## Safety boundary

- HTTPS + explicitly configured proxy + `HTTPUpstreamProfileOpenAI` only.
- Observe `TLSHandshakeStart` and a transient `TLSHandshakeDone` error; missing trace evidence is not permission to retry.
- `GotConn`, any header/request write, or the first response byte vetoes retry. Evidence is retained across redirects within the first `Client.Do`.
- The request body must be reconstructable. Rebuild it for the second attempt, preserve the caller's context/trace, and use a cancellable 150 ms backoff.
- At most one additional attempt with the same client/proxy and unchanged TLS verification.
- Do not retry certificate errors, cancellation, received HTTP errors, or an already-sent POST. Do not change account cooldown/failover policy.
- Emit a credential-free retry marker.

This mitigates a narrowly identified pre-request connection failure; it does not repair a proxy provider, promise uninterrupted streaming, or guarantee every network error will recover.

## Validation

Run separately against this standalone upstream-based branch, not only against a combined downstream build:

- Full backend unit suite: `go test -json -p 2 -tags=unit ./... -count=1` — passed; 20,311 test/subtest pass events, 17 skipped tests.
- Focused safety tests — 23 pass events.
- Focused race suite, three rounds — 69 pass events, no failures/skips.
- Backend build and `go vet -p 2 -tags=unit ./...` — passed.
- `golangci-lint v2.13.0` with the upstream configuration — passed after fixing test cleanup/type-assertion checks; the three-round race suite was rerun successfully on the final test code.
- Real local TLS-listener tests and synthetic transport tests cover reset recovery, exactly-once POST delivery, persistent failure bounds, cancellation, certificate failures, missing trace, writes/response evidence, and redirects.

Skipped full-suite cases require optional external services/live credentials. No paid model requests or production changes were made. The integration-tag suite and production multi-account acceptance were not run for this split.


blackdm666:

PR #7445, opened 2026-09-21
feat(admin): remember dashboard date range and granularity

zhuyinghua6961:
## Summary

Related to #7406.

This is a partial implementation covering only the admin dashboard's date range and granularity preferences. Previously, these filters lived only in component state and were reset on every mount.

- Persist versioned preferences per authenticated user in localStorage.
- Save relative preset identifiers and recalculate their dates on entry instead of restoring stale dates.
- Preserve explicit custom dates with strict date and range-order validation.
- Restore the last valid granularity before the initial requests, without duplicate loads or initialization writes.
- Preserve automatic granularity selection when the user changes the date range.
- Safely fall back for malformed data, unsupported versions, and storage read/write failures.
- Extract and reuse the existing date preset calculations and local-date formatter without adding dependencies.

## Scope and compatibility

- The current last-24-hours/hourly default remains unchanged. #7197 is still open and is intentionally not implemented here.
- User dashboard preferences, Ops filters, and other page filters are not included.
- Existing Ops URL-query restoration remains unchanged. The admin dashboard has no existing URL-filter restoration mechanism, and this PR does not introduce one.
- No backend, billing, database, or other page default changes.
- Preferences are local to the browser and do not sync across devices.

## Validation

- Confirmed the remount regression test fails before connecting persistence, then passes with the fix.
- All 70 tests across six relevant Vitest suites pass, including admin dashboard, date picker, local-date formatting, and existing usage-page/integration tests.
- All 43 core tests also pass under `TZ=America/Los_Angeles`.
- Frontend `vue-tsc --noEmit` passes.
- ESLint passes for all six changed files (without `--fix`).
- `git diff --check` passes.

Coverage includes original defaults, remount restoration, all relative presets, day/month boundaries, explicit custom ranges, user isolation, unsupported/corrupt storage data, storage read/write failures, automatic granularity selection, and restored filters being used by the first requests without duplicate initialization loads.


zhuyinghua6961:

PR #7472, opened 2026-09-22
feat: optionally enforce API key spending limits in simple mode

superzcj:
## Problem

In simple mode, API-key monetary limits can be configured in the UI, but requests bypass their checks and usage accounting. A key configured for $30 per 5 hours can keep forwarding requests while its window counter remains zero (see #4124).

## Change

Add `SIMPLE_MODE_KEY_RATE_LIMIT_ENABLED=true` (disabled by default) to enforce API-key 5-hour, 1-day and 7-day spending limits in simple mode. This mode continues to allow zero-balance users and does not charge user balances or subscriptions.

Both gateway services reuse the existing usage-billing path, with an API-key-only command and early-return post-processing. The implementation uses database-backed window accounting and request deduplication, with database-authoritative eligibility checks so a stale or unavailable Redis cache cannot relax the new limits. Persistent Responses WebSockets are checked after account selection and before follow-up turns (both passthrough and context-pool ingress). Standard mode and default simple-mode behavior remain unchanged.

This retains post-request accounting: requests already in flight can overshoot the configured amount. It does not introduce spending reservations or reconstruct historical simple-mode consumption.

## Validation

- Full backend unit suite (`go test -tags=unit ./...`).
- Backend lint (`golangci-lint` v2.13.0).
- Frontend typecheck and existing KeysView tests (24 passed) after removing UI changes from this PR.
- Regression coverage for 5h/1d/7d eligibility, expired windows, zero balance, DB read failures, API-key-only billing, legacy-fallback guards, and cache-invalidation failures preserving last-used updates without financial cache writes.
- Actual Gateway/OpenAI usage-recording paths, streaming and non-streaming, with enabled/disabled flags. Disabling the new branches makes the enabled-case regressions fail.
- WebSocket first-turn and follow-up regressions for passthrough/context-pool modes: reproduced the bypass before adding per-turn checks, then passed after the fix.
- Deployment shell checks, including Compose configuration checks.

Docker-backed integration tests cannot run on the local development host (no container runtime); the upstream CI run is currently awaiting maintainer approval for this contribution.

Refs #4124.


superzcj:

PR #7473, opened 2026-09-22
fix(codex): include OpenCode Go in model capability resolution

lzfxxx:
## Summary

OpenCode Go accounts were excluded from Codex model-manifest capability resolution. As a result, models served by `opencode_go` accounts were advertised with the generic fallback (`text` only and `none` reasoning) even when their synced upstream metadata declared image input and reasoning levels.

This change treats `PlatformOpenCodeGo` as a capability-capable platform in the same Codex manifest path already used for OpenAI and DeepSeek accounts.

## Changes

- Include `PlatformOpenCodeGo` in the platform list used by `loadCodexGroupCatalogAccounts` when loading capability-catalog members.
- Allow `PlatformOpenCodeGo` in `groupCodexModelSupportsImageInput`.
- Read `opencode_go` account metadata in `accountCodexModelSupportsImageInput`.

The production change is limited to these three platform checks. No routing, protocol selection, billing, or database behavior is changed.

## Tests

Added a focused regression test that uses a stub which only returns an OpenCode Go account when the platform is included in the catalog candidate query. The test verifies that generated Codex model metadata includes:

- `input_modalities: ["text", "image"]`
- `context_window`
- `default_reasoning_level`
- `supported_reasoning_levels`

Validation run:

```bash
cd backend
GOTOOLCHAIN=auto go test -tags=unit ./internal/service -count=1
GOTOOLCHAIN=auto go vet ./internal/service/...
```

The full service test package passes.


lzfxxx:

PR #7512, opened 2026-09-22
feat(tls): post-quantum X25519MLKEM768 key share in the default Claude Code fingerprint

anyin233:
## Context

Real Claude Code `2.1.280` (Node.js runtime on OpenSSL 3.5+) now offers the post-quantum hybrid group `X25519MLKEM768` (`0x11ec`, RFC 9794 / draft-ietf-tls-ecdhe-mlkem) as its **first** supported group and sends a key share for it. Captured ClientHello (2026-09-22):

```text
JA3: 0303,1301-1302-...-0035,0017-ff01-000a-000b-0023-0010-0005-000d-0012-0033-002d-002b,11ec-001d-0017-0018,00
key shares: 0x11ec (1216-byte ML-KEM encapsulation key + X25519), 0x1d (X25519)
```

The bundled utls `v1.8.2` cannot represent this at all — it has no ML-KEM support — so the gateway's default TLS fingerprint (and any per-account profile using the built-in defaults) offered only the three classical groups of the older Node.js generation. This PR closes that gap.

### utls bump to post-`v1.8.2` master (`23b1dac1`)

utls master implements real ML-KEM-768 key-share generation on top of Go's `crypto/mlkem` (Go ≥ 1.24): an empty-Data `KeyShare{Group: X25519MLKEM768}` is populated at handshake time with a genuine encapsulation key + X25519 ephemeral, and the decapsulation keys are registered so a server that selects the PQ share completes the handshake normally. This is exactly the piece that was impossible to fake safely on `v1.8.2` (a forged static Data blob would fail whenever the server selects it).

### Default fingerprint upgraded to the captured shape

- `defaultCurves`: `[X25519MLKEM768, X25519, P256, P384]` — the PQ hybrid first, matching the capture.
- Default key shares: `X25519MLKEM768` + `X25519`, **independently generated keys**, mirroring OpenSSL 3.5 (which does not reuse the classical ephemeral inside the hybrid share; utls' `keyShareHybridReuseMarker` would force that reuse, so it is deliberately not used).
- Cipher suites, extension order, ALPN, signature algorithms: verified identical to the capture — untouched.

Per-account TLS profiles keep full override semantics; the defaults simply match current Claude Code now.

### New integration tests

`dialer_mlkem_capture_test.go` (build tag `integration`, external capture server, mirroring the existing `dialer_capture_test.go` pattern):

- `TestMLKEM768JA3MatchesClaudeCode280` — drives the patched spec builder against the capture server and byte-compares JA3 with the real `2.1.280` ClientHello:

```text
=== RUN TestMLKEM768JA3MatchesClaudeCode280
    handshake OK: alpn="http/1.1" cipher=0xcca8
    JA3 matches real Claude Code 2.1.280: d1fc7621a3b4c3f2b415f5c993dec798
--- PASS
```

- `TestDefaultProfilePQInterop` — the unmodified default profile (SNI + ECH GREASE + PQ share) completes full handshakes.

### Compatibility

- **Classical-only servers**: unaffected — they select the X25519 share exactly as before.
- **PQ-capable servers**: get a real, decapsulatable ML-KEM share instead of an absent one.
- **Downstream consumers of the `Profile` struct**: no API changes; profiles that explicitly set `Curves`/`KeyShareGroups` keep their values.

## Test Plan

```bash
cd backend
GOTOOLCHAIN=auto go build ./...
GOTOOLCHAIN=auto go test ./internal/pkg/tlsfingerprint/ -count=1
# with a local capture server:
TLSFINGERPRINT_CAPTURE_ADDR=127.0.0.1:9443 TLSFINGERPRINT_CAPTURE_FILE=capture.jsonl \
GOTOOLCHAIN=auto go test -tags=integration -run TestMLKEM768 ./internal/pkg/tlsfingerprint/ -v
# → JA3 matches real Claude Code 2.1.280: d1fc7621a3b4c3f2b415f5c993dec798
```

`go vet` clean, `gofmt` clean. End-to-end: gateway built from this branch forwards Claude OAuth traffic with the PQ-fingerprinted dialer (all Anthropic forward paths route through `tlsfingerprint` when the account-level TLS fingerprint flag is enabled).

## Relationship to other PRs

- #7508 raises the Claude Code client-version baseline to `2.1.280` (same capture session, HTTP-identity layer). This PR is the TLS-layer counterpart and is intentionally independent (no overlapping files) so either can land first.


anyin233:
Closing in favor of folding the TLS post-quantum work into #7508 (same capture session, one coherent Claude Code 2.1.280 adaptation PR), per author preference.

PR #7534, opened 2026-09-23
fix(antigravity): keep mid-conversation system messages in contents to preserve Gemini prefix cache

maidol:
## Problem

On the Antigravity route, Gemini prompt-cache hit rate decays as a conversation grows. Measured on a real Claude Code session through this gateway: `cache_read_input_tokens` stayed flat at ~65k–70k while the prompt grew from 104k to 151k tokens, i.e. the hit rate fell **62.5% → 46.8%**. The conversation history is never cached — only tools + systemInstruction are.

## Root cause

`buildContents` hoists **every** `role:"system"` message out of `messages` and appends it to the tail of Gemini `systemInstruction`:

```go
if role == "system" {
    systemParts = append(systemParts, parts...)
    continue
}
```

Clients that enable `anthropic-beta: mid-conversation-system-2026-04-07` (Claude Code 2.1.x does) inject a new `role:"system"` message **every turn**. Since `systemInstruction` precedes `contents` in the prompt, appending to it means each turn inserts new text *in front of* the existing conversation, so the upstream prefix cache is invalidated past that point on every single request.

## Evidence

**1. Production traffic (38 consecutive requests captured at the gateway's client side).**
Replaying those request bodies through the current transformer shows `contents` is byte-for-byte append-only every time (`lcp(contents) = 100%`), while `systemInstruction` gains one part per request. Measured against the observed usage:

| prompt tokens | cache_read | hit | `cache_read / (tools + previous systemInstruction)` |
|---|---|---|---|
| 104,353 | 65,258 | 62.5% | 0.2402 |
| 119,911 | 69,364 | 57.8% | 0.2539 |
| 139,889 | 69,922 | 50.0% | 0.2540 |
| 151,034 | 70,687 | 46.8% | 0.2556 |

That ratio is constant to ±1% across 30 requests while the prompt grows 45% — the cached prefix tracks `tools + systemInstruction` and is independent of conversation length.

**2. Natural control in the same data.** Exactly one of the 38 requests had a `systemInstruction` byte-identical to its predecessor. That single request is also the only one whose `cache_read` jumped to **141,231 (95.7%)** — ratio 0.512, double the others, because the cache reached into `contents`.

**3. Controlled experiment against a live Antigravity upstream.** Same conversation, same tools/system:

```
append only (no mid-conversation system message)   prompt=47142  cached=45006  95.5%
append + one mid-conversation system message       prompt=47165  cached=0       0.0%
```

With a realistic tools+system block (28,365 tokens), turns carrying a system message land at `cached ≈ 24.3k–24.5k` — i.e. exactly tools+system, conversation excluded — while append-only turns reach ~91%.

## Fix

Only system messages that appear **before the conversation starts** are merged into `systemInstruction`. A mid-conversation one stays where it is: merged into the preceding `user` turn, or becoming its own `user` turn if the previous content is a model turn. This keeps the upstream payload strictly append-only.

Replaying the same 38 production requests through the patched transformer: `systemInstruction` is byte-identical across every consecutive pair within a session, `contents` is strictly append-only, and the cacheable prefix share rises from 31–42% to **97–99.8%** of the payload.

## Tests

- `mid conversation system role stays in contents` — a mid-conversation system message is not hoisted and merges into the preceding user turn.
- `mid conversation system after assistant becomes its own user turn` — covers the other branch.
- `appending a turn never rewrites the existing prefix` — `systemInstruction` byte-stable, existing `contents` unchanged.
- `TestTransformClaudeToGeminiWithOptions_UpstreamPrefixIsAppendOnly` — walks five appended turns and asserts the invariant on the **whole serialized payload prefix**, not on where a single message lands; a per-message assertion cannot detect "something was inserted in front".

All four fail on the current behaviour and pass with the fix. Existing sub-tests covering system messages that precede the conversation (`message system role moves to system instruction`, `top level and message system instructions are merged`) are unchanged and still pass — that path is untouched.

`go build -gcflags=-e ./...` OK, `go vet -tags=unit ./...` clean, `go test -tags=unit -count=1 ./internal/pkg/antigravity/...` ok.

## Compatibility

Behaviour changes only for requests that carry `role:"system"` messages *after* the first conversation turn. Clients that only send a leading system block are unaffected.


maidol:

maidol:
recheck

PR #7545, opened 2026-09-23
feat(temp-unsched): add account_wide option to temp unschedulable rules

atomlong:
## Problem

The temp unschedulable rules currently block only the `(account, model)` pair for known-model failures. For providers whose quota is shared across all models on the account (e.g. monthly included credits drained account-wide), a quota-exhaustion error (402 "depleted") on one model means **every** model on that account is drained — but the scheduler keeps picking the account for other models and burns failover attempts on repeated errors (observed `switch_count` 8/10 in production).

## Solution

Add an opt-in `account_wide` flag to `TempUnschedulableRule`: when enabled, a matched rule takes the legacy account-wide block (`SetTempUnschedulable`) instead of the model-scoped one (`SetModelRateLimit`). Default `false` keeps existing behavior unchanged — old rule data needs zero migration.

## Changes

**Backend**
- Parse `account_wide` in `GetTempUnschedulableRules` (non-bool values fail-safe to `false`)
- `triggerTempUnschedulable`: skip the model-scoped branch when the rule opts in; 401 semantics unaffected (401 keeps account-wide either way)
- Record `account_wide` in the persisted `TempUnschedState` reason JSON for observability
- Clarify the fastpath comment in `openai_account_runtime_block_fastpath.go`: account-wide rules also route through the model-isolated early return, but apply their own bounded (until-based) block inside `triggerTempUnschedulable`, so the unbounded runtime blocker is still skipped

**Frontend**
- Checkbox in the rule editor (create/edit account modals), themed per existing checkbox conventions
- Rule-order hint (first match wins → put broader-scope rules first), zh/en i18n
- `TempUnschedStatusModal` shows the block scope. Note the status endpoint only ever returns account-level blocks (model-scoped ones persist in `extra.model_rate_limits`), so the scope card always reads account-wide; `account_wide` distinguishes rule-declared blocks from 401/stream-timeout fallbacks
- Types: `TempUnschedulableRule.account_wide`, `TempUnschedulableState.account_wide` (optional)

**Tests**
- Backend: parse cases (explicit true / omitted / wrong type), account-wide trigger with persisted-reason assertion (`account_wide` recorded in state JSON), default model-scoped regression guard, struct omitempty defaults
- Frontend: `EditAccountModal.tempUnschedAccountWide.spec.ts` (form echo-back, submit passthrough, toggling); `TempUnschedStatusModal` scope-display cases (rule-declared, fallback, no-keyword states)

## Testing

- `go test -tags=unit ./internal/service/` ✅, `go vet` ✅
- Full upstream CI suite green on identical code (fork branch build)
- `vue-tsc --noEmit` ✅, `check:i18n` ✅, account component specs (395 tests) ✅
- Deployed and verified on a live instance: a shared-quota account hit by 402 "depleted" is now parked account-wide for the configured 43200m instead of burning failover per model

## Follow-ups (not in this PR)

- `buildTempUnschedRules` is duplicated in both account modals; Create-side spec coverage still missing
- Optional: surface `account_wide` in the status modal rule details


atomlong:
> 
> 
> 
> 
> 

atomlong:
recheck

atomlong:

PR #7549, opened 2026-09-23
fix(ccswitch): stop usage query from requesting /v1/v1/usage

helianthus1:
#7395 made Codex (and Grok) CC Switch imports store a base URL ending in `/v1`, but the balance script sent with the import still builds its URL as `"{{baseUrl}}/v1/usage"`. CC Switch substitutes the stored base URL as-is, so every newly imported Codex/Grok provider requests `https://<host>/v1/v1/usage`, gets `404 page not found`, and shows "query failed" instead of the balance. Claude imports and Codex providers created before #7395 are unaffected because their base URL has no `/v1`.

Fix: move the script into `ccswitchImport.ts` as `CC_SWITCH_USAGE_SCRIPT` and strip a trailing `/` and an existing trailing `/v1` from `{{baseUrl}}` before appending `/v1/usage`. CC Switch replaces `{{baseUrl}}` as text and then evaluates the script in QuickJS, only checking that the resulting request URL has the same host and port as the base URL, so the expression is evaluated and passes that check. Normalizing inside the script, rather than choosing the path per platform at import time, also keeps working when users later edit the base URL in CC Switch with or without `/v1`.

Validation: 5 new tests substitute the template variables the way CC Switch does, evaluate the script, and check the request URL for base URLs with and without `/`, `/v1` and `/v1/`, plus the endpoint each platform import (anthropic / openai / grok / gemini) stores. 4 of them fail before the fix. The full frontend suite (332 files, 2489 tests), `vue-tsc --noEmit`, full ESLint and `git diff --check` pass. The same change was deployed on a production sub2api instance: after patching the script, an affected Codex provider in CC Switch 3.20.4 switched from "query failed" to showing the balance, with the request going to `/v1/usage` (200) instead of `/v1/v1/usage` (404).


helianthus1:

PR #7555, opened 2026-09-23
fix(scheduler): keep auto reset-credit fields in the scheduler projection

Yukinoshita-Yukino:
## Problem

OpenAI accounts with automatic reset credits enabled are paused at the ops quota auto-pause threshold (`openai_account_quota_auto_pause.default_threshold_*`) even when a fresh `codex_auto_reset_credit_state` reports an available credit.

`shouldAutoPauseOpenAIAccountByQuota` is designed to keep such accounts schedulable until the reset-credit threshold and then redeem a credit. But candidate filtering reads the Redis metadata projection, and `filterSchedulerExtra` drops these keys:

- `auto_reset_credit_enabled`
- `auto_reset_credit_5h_threshold` / `auto_reset_credit_7d_threshold`
- `codex_auto_reset_credit_state`

So `ResolveOpenAIAutoResetCreditConfig` always returns `Enabled=false` on the scheduler side, the pass-through branch never runs, and accounts with credits sit idle until their 5h/7d window resets on its own.

**Repro:** auto-pause 7d threshold 0.9, reset-credit threshold 0.95, account at 7d = 92% with a fresh `available` state. The full DB account is not paused; the same account after `buildSchedulerMetadataAccount` is paused on `7d`. In our deployment every account in the 90%~95% band received zero traffic although most had credits available.

## Changes

1. Keep the four keys above in the scheduler projection.
2. Once the config is visible, every scheduling pass calls `notifyOpenAIAutoReset` for each paused credit-enabled candidate, and each notification makes a worker reload the account (4 workers per instance, dedup only while queued). Add `notifyOpenAIAutoResetFromScheduler` with a 30s per-account cooldown and use it only for the three calls inside `shouldAutoPauseOpenAIAccountByQuota`. Other event-driven notifications are unchanged; the per-minute scan remains the fallback.

## Tests

- `TestSchedulerMetadataPayload_KeepsOpenAIAutoResetCreditFields` round-trips the real JSON meta payload and resolves the credit config from it. It fails without change 1.
- `TestNotifyOpenAIAutoResetFromScheduler_CoolsDownPerAccount` covers the cooldown.
- `go test ./internal/service/` passes; `go vet` clean.

## Notes

- No cache rebuild needed: `UpdateExtra` on `codex_auto_reset_credit_state` already resyncs the account snapshot, so paused accounts pick up the new projection on their next state refresh (~10 min).
- #7541 adds `auto_reset_credit_5h_enabled` / `auto_reset_credit_7d_enabled`; whichever lands second should add those keys to the projection too.


Yukinoshita-Yukino:

PR #7559, opened 2026-09-23
Fix Ollama Cloud monthly usage parsing and entrypoint permissions

magicJie:
Ollama's current settings page uses an `Included usage` plan, a monetary `Monthly usage` meter, `Current balance`, `data-time` reset timestamps, and model request counts in segment `aria-label`s. The existing parser only recognized the older 5-hour/7-day layout, so a valid browser session produced `invalid_html` and no usage snapshot.

This change adds a separate `monthly` window while preserving the existing 5-hour/7-day fields. It calculates the monthly percentage from the amount used and limit (including an aria-label fallback), rather than the segment widths that represent model distribution. The account UI shows the monthly window and model counts, and the Cookie hint now matches `__Secure-session`. Monthly included usage is not treated as a hard rate-limit reset because usage credits can extend service beyond the included allowance.

A restrictive source checkout can give `docker-entrypoint.sh` mode 0600. `chmod +x` then produces 0711; after the entrypoint drops to the `sub2api` user and re-executes itself, the container exits with `Permission denied`. Both Dockerfiles now set mode 0755 explicitly.

Validation:
- `go test ./internal/service -run 'TestParseOllamaCloudUsageHTML|TestOllamaCloudUsageExhaustionResetAt|TestOllamaCloudUsageRefresh' -count=1`
- Frontend production build (i18n check, Vue typecheck, Vite build)
- Frontend Ollama settings/cell and locale tests: 13 passed
- Docker image entrypoint starts successfully as both root and UID 1000


magicJie:

